PR #3, opened 2021-03-24
Add parts and steps definitions

cmatsuoka:
Parts and steps are the basic data types craft-parts will work with.
Together, they define the lifecycle of a project (i.e. how to process
each step of each part in order to obtain the final primed result).

-----

### DIFF
--- a/craft_parts/dirs.py
+++ b/craft_parts/dirs.py
@@ -21,7 +21,7 @@ from typing import Union
 
 
 class ProjectDirs:
-    """The project main work directories.
+    """The project's main work directories.
 
     :param work_dir: the parent directory containing the parts, prime and stage
         subdirectories. If not specified, the current directory will be used.
--- a/craft_parts/dirs.py
+++ b/craft_parts/dirs.py
@@ -1,0 +1,51 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2021 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Definitions for project directories."""
+
+from pathlib import Path
+from typing import Union
+
+
+class ProjectDirs:
+    """The project main work directories.
+
+    :param work_dir: the parent directory containing the parts, prime and stage
+        subdirectories. If not specified, the current directory will be used.
+    """
+
+    def __init__(self, *, work_dir: Union[Path, str] = "."):
+        self._work_dir = Path(work_dir).absolute()
+
+    @property
+    def work_dir(self) -> Path:
+        """Return the root of the work directories used for project processing."""
+        return self._work_dir
+
+    @property
+    def parts_dir(self) -> Path:
+        """Return the directory containing work subdirectories for each part."""
+        return self._work_dir / "parts"
+
+    @property
+    def stage_dir(self) -> Path:
+        """Return the staging area containing installed files from all parts."""
+        return self._work_dir / "stage"
+
+    @property
+    def prime_dir(self) -> Path:
+        """Return the primed tree containing the final artifacts to deploy."""
+        return self._work_dir / "prime"
--- a/craft_parts/errors.py
+++ b/craft_parts/errors.py
@@ -14,45 +14,56 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-"""The exceptions that can be raised when using craft_parts."""
+"""Craft parts errors."""
 
-from abc import ABC
-
-
-class CraftPartsError(Exception, ABC):
-    """Base class for Craft Parts exceptions."""
-
-    fmt = "Daughter classes should redefine this"
-
-    def __init__(self, **kwargs) -> None:
-        super().__init__()
-        for key, value in kwargs.items():
-            setattr(self, key, value)
-
-    def __str__(self):
-        return self.fmt.format([], **self.__dict__)
+import dataclasses
+from typing import Optional
 
 
-class CraftPartsReportableError(CraftPartsError):
-    """Base class for reportable Craft Parts exceptions."""
+@dataclasses.dataclass(repr=True)
+class PartsError(Exception):
+    """Unexpected error.
+
+    :param brief: Brief description of error.
+    :param details: Detailed information.
+    :param resolution: Recommendation, if any.
+    """
+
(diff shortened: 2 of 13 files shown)


craft_parts/parts.py
```diff
@@ -0,0 +1,243 @@
...
+        """Return the part name."""
+        return self._name
+
+    @property
+    def properties(self) -> Dict[str, Any]:
+        """Return the part properties."""
+        return self._data.copy()
+
+    @property
+    def parts_dir(self) -> Path:
```
sergiusens, line 70: why is this necessary? Shouldn't this be handled a layer above?

sergiusens (changes requested):
Looks good, a few questions and maybe some improvements could be made to the defined exceptions

craft_parts/errors.py
```diff
@@ -0,0 +1,58 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""The exceptions that can be raised when using craft_parts."""
+
+from abc import ABC
+
+
+class CraftPartsError(Exception, ABC):
```
cjp256, line 22: Consider modelling it after ProviderError.
https://github.com/canonical/craft-providers/blob/main/craft_providers/errors.py#L74

Which borrows from the `SnapcraftException` pattern of brief/details/resolution.

craft_parts/parts.py
```diff
@@ -0,0 +1,243 @@
...
+    :param data: A dictionary containing the part properties.
+    :param project_dirs: The project work directories.
+    """
+
+    def __init__(
+        self,
+        name: str,
+        data: Dict[str, Any],
+        *,
+        project_dirs: ProjectDirs = None,
```
cjp256, line 46: I'm curious why project_dirs is required here (and not just part_dir). Can you open a draft PR of the working tree so I can see an up-to-date big picture? pretty plz :D

craft_parts/parts.py
```diff
@@ -0,0 +1,243 @@
...
+
+    :param name: The part name.
+    :param data: A dictionary containing the part properties.
+    :param project_dirs: The project work directories.
+    """
+
+    def __init__(
+        self,
+        name: str,
+        data: Dict[str, Any],
```
cjp256, line 44: I know it's probably asking too much right now, but I would recommend avoiding the data/dict pattern for the init.  Enumerate the properties with their sane defaults in the constructor.  add a @classmethod to unmarshal a dict (and perform any required schema/type checking).

craft_parts/parts.py
cmatsuoka: Yes, this shouldn't be here. It looks like a leftover from a previous refactoring (and it's not used in the rest of the PoC code).

craft_parts/dirs.py
```diff
@@ -0,0 +1,51 @@
...
+"""Definitions for project directories."""
+
+from pathlib import Path
+from typing import Union
+
+
+class ProjectDirs:
+    """The project main work directories.
+
+    :param work_dir: the parent directory containing the parts, prime and stage
```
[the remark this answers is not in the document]
cmatsuoka: According to PEP257 it seems so, with the general class description in the class docstring and the constructor description under init (which many try to avoid because the short description for init is always a bit silly). I'll do as described in the PEP, minus list of methods and variables, to see how Sphinx handles it.

craft_parts/errors.py
```diff
@@ -0,0 +1,58 @@
...
+    """Base class for Craft Parts exceptions."""
+
+    fmt = "Daughter classes should redefine this"
```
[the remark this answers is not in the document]
cmatsuoka: Error classes similar to what's used in craft-providers (as suggested by @cjp256 in a comment below) could be interesting, bringing more consistency to Craft packages. (but also more work to update all the errors currently in old snapcraft style).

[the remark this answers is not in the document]
sergiusens: Yes, what Chris recommends is great. I don't think you need to worry how errors bubble up as we are only going to use this in new applications for now. If the concern is another, then we can tech debt this.

craft_parts/parts.py
cmatsuoka: We'll have a set of `ProjectInfo`/`PartInfo`/`StepInfo` classes that are used to pass information to callbacks. `ProjectInfo` is created first and instantiates `ProjectDirs` from a given `work_dir`, then passes it to each `Part` instance so that the path of each project directory is defined only once -- if later we decided that each dir should be specified independently this would only affect `ProjectDirs`. If we create `Part` based only on `parts_dir` we would also need to redefine the paths in the `Part` initialization. The context will appear in the next PR which is currently being prepared, I'll provide a draft soon.

craft_parts/parts.py
cmatsuoka: The reason the original set of properties is kept in dictionary form is because state management filters properties of interest from a dict of properties, for each step (like snapcraft does). Except for that use, the class protects internal data using read-only accessors so refactoring it to get rid of the dict pattern and have all data stored in their own attributes shouldn't be a problem -- but we need a better solution for state building first.

craft_parts/dirs.py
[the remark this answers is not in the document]
cmatsuoka: Sphinx seems to prefer parameters documented in the class, contrary to what PEP257 recommends (but the PEP also recommends listing public methods and attributes in the class docstring which would result in redundancies in the processed docs). There are ways to make Sphinx concatenate both descriptions but the parameter list appears twice. Having the documentation formatted correctly is important, so unless @cjp256 or @facundobatista know a better way to handle this I'll leave it in the format that works for Sphinx.

craft_parts/errors.py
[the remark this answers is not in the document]
cmatsuoka: Updated to follow the craft-providers error format.

craft_parts/dirs.py
```diff
@@ -0,0 +1,51 @@
...
+        subdirectories. If not specified, the current directory will be used.
+    """
+
+    def __init__(self, *, work_dir: Union[Path, str] = "."):
```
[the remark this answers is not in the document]
cmatsuoka: `ProjectDirs` will not be created directly by the user so the default value shouldn't be a concern here. The default value is convenient for test code cleanliness so I'd like to keep it for now.

craft_parts/errors.py
cmatsuoka: Updated to follow the craft-providers error format (with minor implementation changes).

sergiusens (approved):

craft_parts/dirs.py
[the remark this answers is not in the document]
cjp256: Generally speaking from everything I've read about it, it's best to pick one way or the other.  You can make it work either way, and probably both after wrestling with sphinx and autodoc, but idk.  The class to constructor is a 1:1 mapping so I don't care :D

craft_parts/parts.py
cjp256: I like where your head is at.

cjp256 (approved):

craft_parts/parts.py
cjp256: OK thanks

craft_parts/dirs.py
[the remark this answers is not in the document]
cjp256: :+1:

PR #4, opened 2021-03-28
Add lifecycle manager and sequencer stub

cmatsuoka:
The lifecycle manager holds information about the parts specification
and the project state, and coordinates planning and execution of actions
required to run steps for a given set of parts.

-----


tests/unit/test_errors.py
```diff
@@ -54,3 +54,11 @@ def test_invalid_part_name():
     assert err.brief == "A part named 'foo' is not defined in the parts list."
     assert not err.details
     assert err.resolution == "Review the parts definition and make sure it's correct."
+
+
+def test_invalid_architecture():
+    err = errors.InvalidArchitecture("m68k")
+    assert err.arch_name == "m68k"
+    assert err.brief == "Architecture 'm68k' is not supported."
+    assert not err.details
```
sergiusens, line 63:
```suggestion
    assert err.details is None
```

sergiusens (commented):
Good start! Comments inline.

craft_parts/actions.py
```diff
@@ -0,0 +1,104 @@
...
+            and self._type == other._type
+            and self._reason == other._reason
+        )
+
+    def __repr__(self):
+        reason = f", {self._reason!r}" if self._reason else ""
+        return f"Action({self.part_name!r}, {self.step!r}, {self.type!r}{reason})"
+
+    @property
+    def part_name(self) -> str:
```
facundobatista, line 87: For this and the rest: why not just keep the attribute public and remove these getters?

If in the future we need to do some processing before returning the attribute, we can add them.

Or you're doing this because it's really a problem if somebody goes and change any of these attributes? Is that something that would happen and we really need to "fight it"?

craft_parts/infos.py
```diff
@@ -0,0 +1,211 @@
...
+    @property
+    def project_options(self) -> Dict[str, Any]:
+        """Obtain a project-wide options dictionary."""
+        return {
+            "application_name": self.application_name,
+            "arch_triplet": self.arch_triplet,
+            "target_arch": self.target_arch,
+        }
+
+    def _set_machine(self, target_arch):
```
facundobatista, line 122: Please add a docstring.

craft_parts/infos.py
```diff
@@ -0,0 +1,211 @@
...
+        self.__platform_arch = _get_platform_architecture()
```
facundobatista, line 123: Why starting with **double** underscore?

craft_parts/infos.py
```diff
@@ -0,0 +1,211 @@
...
+        if not target_arch:
+            target_arch = self.__platform_arch
+            logger.debug("Setting target machine to %s", target_arch)
+
+        machine = _ARCH_TRANSLATIONS.get(target_arch, None)
```
facundobatista, line 128: `None` is the default there in case `target_arch` is not in the dict.

craft_parts/infos.py
```diff
@@ -0,0 +1,211 @@
...
+        if not machine:
+            raise errors.InvalidArchitecture(target_arch)
+
+        self.__target_machine = target_arch
+        self.__machine_info = machine
+
+
+def _get_platform_architecture() -> str:
```
facundobatista, line 136: Please add a docstring.

craft_parts/infos.py
```diff
@@ -0,0 +1,211 @@
...
+    # TODO: handle Windows architectures
```
facundobatista, line 137: AFAIK `platform.machine()` also works in Windows.

craft_parts/parts.py
```diff
@@ -177,12 +177,39 @@ def get_scriptlet(self, step: Step) -> Optional[str]:
...
+    :returns: The list of parts corresponding to the given names.
+
+    :raises InvalidPartName: if a part name is not defined.
+    """
+    if names:
+        # check if all part names are valid
+        valid_names = [p.name for p in part_list]
+        for name in names:
+            if name not in valid_names:
+                raise errors.InvalidPartName(name)
```
facundobatista, line 196: It's easier to do...

```
        valid_names = {p.name for p in part_list}
        invalid_names = set(names) - valid_names
        if invalid_names:
            ...
```

I like this more because you don't end up with an `in` comparison inside a `for` loop.

facundobatista (changes requested):
Awesome! I added some comments inline, thanks!

craft_parts/actions.py
```diff
@@ -0,0 +1,104 @@
...
+from typing import Optional
+
+from craft_parts.steps import Step
+
+
+@enum.unique
+class ActionType(enum.IntEnum):
+    """The type of the action to be executed.
+
+    Action execution can be modified according to its type. An
```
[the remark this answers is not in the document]
cmatsuoka: Change to a list-style description. Sphinx could offer better support for enum class documentation.

craft_parts/errors.py
```diff
@@ -67,3 +67,17 @@ def __init__(self, part_name: str):
         resolution = "Review the parts definition and make sure it's correct."
 
         super().__init__(brief=brief, resolution=resolution)
+
+
+class InvalidArchitecture(PartsError):
+    """The machine architecture is not supported.
+
+    :param arch_name: the unsupported architecture name.
```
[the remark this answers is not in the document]
cmatsuoka: Ran a grep and found more instances of noncapitalized parameter descriptions. The doc style linter could include this check as well.

craft_parts/infos.py
```diff
@@ -0,0 +1,211 @@
...
+            project_dirs = ProjectDirs()
+
+        self._application_name = application_name
+        self._set_machine(arch)
+        self._plugin_version = plugin_version
+        self._parallel_build_count = parallel_build_count
+        self._dirs = project_dirs
+        self._custom_args = custom_args
+
+    def __getattr__(self, name):
```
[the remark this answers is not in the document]
cmatsuoka: There's a cascade of `ProjectInfo` -> `PartInfo` -> `StepInfo`, each one "inheriting" attributes from the previous data class (that could be done by subclassing but `Partinfo` is rather large and initializing the `StepInfo` superclass was a bit ugly). These classes will be passed to callbacks. Cascading getattrs are certainly more expensive than setattr, but the performance penalty should be negligible and avoids data duplication -- I'll investigate the impact, refactoring to use setattr should be easy to do in a follow-up if we decide it's a better way to handle this.

craft_parts/infos.py
```diff
@@ -0,0 +1,211 @@
...
+        return self._plugin_version
+
+    @property
+    def parallel_build_count(self) -> int:
+        """Return the maximum allowable number of concurrent build jobs."""
+        return self._parallel_build_count
+
+    @property
+    def target_arch(self) -> str:
+        """Return the architecture used for deb packages."""
```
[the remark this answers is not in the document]
cmatsuoka: If dealing with a platform using different architecture names, this also could return the native arch name for the platform, e.g. CentOS arch names. But there will be discussions on how to handle this so let's wait to determine how we'll implement it (and also the entire package/repo handling will have to support a different package format in this case).

craft_parts/manager.py
```diff
@@ -0,0 +1,116 @@
...
+    order to obtain a tree of installed files from the specification on how to
+    process its parts, and provides a mechanism to execute each of these actions.
+
+    :param all_parts: A dictionary containing the parts specification according
+        to the :ref:`parts schema<parts-schema>`. The format is compatible with the
+        output generated by PyYAML's ``yaml.load``.
+    :param application_name: An unique identifier for the application using Craft
+        Parts. This string will be used as a directory name when creating persistent
+        data that shouldn't be shared with other applications.
+    :param build_packages: A list of additional build packages to install.
```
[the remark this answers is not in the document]
cmatsuoka: This can be removed for now and added back when we have context for it.

craft_parts/actions.py
cmatsuoka: I'd like to be sure that these attributes won't be changed (either intentionally or by accident) by the user or even by ourselves. Many of these data structures are exposed to the application and a change there could be unpredictable and potentially hard to track down, so we're being very cautious.

craft_parts/infos.py
cmatsuoka: This code was transposed directly from Snapcraft and its style was not properly adjusted. As @sergiusens explained there is a reason for that in the original context, but not here. Fixing.

cmatsuoka: Also rename attributes to be more consistent and less confusing.

craft_parts/sequencer.py
```diff
@@ -0,0 +1,83 @@
...
+    :param part_list: The list of parts to process.
+    :param project_info: Information about this project.
+    """
+
+    def __init__(self, *, part_list: List[Part], project_info: ProjectInfo):
+        self._part_list = sort_parts(part_list)
+        self._project_info = project_info
+        self._actions: List[Action] = []
+
+    def plan(self, target_step: Step, part_names: List[str] = None) -> List[Action]:
```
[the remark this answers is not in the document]
cmatsuoka: Agree, we should follow the recommendations whenever possible (especially in the user-facing public API).

craft_parts/infos.py
cmatsuoka: We'll probably need extra handling here to tell that this is Windows along with the architecture. I think this will become something more like `_get_host_platform()` instead of a plain `_get_host_architecture()`.

craft_parts/infos.py
```diff
@@ -0,0 +1,211 @@
...
+    ):
+        if not project_dirs:
+            project_dirs = ProjectDirs()
+
+        self._application_name = application_name
+        self._set_machine(arch)
+        self._plugin_version = plugin_version
+        self._parallel_build_count = parallel_build_count
+        self._dirs = project_dirs
+        self._custom_args = custom_args
```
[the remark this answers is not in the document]
cmatsuoka: The current API just passes any extra keyword arguments to `LifecycleManager()` directly to infos so they can be used by user callbacks. I'll update the docstring for this parameter.

craft_parts/parts.py
```diff
@@ -177,7 +177,9 @@ def get_scriptlet(self, step: Step) -> Optional[str]:
         return self._data.get(scr[step])
 
 
-def part_list_by_name(names: Optional[List[str]], part_list: List[Part]) -> List[Part]:
+def part_list_by_name(
+    names: Optional[Sequence[str]], part_list: List[Part]
```
sergiusens, line 181: probably want part_list to be a Sequence as well

sergiusens (approved):

craft_parts/parts.py
cmatsuoka: Will adjust that in a follow-up, I want to check all the consequences in the rest of the code.

craft_parts/actions.py
```diff
@@ -0,0 +1,106 @@
...
+        *,
+        action_type: ActionType = ActionType.RUN,
+        reason: Optional[str] = None,
+    ):
+        self._part_name = part_name
+        self._step = step
+        self._type = action_type
+        self._reason = reason
+
+    def __eq__(self, other):
```
cjp256, line 76: Other than the pattern that facundo asked about, this class could be simplified to:
```
@dataclasses.dataclass
class Action:
  part_name: str
  step: Step
  action_type: ActionType = ActionType.RUN
  reason: Optional[str] = None
```

It would provide the `__eq__` and `__repr__` like you have here.


To provide the "protection" against external changes (which, imo i would not spend the cycles stressing about in this use case), you could freeze it after init:
```
class Action:
  part_name: str
  step: Step
  action_type: ActionType = ActionType.RUN
  reason: Optional[str] = None
```

craft_parts/infos.py
```diff
@@ -0,0 +1,209 @@
...
+class ProjectInfo:
+    """Project-level information containing project-specific fields.
+
+    :param application_name: A unique identifier for the application using
+        Craft Parts.
+    :param arch: The architecture to build for. Defaults to the host system
+        architecture.
+    :param parallel_build_count: The maximum number of concurrent jobs to be
+        used to build each part of this project.
+    :param plugin_version: The plugin API version. Currently only ``v2`` is
```
cjp256, line 38: Consider dropping the notion of plugin version until we have a need for it.  When we do need to start versioning things, perhaps we can/would do it differently (e.g. not at a project level).

craft_parts/infos.py
```diff
@@ -0,0 +1,209 @@
...
+
+
+class ProjectInfo:
+    """Project-level information containing project-specific fields.
+
+    :param application_name: A unique identifier for the application using
+        Craft Parts.
+    :param arch: The architecture to build for. Defaults to the host system
+        architecture.
+    :param parallel_build_count: The maximum number of concurrent jobs to be
```
cjp256, line 36: I'm sure this is just carried over to support existing code, but parallel_build_count and arch does not feel project-specific.

craft_parts/infos.py
```diff
@@ -0,0 +1,209 @@
...
+        used to build each part of this project.
+    :param plugin_version: The plugin API version. Currently only ``v2`` is
+        supported.
+    :param project_dirs: The project work directories.
+    :param custom_args: Any additional arguments defined by the application
+        when creating a :class:`LifecycleManager`.
```
cjp256, line 42: But LifecycleManager creates ProjectInfo?

tests/unit/test_actions.py
```diff
@@ -0,0 +1,58 @@
...
+def test_action_type():
+    assert f"{ActionType.RUN!r}" == "ActionType.RUN"
+    assert f"{ActionType.RERUN!r}" == "ActionType.RERUN"
+    assert f"{ActionType.UPDATE!r}" == "ActionType.UPDATE"
+    assert f"{ActionType.SKIP!r}" == "ActionType.SKIP"
+
+
+def test_action_representation():
+    action = Action("foo", Step.PULL, action_type=ActionType.SKIP, reason="is tired")
+    assert f"{action!r}" == "Action('foo', Step.PULL, ActionType.SKIP, 'is tired')"
```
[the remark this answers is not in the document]
cjp256: lol

tests/unit/test_infos.py
```diff
@@ -0,0 +1,101 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
```
cjp256, line 1: perhaps test_project_info.py ?

tests/unit/test_manager.py
```diff
@@ -0,0 +1,69 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
```
cjp256, line 1: perhaps test_lifecycle_manager.py?

craft_parts/infos.py
cjp256: What if we make the application responsible for creating both so that the pass-through args aren't needed?  I think we'd lose some static checks using this method way...

craft_parts/infos.py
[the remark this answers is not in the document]
cjp256: I'd suggest just being explicit about it, even if all are independent classes, avoiding obfuscation.

craft_parts/infos.py
facundobatista: thanks

craft_parts/infos.py
facundobatista: ok

craft_parts/parts.py
facundobatista: ack

facundobatista (approved):
Awesome, thanks!!

tests/unit/test_infos.py
cmatsuoka: This test file will also contain tests for `PartInfo` and `StepInfo` (which will also be added to `infos.py`).

tests/unit/test_manager.py
cmatsuoka: I think it's a better name, I also should rename `manager.py` to `lifecycle_manager.py`. This will be done in a separate PR.

craft_parts/actions.py
cmatsuoka: Ah, that's an interesting approach. Maybe this could work in the *Info classes as well? I'll update this in a follow-up PR.

craft_parts/infos.py
cmatsuoka: I'll remove this argument from the user-exposed API and keep it internal until we have a new plugin version.

craft_parts/infos.py
cmatsuoka: They're not strictly project-specific, but they remain constant for the entire project so I kept it under `ProjectInfo` to avoid having a different structure to pass them around.

PR #6, opened 2021-03-31
schemas: add parts schema and schema validator

cmatsuoka:
Validate the parts definition data against the schema. The validator is
largely ported from snapcraft with minor refactoring. Error handling and
`jsonschema.ValidationError` post-processing is also ported from snapcraft.

-----

### DIFF
--- a/craft_parts/data/schema/parts.json
+++ b/craft_parts/data/schema/parts.json
@@ -1,232 +1,9 @@
 {
     "$schema": "http://json-schema.org/draft-04/schema#",
-    "definitions": {
-        "grammar-string": {
-            "oneOf": [
-                {
-                    "type": "string",
-                    "usage": "<string>"
-                },
-                {
-                    "type": "array",
-                    "items": {
-                        "minitems": 1,
-                        "uniqueItems": true,
-                        "oneOf": [
-                            {
-                                "type": "object",
-                                "usage": "on <selector>[,<selector>...]:",
-                                "additionalProperties": false,
-                                "patternProperties": {
-                                    "^on\\s+.+$": {
-                                        "$ref": "#/definitions/grammar-string"
-                                    }
-                                }
-                            },
-                            {
-                                "type": "object",
-                                "usage": "to <selector>[,<selector>...]:",
-                                "additionalProperties": false,
-                                "patternProperties": {
-                                    "^to\\s+.+$": {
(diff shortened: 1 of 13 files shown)


craft_parts/data/schema/parts.json
```diff
@@ -0,0 +1,423 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "definitions": {
+        "grammar-string": {
```
cjp256, line 4: Did we agree that the grammar would be processed before calling craft-parts?

cjp256: I know that this is a port, but personally I would prefer that craft-parts be purely a python object/dataclass definition [in the future] and leave any json/yaml schema validation to the application.  (and most of this PR would not be required).  If sanity checks are required for craft-parts, they perhaps can/should be done on init.

cjp256 (dismissed):

craft_parts/data/schema/parts.json
```diff
@@ -0,0 +1,423 @@
...
+                    "description": "ppa path: e.g. 'canonical-kernel-team/unstable'"
+                }
+            },
+            "required": [
+                "type",
+                "ppa"
+            ],
+            "validation-failure": "{!r} is not properly configured PPA repository"
+        },
+        "system-username-scope": {
```
sergiusens, line 214: this only applies to snaps

craft_parts/data/schema/parts.json
sergiusens: grammar should not be part of craft-parts, it should receive processed ready to use entries

sergiusens (commented):
If schema can be removed from craft-parts, I am all for it

craft_parts/data/schema/parts.json
cmatsuoka: There are other definitions that are not used, will remove them all.

craft_parts/data/schema/parts.json
cmatsuoka: @cjp256 I like the idea of having a parts dataclass to avoid dicts. We can consider refactoring it that way before 1.0 and the bulk of this change would be the `Part` refactoring you suggested earlier, but dealing with plugin schema definitions can be tricky. Any idea on how to handle them?

cmatsuoka:
Superseded by PR #9.

PR #7, opened 2021-04-02
parts: refactor to remove stored dictionary data

cmatsuoka:
Refactor the parts definition to unmarshal parts specification data
instead of storing it. Validation should be done before populating the
spec object instead of relying on jsonschema validation.

This PR conflicts with PR #6.

-----

### DIFF
--- a/craft_parts/errors.py
+++ b/craft_parts/errors.py
@@ -89,7 +89,7 @@ class PartSpecificationError(PartsError):
     :param message: The error message.
     """
 
-    def __init__(self, part_name: str, message: str):
+    def __init__(self, *, part_name: str, message: str):
         self.part_name = part_name
         self.message = message
         brief = f"Part {part_name!r} definition error: {message}."
--- a/craft_parts/errors.py
+++ b/craft_parts/errors.py
@@ -81,3 +81,18 @@ class InvalidArchitecture(PartsError):
         resolution = "Make sure the architecture name is correct."
 
         super().__init__(brief=brief, resolution=resolution)
+
+
+class PartSpecificationError(PartsError):
+    """Parts were not correctly specified.
+
+    :param message: The error message.
+    """
+
+    def __init__(self, part_name: str, message: str):
+        self.part_name = part_name
+        self.message = message
+        brief = f"Part {part_name!r} definition error: {message}."
+        resolution = f"Review part {part_name!r} and make sure it's correct."
+
+        super().__init__(brief=brief, resolution=resolution)
--- a/craft_parts/parts.py
+++ b/craft_parts/parts.py
@@ -23,6 +23,7 @@ from typing import Any, Dict, List, Optional, Sequence, Set
 from craft_parts import errors, unmarshal
 from craft_parts.dirs import ProjectDirs
 from craft_parts.steps import Step
+from craft_parts.utils import formatting_utils
 
 
 @dataclass(frozen=True)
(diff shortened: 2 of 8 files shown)


craft_parts/unmarshal.py
```diff
@@ -0,0 +1,93 @@
...
+        return value
+
+    def pop_boolean(self, key: str, default: bool = False) -> bool:
+        """Pop an item and validates it as a boolean."""
+        value = self._data.pop(key, default)
+        if not isinstance(value, bool):
+            raise ValueError(f"{key!r} must be a boolean")
+        return value
+
+    def pop_list_str(self, key: str, default: List[str] = None) -> List[str]:
```
cjp256, line 59:
```suggestion
    def pop_list_str(self, key: str, default: Optional[List[str]] = None) -> List[str]:
```

craft_parts/unmarshal.py
```diff
@@ -0,0 +1,93 @@
...
+        self._data = data
+
+    def pop_string(self, key: str, default: str = "") -> str:
+        """Pop an item and validates it as a string."""
+        value = self._data.pop(key, default)
+        if not isinstance(value, str):
+            raise ValueError(f"{key!r} must be a string")
+        return value
+
+    def pop_optional_string(self, key: str, default: str = None) -> Optional[str]:
```
cjp256, line 38:
```suggestion
    def pop_optional_string(self, key: str, default: Optional[str] = None) -> Optional[str]:
```

tests/unit/test_unmarshal.py
```diff
@@ -0,0 +1,264 @@
...
+    "data,value,error",
+    [
+        [{}, "", None],
+        [{"test": ""}, "", None],
+        [{"test": "foo"}, "foo", None],
+        [{"test": False}, "", "'test' must be a string"],
+    ],
+)
+def test_unmarshal_string(data, value, error):
+    udata = unmarshal.DataUnmarshaler(data)
```
cjp256, line 32: I'd suggest breaking out the error scenarios into its own test(s) to reduce test complexity.

craft_parts/errors.py
```diff
@@ -81,3 +81,18 @@ def __init__(self, arch_name: str):
...
         super().__init__(brief=brief, resolution=resolution)
+
+
+class PartSpecificationError(PartsError):
+    """Parts were not correctly specified.
+
+    :param message: The error message.
+    """
+
+    def __init__(self, part_name: str, message: str):
```
cjp256, line 92: I have a strong personal preference for kwargs when the args are not intuitive :D 
```suggestion
    def __init__(self, *, part_name: str, message: str):
```

craft_parts/parts.py
```diff
@@ -16,15 +16,137 @@
...
+
+        :raise ValueError: If data validation fails.
+        """
+        if not isinstance(data, dict):
+            raise ValueError("part data is not a dictionary")
+
+        udata = unmarshal.DataUnmarshaler(data)
+
+        return cls(
+            plugin=udata.pop_optional_string("plugin"),
```
cjp256, line 76: I will say that seeing pop in unmarshal is a bit odd.  From a library level, the passed in dict is going to be deepcopy()'d before craft-parts modifies it, right?

cjp256: And the popping is just to ensure there are no extra fields after Parts and Plugin?

If the plugin properties were stripped first (i.e. `plugin: make`  filters out `make-*`, then unmarshal can sanity check by doing its own `data.copy()` and check for leftover fields before returning.

tests/unit/test_unmarshal.py
cmatsuoka: Reorganize tests to make them easier to read.

craft_parts/parts.py
cmatsuoka: Changing to make data consumption optional (by passing `consume=True` in `DataUnmarshaler` constructor) and rename `pop_*` methods to `get_*` to make the utility more generic, implementing the regular non-consuming case as the default. Then use the special consuming case in part specification unmarshal to detect if fields left without changing caller data, raising an error if unexpected fields found. Reorganize tests.

cmatsuoka:
Superseded by PR #9.

PR #8, opened 2021-04-05
utils: add formatting utils

cmatsuoka:
Port humanize_list and pluralize from Snapcraft formatting_utils.

Signed-off-by: Claudio Matsuoka <claudio.matsuoka@canonical.com>

- [x] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?

-----

PR #9, opened 2021-04-06
parts: refactor to remove stored dictionary data

cmatsuoka:
Refactor the parts definition to unmarshal parts specification data
instead of storing it. Data is validated using pydantic based on
Python type annotations.

This PR replaces PRs #6, #7, and #8.

-----


craft_parts/parts.py
```diff
@@ -16,15 +16,95 @@
...
+    source: Optional[str]
+    source_checksum: Optional[str] = ""
+    source_branch: Optional[str] = ""
+    source_commit: Optional[str] = ""
+    source_depth: Optional[int] = 0
+    source_subdir: Optional[str] = ""
+    source_tag: Optional[str] = ""
+    source_type: Optional[str] = ""
+    disable_parallel: Optional[bool] = False
+    after: Optional[List[str]] = []
```
cjp256, line 42: They make it work, but a default that appears mutable? yikes :D

cjp256: Given, `after: Optional[List[str]] = []`, I'd assume after can be None, yet the default is an empty list.

I'd go with the empty list default, and drop the Optional wherever the meaning is effectively the same.

This would avoid unnecessary handling like below:
```
 if not self.spec.after:
        return []
 return self.spec.after
```

craft_parts/parts.py
```diff
@@ -16,15 +16,95 @@
...
+
 from craft_parts import errors
 from craft_parts.dirs import ProjectDirs
 from craft_parts.steps import Step
 
 
+class PartSpec(BaseModel):
+    """The part specification data."""
+
+    plugin: Optional[str]
```
cjp256, line 32: I'd suggest being explicit about the default for the optionals.
```suggestion
    plugin: Optional[str] = None
```

Is plugin really optional?

craft_parts/parts.py
```diff
@@ -16,15 +16,95 @@
...
+    source: Optional[str]
+    source_checksum: Optional[str] = ""
```
cjp256, line 34: A lot of these empty quotes should probably be Nones?

craft_parts/parts.py
cmatsuoka: Pydantic handles list defaults in a sane way: https://stackoverflow.com/questions/63793662/how-to-give-a-pydantic-list-field-a-default-value (good). Unfortunately `Optional` becomes mandatory when unmarshaling with pydantic, otherwise it errors whenever the property is not listed in the input yaml (bad).

craft_parts/parts.py
cmatsuoka: Now it is. If plugin is not specified, we use the part name instead. This is used e.g. in the charmcraft `bundle` part.

craft_parts/parts.py
cmatsuoka: Possibly, I'll investigate. The empty string default for source-* properties predates the adoption of the unmarshal pattern.

cmatsuoka: Changed to use `None`.

cjp256 (approved):
LGTM - maybe just add it to the bucket list to turn the empty strings into Nones in the future

sergiusens (approved):

PR #10, opened 2021-04-07
states: add step state base class

cmatsuoka:
The context of a step is preserved to detect state changes between
executions and decide if new runs can be skipped. This change adds the
base class for step states.

-----


craft_parts/state_manager/step_state.py
```diff
@@ -0,0 +1,121 @@
...
+from abc import ABC, abstractmethod
+from pathlib import Path
+from typing import Any, Dict, Set
+
+from pydantic import Field
+from pydantic_yaml import YamlModel  # type: ignore
+
+
+class StepState(YamlModel, ABC):
+    """The context used to run a step for a part.
```
facundobatista, line 29: "is used"?

craft_parts/state_manager/step_state.py
```diff
@@ -0,0 +1,121 @@
...
+
+    The step state contains environmental and project-specific configuration
+    data collected at step run time. Those properties are used to decide whether
+    the step should run again on a new lifecycle execution.
+    """
+
+    part_properties: Dict[str, Any] = Field({}, alias="properties")
```
facundobatista, line 36: Why will we have two ways to refer to this? Wouldn't it make the learning more complex?

craft_parts/state_manager/step_state.py
```diff
@@ -0,0 +1,121 @@
...
+        :return: The newly created dictionary.
+        """
+        return self.dict(by_alias=True)
+
+    def write(self, filepath: Path) -> None:
+        """Write this state to disk."""
+        os.makedirs(filepath.parent, exist_ok=True)
+        yaml_data = self.yaml(by_alias=True)
+        with open(filepath, "w") as f:
+            f.write(yaml_data)
```
facundobatista, line 106: easier to just do `filepath.write_text(yaml_data)`

craft_parts/state_manager/step_state.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+def _get_differing_keys(dict1, dict2) -> Set[str]:
```
facundobatista, line 109: Please provide a docstring and extensive tests for this function.

Thanks!

craft_parts/state_manager/step_state.py
```diff
@@ -0,0 +1,121 @@
...
+    differing_keys = set()
+    for key, dict1_value in dict1.items():
+        dict2_value = dict2.get(key)
+        if dict1_value != dict2_value:
+            differing_keys.add(key)
```
facundobatista, line 114: For the following case, should `foobar` be included in the `differing_keys` set?

- dict1: `{'foobar': None}`
- dict2: `{}`

facundobatista (changes requested):
Nice! Added some comments inline, thanks!!!

craft_parts/state_manager/step_state.py
cmatsuoka: Changing to call it `part-properties` in the state file.

craft_parts/state_manager/step_state.py
cmatsuoka: It shouldn't, will document the rules.

craft_parts/state_manager/step_state.py
cmatsuoka: Rephrased that, we don't actually "use" a context to run a step.

craft_parts/state_manager/step_state.py
```diff
@@ -0,0 +1,125 @@
...
+    def marshal(self) -> Dict[str, Any]:
+        """Create a dictionary containing the part state data.
+
+        :return: The newly created dictionary.
+        """
+        return self.dict(by_alias=True)
+
+    def write(self, filepath: Path) -> None:
+        """Write this state to disk."""
+        os.makedirs(filepath.parent, exist_ok=True)
```
sergiusens, line 102:
```suggestion
        filepath.parent.mkdir(parents=True, exist_ok=True)
```

craft_parts/state_manager/step_state.py
```diff
@@ -0,0 +1,125 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""The step state preserves step execution context information."""
+
+import os
```
sergiusens, line 19:
```suggestion
```

sergiusens (approved):
Looks good, just one suggestion

sergiusens (approved):

craft_parts/state_manager/step_state.py
cjp256: _get_differing_keys can be simplified to:

diff_items = (dict1.items() - dict2.items()) | (dict2.items() - dict1.items())`
return [i[0] for i in diff_items]

craft_parts/state_manager/step_state.py
```diff
@@ -0,0 +1,124 @@
...
+        )
+
+    def marshal(self) -> Dict[str, Any]:
+        """Create a dictionary containing the part state data.
+
+        :return: The newly created dictionary.
+        """
+        return self.dict(by_alias=True)
+
+    def write(self, filepath: Path) -> None:
```
cjp256, line 99: In craft-providers i'm using `load()` classmethod and `save()` instance method for the names of marshaling/unmarshaling to/from disk.  I don't care what we pick, but we should probably pick terms and be consistent like with marshal/unmarshal. :D

cjp256 (approved):

craft_parts/state_manager/step_state.py
cmatsuoka: It's a good point, I'll check that and refactor in a follow-up.

craft_parts/state_manager/step_state.py
facundobatista: Only if all the values from both dicts are hashable. Signature type hints say values can be `Any`.

tests/unit/state_manager/test_step_state.py
```diff
@@ -0,0 +1,158 @@
...
+            [{}, {"b": 2}, {"b"}],
+            [{"a": 1}, {"b": 2}, {"a", "b"}],
+            [{"a": None}, {}, set()],
+            [{}, {"b": None}, set()],
+            [{"a": None}, {"b": None}, set()],
+            [{"a": 1}, {"a": 1}, set()],
+            [{"a": None}, {"a": 1}, {"a"}],
+            [{"a": 1}, {"a": 1, "b": 2}, {"b"}],
+            [{"a": 1, "b": 2}, {"a": 1}, {"b"}],
+        ],
```
facundobatista, line 155: Missing at least the case of same key with different not-None values.

facundobatista (approved):
Great work!! Approved, but please take a look to the comment in the test function.

Thanks!

tests/unit/state_manager/test_step_state.py
cmatsuoka: Adding this test.

PR #12, opened 2021-04-08
states: add concrete step states

cmatsuoka:
`PullState`, `BuildState`, `StageState` and `PrimeState` are the
concrete subclasses of `StepState`.

Note: this is stacked on top of PR #10, relevant commit is 6412b19.

-----


craft_parts/state_manager/step_state.py
```diff
@@ -0,0 +1,124 @@
...
+        return self.dict(by_alias=True)
+
+    def write(self, filepath: Path) -> None:
+        """Write this state to disk."""
+        filepath.parent.mkdir(parents=True, exist_ok=True)
+        yaml_data = self.yaml(by_alias=True)
+        filepath.write_text(yaml_data)
+
+
+def _get_differing_keys(dict1: Dict[str, Any], dict2: Dict[str, Any]) -> Set[str]:
```
cjp256, line 106: _get_differing_keys can be simplified to:

```
diff_items = (dict1.items() - dict2.items()) | (dict1.items() - dict2.items())`
return [i[o] for i in diff_items]
```

cjp256 (approved):

craft_parts/state_manager/states.py
```diff
@@ -0,0 +1,67 @@
...
+from .build_state import BuildState
+from .prime_state import PrimeState
+from .pull_state import PullState
+from .stage_state import StageState
+from .step_state import StepState
+
+logger = logging.getLogger(__name__)
+
+
+def load_state(part: Part, step: Step) -> Optional[StepState]:
```
cjp256, line 37: Perhaps `load(cls, state_file: pathlib.Path)` should be a classmethod in StepState to keep it together?

cjp256: fwiw I don't care what the names are, but to standardize a bit we should choose something like we have for marshal/unmarshal.

craft_parts/state_manager/step_state.py
cmatsuoka: The current implementation (that comes from snapcraft) handles items with value `None` as the equivalent of not having the key declared. I'll check the consequences of comparing without this rule and update if it works in our scenarios.

cmatsuoka: Since this call is always used after it passes the properties/options of interest filter, the `None` special case doesn't really matter and the implementation can be simplified as suggested.

cmatsuoka:
Converted to draft, this will be sliced into different PRs for each step class implementation.

cmatsuoka:
Replaced by PRs adding stage states separately.

PR #13, opened 2021-04-09
states: add pull state class

cmatsuoka:
The PullState class holds context information collected when the
pull step runs for a part.

This PR partially replaces PR #12.

-----

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
+PyYAML==5.4.1
 pydantic==1.8.1
 pydantic-yaml==0.2.3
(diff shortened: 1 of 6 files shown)


cjp256 (approved):

requirements.txt
```diff
@@ -1,2 +1,3 @@
+PyYAML==5.4.1
```
sergiusens, line 1: too late for this now, but I wonder if we can switch to https://pypi.org/project/strictyaml/

sergiusens (approved):

requirements.txt
cmatsuoka: I'll test strictyaml with craft-parts-orig parallel and refactor craft-parts if everything goes well.

PR #16, opened 2021-04-09
states: add prime state class and state loader

cmatsuoka:
The PrimeState class holds context information collected when the
prime step runs for a part. Also add a state loader helper that
reads persistent state based on part and state, unmarshaling data
into an appropriate state object.

This PR replaces PR #12.

-----

### DIFF
--- a/craft_parts/state_manager/states.py
+++ b/craft_parts/state_manager/states.py
@@ -51,15 +51,15 @@ def load_state(part: Part, step: Step) -> Optional[StepState]:
         state_data = yaml.safe_load(f)
 
     if step == Step.PULL:
-        state = PullState.unmarshal(state_data)
+        state_class = PullState
     elif step == Step.BUILD:
-        state = BuildState.unmarshal(state_data)
+        state_class = BuildState
     elif step == Step.STAGE:
-        state = StageState.unmarshal(state_data)
+        state_class = StageState
     else:
-        state = PrimeState.unmarshal(state_data)
+        state_class = PrimeState
 
-    return state
+    return state_class.unmarshal(state_data)
 
 
 def state_file_path(part: Part, step: Step) -> Path:
--- a/craft_parts/state_manager/states.py
+++ b/craft_parts/state_manager/states.py
@@ -1,0 +1,67 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2016-2021 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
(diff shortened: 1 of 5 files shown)


craft_parts/state_manager/states.py
```diff
@@ -0,0 +1,67 @@
...
+        state_data = yaml.safe_load(f)
+
+    if step == Step.PULL:
+        state = PullState.unmarshal(state_data)
+    elif step == Step.BUILD:
+        state = BuildState.unmarshal(state_data)
+    elif step == Step.STAGE:
+        state = StageState.unmarshal(state_data)
+    else:
+        state = PrimeState.unmarshal(state_data)
```
facundobatista, line 60: For less redundancy, in view of maybe a future change, what do you think about assigning just the class in each step, and then using it?

Something like:

```
if step == Step.PULL:
    state_class = PullState
elif:
    ....
else:
    state_class = PrimeStep
state = state_class.unmarshal(state_data)
```

pyproject.toml
```diff
@@ -17,6 +17,7 @@ max-line-length = "88"
 max-attributes = 15
 max-args= 6
 max-locals = 16
+good-names = "f"
```
facundobatista, line 20: This is for it doesn't complain about you calling `f` the file handler? What about just calling it `fh`?

facundobatista (approved):
Nice!! Approving, but please check the inline comments.

pyproject.toml
cmatsuoka: It wants a name with at least 3 characters. I could use `file` (which seems ok to use in Python 3) but since it was used by Python in the past and my syntax highlighter still colors it differently, just using the canonical variable `f` looked more appropriate.

craft_parts/state_manager/states.py
cmatsuoka: This is nice, will do it.

craft_parts/state_manager/states.py
```diff
@@ -0,0 +1,67 @@
...
+
+    :return: The step state.
+    """
+    filename = state_file_path(part, step)
+    if not filename.is_file():
+        return None
+
+    logger.debug("load state file: %s", filename)
+    with open(filename) as f:
+        state_data = yaml.safe_load(f)
```
sergiusens, line 51:
```suggestion
    with filename.open() as yaml_file:
        state_data = yaml.safe_load(yaml_file)
```

pyproject.toml
sergiusens, line 20: If you get more verbose with the name as my suggestion, you can probably remove this line...
```suggestion
```

sergiusens (approved):
approved with comments

pyproject.toml
cmatsuoka: Or I can use `yaml_file` as @sergiusens suggested below, will do it.

PR #18, opened 2021-04-13
states: add stage manager class

cmatsuoka:
The State Manager initializes and keeps track of states for each part
and step, telling the lifecycle sequencer whether a step should run
for any given part.

This PR starts the implementation of `StateManager`, more methods
will be added in future PRs.
    
-----

### DIFF
--- a/tests/unit/state_manager/test_state_manager.py
+++ b/tests/unit/state_manager/test_state_manager.py
@@ -15,10 +15,14 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import dataclasses
+import time
+from pathlib import Path
 
 import pytest
 
-from craft_parts.state_manager import state_manager, states
+from craft_parts.infos import ProjectInfo
+from craft_parts.parts import Part
+from craft_parts.state_manager import StateManager, state_manager, states
 from craft_parts.steps import Step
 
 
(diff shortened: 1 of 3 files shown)


tests/unit/state_manager/test_state_manager.py
```diff
@@ -172,3 +176,79 @@ def test_set_step_updated(self):
...
+        p2 = Part("bar", {})
+
+        s1 = states.PullState()
+        s2 = states.BuildState()
+        s3 = states.PullState()
+        s4 = states.PrimeState()
+
+        # create state files
+        s4.write(Path("parts/bar/state/prime"))
+        time.sleep(0.1)
```
cjp256, line 241: is the resolution on the timestamp not good enough? :D

cjp256 (approved):

tests/unit/state_manager/test_state_manager.py
cmatsuoka: I'm using `st_mtime_ns` which should return nanoseconds in systems supporting it. Ext4 has ns in timestamps, while NTFS has a 0.1us resolution. According to the [`st_mtime_ns` documentation](https://docs.python.org/3/library/os.html#os.stat_result.st_mtime_ns) FAT32 on Windows has a 2-second (!) resolution, so it's unsupported.

tests/unit/state_manager/test_state_manager.py
```diff
@@ -172,3 +176,79 @@ def test_set_step_updated(self):
...
+        s3 = states.StageState()
+        s3.write(Path("parts/p3/state/stage"))
+
+        # only p3:stage has a state file when the stage manager is created
+        sm = StateManager(project_info=info, part_list=[p1, p2, p3])
+
+        for part in [p1, p2, p3]:
+            for step in list(Step):
+                ran = sm.has_step_run(part, step)
+                assert ran == (part == p3 and step == Step.STAGE)
```
sergiusens, line 200: if only step 3 has run, then p1 and p2 should have an assertion that they have not

cmatsuoka: The nested for loops iterate over all steps for all parts, and the assertion tests them all against False, except for the step that statisfies the condition (part p3 and step stage), which is asserted True.

sergiusens (approved):

PR #20, opened 2021-04-13
infos: add part information class

cmatsuoka:
The part information class holds information to be passed to pre and
post-step callbacks for user environment setup.

-----

### DIFF
--- a/craft_parts/infos.py
+++ b/craft_parts/infos.py
@@ -163,6 +163,8 @@ class PartInfo:
         self._part_state_dir = part.part_state_dir
 
     def __getattr__(self, name):
+        # Use composition and attribute cascading to avoid setting attributes cumulatively
+        # in the init method.
         if hasattr(self._project_info, name):
             return getattr(self._project_info, name)
 
--- a/craft_parts/infos.py
+++ b/craft_parts/infos.py
@@ -18,10 +18,12 @@
 
 import logging
 import platform
+from pathlib import Path
 from typing import Any, Dict, List, Optional
 
 from craft_parts import errors, utils
 from craft_parts.dirs import ProjectDirs
+from craft_parts.parts import Part
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


craft_parts/infos.py
```diff
@@ -139,6 +141,69 @@ def _set_machine(self, arch: Optional[str]):
...
+        self._project_info = project_info
+        self._part_name = part.name
+        self._part_src_dir = part.part_src_dir
+        self._part_src_subdir = part.part_src_subdir
+        self._part_build_dir = part.part_build_dir
+        self._part_build_subdir = part.part_build_subdir
+        self._part_install_dir = part.part_install_dir
+        self._part_state_dir = part.part_state_dir
+
+    def __getattr__(self, name):
```
sergiusens, line 165: please document why, this is strange :-)

cmatsuoka: The final goal is to make `ProjectInfo` attributes available to `PartInfo`. The *Info classes use composition with attribute cascading to avoid setting a lot of attributes in the constructor (and missing something in the process). I can also implement it that way if you think inheritance is a better choice here. There will be one more class, `StepInfo`, that adds to that sequence (`ProjectInfo` -> `PartInfo` -> `StepInfo`). The info classes are not used internally and are just a way to feed relevant data to pre/post-step callbacks.

cjp256: Please do use inheritance :D It doesn't matter if there are a lot of attrs.

sergiusens:
El mar., 13 de abril de 2021 18:08, Chris Patterson <
***@***.***> escribió:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In craft_parts/infos.py
> <https://github.com/canonical/craft-parts/pull/20#discussion_r612777841>:
>
> > +
> +    def __init__(
> +        self,
> +        project_info: ProjectInfo,
> +        part: Part,
> +    ):
> +        self._project_info = project_info
> +        self._part_name = part.name
> +        self._part_src_dir = part.part_src_dir
> +        self._part_src_subdir = part.part_src_subdir
> +        self._part_build_dir = part.part_build_dir
> +        self._part_build_subdir = part.part_build_subdir
> +        self._part_install_dir = part.part_install_dir
> +        self._part_state_dir = part.part_state_dir
> +
> +    def __getattr__(self, name):
>
> Please do use inheritance :D It doesn't matter if there are a lot of attrs.
>

I personally prefer composition here

cmatsuoka:
I'm investigating the feasibility of implementing *Infos using dataclasses.

cmatsuoka:
One reason to prefer a flat structure is when you need to access e.g. the application name from `StepInfo` in a callback:

Using a flat structure: `name = info.application_name`
Using regular composition: `name = info.part_info.project_info.application_name`

cmatsuoka:
Agree that composition is better than inheritance in this case, we don't want to re-initialize project variables at each new step.

cjp256:
> Agree that composition is better than inheritance in this case, we don't want to re-initialize project variables at each new step.

Same, ignore my earlier statement :D

sergiusens (approved):

cjp256 (approved):

PR #21, opened 2021-04-14
states: add outdated step verification

cmatsuoka:
A part step is considered outdated if one of previous steps ran more
recently, or if logic in the source handler for that part says so.
This change implements the step execution order verification,
leaving source checks to be added later (after source handlers are
added to the codebase).

-----

### DIFF
--- a/craft_parts/state_manager/reports.py
+++ b/craft_parts/state_manager/reports.py
@@ -1,0 +1,61 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2018-2021 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Provide a report on why a step is outdated."""
+
+from craft_parts.steps import Step
+from craft_parts.utils import formatting_utils
+
+
+class OutdatedReport:
+    """The OutdatedReport class explains why a given step is outdated.
+
+    An outdated step is defined to be a step that has run, but since doing so
+    one of the following things have happened:
+
+    - A step earlier in the lifecycle has run again.
+    - The source on disk has been updated.
+    """
+
+    def __init__(
+        self, *, previous_step_modified: Step = None, source_updated: bool = False
+    ) -> None:
+        """Create a new OutdatedReport.
+
+        :param previous_step_modified: Step earlier in the lifecycle that has changed.
+        :param source_updated: Whether or not the source changed on disk.
+        """
+        self.previous_step_modified = previous_step_modified
+        self.source_updated = source_updated
+
+    def reason(self) -> str:
+        """Get summarized report.
+
+        :return: Short summary of why the step is dirty.
+        :rtype: str
+        """
+        reasons = []
+
+        if self.previous_step_modified:
+            reasons.append("{!r} step".format(self.previous_step_modified.name))
+
+        if self.source_updated:
+            reasons.append("source")
+
+        if not reasons:
+            return ""
+
+        return "{} changed".format(formatting_utils.humanize_list(reasons, "and", "{}"))
--- a/craft_parts/state_manager/reports.py
+++ b/craft_parts/state_manager/reports.py
@@ -31,28 +31,27 @@ class OutdatedReport:
     """
 
     def __init__(
-        self, *, previous_step_modified: Step = None, source_updated: bool = False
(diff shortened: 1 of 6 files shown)


craft_parts/state_manager/reports.py
```diff
@@ -0,0 +1,61 @@
...
+    - The source on disk has been updated.
+    """
+
+    def __init__(
+        self, *, previous_step_modified: Step = None, source_updated: bool = False
+    ) -> None:
+        """Create a new OutdatedReport.
+
+        :param previous_step_modified: Step earlier in the lifecycle that has changed.
+        :param source_updated: Whether or not the source changed on disk.
```
cjp256, line 39: imo, "changed" or "modified" would be preferable

craft_parts/state_manager/reports.py
```diff
@@ -0,0 +1,61 @@
...
+        """
+        self.previous_step_modified = previous_step_modified
+        self.source_updated = source_updated
+
+    def reason(self) -> str:
+        """Get summarized report.
+
+        :return: Short summary of why the step is dirty.
+        :rtype: str
```
cjp256, line 48:
```suggestion
```

tests/unit/state_manager/test_state_manager.py
```diff
@@ -222,6 +222,94 @@ def test_set_state(self):
...
+    def test_should_step_run_outdated(self):
+        info = ProjectInfo()
+        p1 = Part("p1", {})
+
+        # p1 build already ran
+        s1 = states.BuildState()
+        s1.write(Path("parts/p1/state/build"))
+
+        # but p1 pull ran more recently
+        time.sleep(0.1)
```
cjp256, line 256:
```suggestion
        time.sleep(0.001)
```
or something that meets the required granularity :D

craft_parts/state_manager/state_manager.py
```diff
@@ -192,6 +193,65 @@ def has_step_run(self, part: Part, step: Step) -> bool:
...
+        ):
+            return True
+
+        previous_steps = step.previous_steps()
+        if previous_steps:
+            return self.should_step_run(part, previous_steps[-1])
+
+        return False
+
+    def outdated_report(self, part: Part, step: Step) -> Optional[OutdatedReport]:
```
cjp256, line 224:
```suggestion
    def check_if_outdated(self, part: Part, step: Step) -> Optional[OutdatedReport]:
```

craft_parts/state_manager/reports.py
```diff
@@ -0,0 +1,61 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Provide a report on why a step is outdated."""
+
+from craft_parts.steps import Step
+from craft_parts.utils import formatting_utils
+
+
+class OutdatedReport:
```
cjp256, line 23: With its usage, would this make more sense as an exception?

cjp256 (approved):

craft_parts/state_manager/reports.py
cmatsuoka: I'll update the descriptions, they come from snapcraft.

craft_parts/state_manager/reports.py
cmatsuoka: I considered it some time ago but given how it's used during the lifecycle processing I prefer the current form (the lifecycle logic is not exception-driven, and handling dirty/outdated steps as exceptions didn't feel right)

tests/unit/state_manager/test_state_manager.py
cmatsuoka: Probably any small interval in the millisecond range would work. Will decrease to the suggested value.

craft_parts/state_manager/reports.py
cmatsuoka: Also leftovers from snapcraft. Will remove.

cmatsuoka: Also updating the return description, the step is outdated and not dirty.

tests/unit/state_manager/test_state_manager.py
cmatsuoka: Interestingly enough, something is not right when you use intervals < 5ms, as you can see in the code vs the actual mtime timestamps from the files:
```
    s4.write(Path("parts/bar/state/prime"))                                                                                                                                              
    time.sleep(0.001)                                                                                                                                                                    
    s3.write(Path("parts/bar/state/pull"))                                                                                                                                               
    time.sleep(0.001)                                                                                                                                                                    
    s1.write(Path("parts/foo/state/pull"))                                                                                                                                               
    time.sleep(0.001)                                                                                                                                                                    
    s2.write(Path("parts/foo/state/build"))             

 foo Step.PULL   1618444454171108403                                                        
 foo Step.BUILD  1618444454171108403                                                                                                                                                      
 bar Step.PULL   1618444454167108375    
 bar Step.PRIME  1618444454167108375
```
Either Python's `time.sleep()` call is not very accurate, or actual file writing is being delayed somewhere and happening in bursts. After a short investigation I realized that I'm running it on Ubuntu Server with HZ=250, which gives me 4ms resolution (matching the 167 to 171ms interval in the timestamps). So I'll use the more conservative value of 20ms, which should be more than enough for systems running with HZ=100.

sergiusens (approved):

PR #22, opened 2021-04-15
plugins: add plugin and plugin properties base classes

cmatsuoka:
The plugin subsystem defines how part sources will be built, using
optional plugin-specific properties defined in the part specification.

-----

### DIFF
--- a/craft_parts/plugins/base.py
+++ b/craft_parts/plugins/base.py
@@ -17,7 +17,7 @@
 """Plugin base class and definitions."""
 
 import abc
-from typing import Dict, List, Optional, Set, Type
+from typing import Dict, List, Set, Type
 
 from craft_parts.infos import PartInfo
 
@@ -29,20 +29,15 @@ class Plugin(abc.ABC):
 
     :cvar properties_class: The plugin properties class.
 
-    :param part_info: the part information for the applicable part.
-    :param options: an object representing part defined properties.
+    :param part_info: The part information for the applicable part.
+    :param properties: Part-defined properties.
     """
 
     properties_class: Type[PluginProperties]
 
-    def __init__(
-        self, *, options: Optional[PluginProperties], part_info: PartInfo
-    ) -> None:
-        if not options:
-            options = PluginProperties()
-
+    def __init__(self, *, properties: PluginProperties, part_info: PartInfo) -> None:
         self._name = part_info.part_name
-        self._options = options
+        self._options = properties
         self._part_info = part_info
 
     @abc.abstractmethod
--- a/craft_parts/plugins/base.py
+++ b/craft_parts/plugins/base.py
@@ -29,7 +29,7 @@ class Plugin(abc.ABC):
 
     :cvar properties_class: The plugin properties class.
 
-    :param part_name: the name of the part this plugin is instantiated to.
+    :param part_info: the part information for the applicable part.
     :param options: an object representing part defined properties.
     """
 
@@ -58,7 +58,7 @@ class Plugin(abc.ABC):
         """Return a dictionary with the environment to use in the build step."""
 
     @property
-    def out_of_source_build(self):
+    def out_of_source_build(self) -> bool:
         """Return whether the plugin performs out-of-source-tree builds."""
         return False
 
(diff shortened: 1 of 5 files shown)


craft_parts/plugins/base.py
```diff
@@ -0,0 +1,67 @@
...
+
+from .properties import PluginProperties
+
+
+class Plugin(abc.ABC):
+    """The base class for plugins.
+
+    :cvar properties_class: The plugin properties class.
+
+    :param part_name: the name of the part this plugin is instantiated to.
```
cjp256, line 32:
```suggestion
    :param part_info: the part information for the applicable part.
```

craft_parts/plugins/properties.py
```diff
@@ -0,0 +1,37 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Definitions and helpers for plugin options."""
+
+from typing import Any, Dict
+
+
+class PluginProperties:
```
cjp256, line 22: Should this be a dataclass?

cjp256 (approved):

craft_parts/plugins/base.py
```diff
@@ -0,0 +1,67 @@
...
+    @abc.abstractmethod
+    def get_build_packages(self) -> Set[str]:
+        """Return a set of required packages to install in the build environment."""
+
+    @abc.abstractmethod
+    def get_build_environment(self) -> Dict[str, str]:
+        """Return a dictionary with the environment to use in the build step."""
+
+    @property
+    def out_of_source_build(self):
```
cjp256, line 61:
```suggestion
    def out_of_source_build(self) -> bool:
```

craft_parts/plugins/properties.py
cmatsuoka: It has no data, but the subclasses implemented in our plugins are. Application-defined plugins have the choice to decide how to implement their property class (dataclass or not) as long as they subclass `PluginProperties`.

craft_parts/plugins/base.py
```diff
@@ -0,0 +1,62 @@
...
+
+import abc
+from typing import Dict, List, Set, Type
+
+from craft_parts.infos import PartInfo
+
+from .properties import PluginProperties
+
+
+class Plugin(abc.ABC):
```
facundobatista, line 27: Please add tests for this, thanks!

facundobatista (changes requested):
Please add tests for the base Plugin class, thanks!

craft_parts/plugins/properties.py
sergiusens: should this not be an abc and unmarshal be an abstractmethod?

cmatsuoka: The empty plugin property class is to be instantiated when a part is created without specifying a plugin properties object. (This is mainly to ensure part consistency in tests, where declaring plugin properties for each test involving part creation is very inconvenient)

craft_parts/plugins/base.py
```diff
@@ -0,0 +1,62 @@
...
+    def get_build_snaps(self) -> Set[str]:
+        """Return a set of required snaps to install in the build environment."""
+
+    @abc.abstractmethod
+    def get_build_packages(self) -> Set[str]:
+        """Return a set of required packages to install in the build environment."""
+
+    @abc.abstractmethod
+    def get_build_environment(self) -> Dict[str, str]:
+        """Return a dictionary with the environment to use in the build step."""
```
facundobatista, line 53: I removed these three methods, and tests still pass.

Is there value in having these definitions? if yes, we should test them, if not, we should remove them.

facundobatista (changes requested):

craft_parts/plugins/base.py
sergiusens: You have to instantiate for the abstractness to kick in

```
>>> class Plugin(abc.ABC):
...     @abc.abstractmethod
...     def foo(self):
...         pass
... 
>>> class P(Plugin):
...     pass
... 
>>> P()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: Can't instantiate abstract class P with abstract methods foo
```

sergiusens (approved):

facundobatista (approved):
Wonderful, thanks!

PR #25, opened 2021-04-19
sequencer: handle step run, rerun and update

cmatsuoka:
Lifecycle planning requires updating step states and emmiting the
appropriate actions to be executed later. This is done in different
ways depending if we're running a step for the first time, re-running
it, or updating it. This change adds the appropriate handlers for
each case.

-----

### DIFF
--- a/craft_parts/sequencer.py
+++ b/craft_parts/sequencer.py
@@ -140,13 +140,16 @@ class Sequencer:
                 directories=set(),
             )
 
-        else:
+        elif step == Step.PRIME:
             state = states.PrimeState(
                 part_properties=part_properties,
                 project_options=self._project_info.project_options,
                 files=set(),
                 directories=set(),
             )
+
+        else:
+            raise RuntimeError(f"invalid step {step!r}")
 
         self._sm.set_state(part, step, state=state)
 
--- a/craft_parts/sequencer.py
+++ b/craft_parts/sequencer.py
@@ -17,11 +17,13 @@
 """Determine the sequence of lifecycle actions to be executed."""
 
 import logging
-from typing import List, Optional, Sequence
+from typing import Dict, List, Optional, Sequence
 
-from craft_parts.actions import Action
+from craft_parts import parts, steps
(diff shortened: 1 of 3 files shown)


craft_parts/sequencer.py
```diff
@@ -81,3 +84,103 @@ def _add_step_actions(
...
         pass
+
+    def _process_dependencies(self, part: Part, step: Step) -> None:
+        prerequisite_step = steps.dependency_prerequisite_step(step)
+        if not prerequisite_step:
+            return
+
+        all_deps = parts.part_dependencies(part.name, part_list=self._part_list)
+
+        deps = {p for p in all_deps if self._sm.should_step_run(p, prerequisite_step)}
```
cjp256, line 95: Does order matter here?

cmatsuoka: It shouldn't because if A depends on B and C, and B depends on C, B was already processed (parts are sorted by dependency in the Sequencer constructor) so C shouldn't run again. This is also the case in Snapcraft (corresponding code is in `snapcraft.internal.lifecycle._runner._Executor._prepare_step`).

cjp256: Thanks!

craft_parts/sequencer.py
```diff
@@ -81,3 +84,103 @@ def _add_step_actions(
...
+
+        elif step == Step.STAGE:
+            state = states.StageState(
+                part_properties=part_properties,
+                project_options=self._project_info.project_options,
+                files=set(),
+                directories=set(),
+            )
+
+        else:
```
sergiusens, line 143: should be elif, consider we may add new steps, else should raise a RuntimeError

sergiusens (changes requested):

craft_parts/sequencer.py
cmatsuoka: Raising RuntimeError in case of invalid step. Added similar handling to state loading for consistency.

sergiusens (approved):

cjp256 (approved):

PR #28, opened 2021-04-20
sources: add simple file cache

cmatsuoka:
Sources with the `source-checksum` property are cached so they won't
be re-downloaded. This change implements a simple file cache that
stores such files under their digest key.

-----

### DIFF
--- a/craft_parts/sources/cache.py
+++ b/craft_parts/sources/cache.py
@@ -16,7 +16,6 @@
 
 """Cache base and file cache."""
 
-import abc
 import logging
 import os
 import shutil
@@ -27,16 +26,7 @@ from xdg import BaseDirectory  # type: ignore
 logger = logging.getLogger(__name__)
 
 
-class Cache(abc.ABC):
-    """Generic cache base class."""
-
-    def __init__(self, name: str):
-        self.cache_root = os.path.join(
-            BaseDirectory.xdg_cache_home, name, "craft-parts"
-        )
-
-
-class FileCache(Cache):
+class FileCache:
     """Cache files based on the supplied key."""
 
     def __init__(self, name: str, *, namespace: str = "files") -> None:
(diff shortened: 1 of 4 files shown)


sergiusens (approved):

craft_parts/sources/cache.py
```diff
@@ -0,0 +1,85 @@
...
+from xdg import BaseDirectory  # type: ignore
+
+logger = logging.getLogger(__name__)
+
+
+class Cache(abc.ABC):
+    """Generic cache base class."""
+
+    def __init__(self, name: str):
+        self.cache_root = os.path.join(
```
sergiusens, line 34: maybe use pathlib

craft_parts/sources/cache.py
```diff
@@ -0,0 +1,85 @@
...
+import os
+import shutil
+from typing import Optional
+
+from xdg import BaseDirectory  # type: ignore
+
+logger = logging.getLogger(__name__)
+
+
+class Cache(abc.ABC):
```
sergiusens, line 30: you probably you don't need this

craft_parts/sources/cache.py
cmatsuoka: This is a good thing to do. I tried it prior to opening this PR but it impacts other subsystems so I suggest we do it in a separate refactoring PR at a later moment.

craft_parts/sources/cache.py
cmatsuoka: Indeed, flattening the structure.

craft_parts/sources/cache.py
```diff
@@ -0,0 +1,77 @@
...
+        :param key: The key to cache the file under.
+
+        :return: The path to the cached file, or None if the file was not cached.
+        """
+        cached_file_path = os.path.join(self.file_cache, key)
+        os.makedirs(os.path.dirname(cached_file_path), exist_ok=True)
+
+        try:
+            if not os.path.isfile(cached_file_path):
+                shutil.copyfile(filename, cached_file_path)
```
cjp256, line 55: not very efficient for a cache :thinking:

cjp256 (approved):
I'm not sure what the exact use case for this cache is, but making some assumptions, I'd think it'd be better to just do the symlink of the digest when a file is downloaded, and the fetcher can check for it if the checksum is known.

craft_parts/sources/cache.py
cmatsuoka: The original code in snapcraft explicitly makes a copy instead of using `link_or_copy`, but I really can't see a reason why this shouldn't be used. Will change the code to allow hard links.

PR #31, opened 2021-04-21
sources: add source handler base classes

cmatsuoka:
Source handlers take care of copying, downloading or unpacking files
specified as part sources. This change adds the base class for source
handlers and downloadable file source handlers.

-----


craft_parts/sources/base.py
```diff
@@ -0,0 +1,217 @@
...
+    """
+
+    # pylint: disable=too-many-arguments
+
+    def __init__(
+        self,
+        source: Union[str, Path],
+        source_dir: Union[str, Path],
+        *,
+        application_name: str = None,
```
cjp256, line 51:
```suggestion
        application_name: Optional[str] = None,
```

and so on.

craft_parts/sources/base.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+class FileSourceHandler(SourceHandler):
+    """Base class for file source types."""
+
+    # pylint: disable=too-many-arguments
+    def __init__(
+        self,
+        source: Union[str, Path],
+        source_dir: Union[str, Path],
```
cjp256, line 117:
```suggestion
        part_src_dir: Union[str, Path],
```

tests/unit/sources/test_base.py
```diff
@@ -0,0 +1,227 @@
...
+        assert self.source.provision_dst == "source/dir"
+        assert self.source.provision_clean_target is False
+        assert self.source.provision_keep is False
+        assert self.source.provision_src == "source/dir/some_file"
+
+        downloaded = Path(new_dir, "source", "dir", "some_file")
+        assert downloaded.is_file()
+        assert downloaded.read_bytes() == b"content"
+
+    def test_file_source_abstract_methods(self):
```
cjp256, line 215: This is just testing python semantics?

If the purpose is to verify abstract methods, perhaps just
`assert FileSourceHandler.__abstractmethods__ == {"method1...",}`

cjp256 (approved):

tests/unit/sources/test_base.py
```diff
@@ -0,0 +1,227 @@
...
+        with pytest.raises(errors.SourceUpdateUnsupported) as raised:
+            self.source.check_if_outdated("/some/file")
+        assert raised.value.name == "FooSourceHandler"
+
+    def test_source_update(self):
+        with pytest.raises(errors.SourceUpdateUnsupported) as raised:
+            self.source.update()
+        assert raised.value.name == "FooSourceHandler"
+
+    def test_source_abstract_methods(self):
```
cjp256, line 61: Instead:
`assert SourceHandler.__abstractmethods__ == {"method1",...}

sergiusens (approved):
Approving on the basis that this is tech debt to improve

craft_parts/sources/base.py
cmatsuoka: `source_dir` is indeed a bad name, it looks like it's a part property when it's actually the part source directory. Will rename as suggested.

craft_parts/sources/base.py
cmatsuoka: It increases character density in the parameter list but it's more consistent with the optional parameter definition.

tests/unit/sources/test_base.py
cmatsuoka: We just wanted to make very explicit that subclasses not implementing those methods won't be allowed, but yes, just checking the abstract methods should have been enough.

PR #33, opened 2021-04-21
plugins: add make plugin

cmatsuoka:
The make plugin is useful to build make-based parts.

-----

### DIFF
--- a/craft_parts/plugins/base.py
+++ b/craft_parts/plugins/base.py
@@ -17,7 +17,9 @@
 """Plugin base class and definitions."""
 
 import abc
-from typing import Dict, List, Set, Type
+from typing import Any, Dict, List, Set, Type
+
+from pydantic import BaseModel
 
 from craft_parts.infos import PartInfo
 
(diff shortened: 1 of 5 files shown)


craft_parts/plugins/base.py
```diff
@@ -60,3 +62,40 @@ def out_of_source_build(self) -> bool:
...
+) -> Dict[str, Any]:
+    """Obtain plugin-specifc entries from part properties.
+
+    :param data: A dictionary containing all part properties.
+    :plugin_name: The name of the plugin.
+
+    :return: A dictionary with plugin properties.
+    """
+    plugin_data: Dict[str, Any] = {}
+    prefix = f"{plugin_name}-"
```
cjp256, line 95: So apparently I lied when I said this was true for v2. The python plugin has 'requirements' and 'constraints' that don't follow this model, and perhaps there are others...

tests/integration/plugins/test_make.py
```diff
@@ -0,0 +1 @@
+# TODO: add make plugin integration test after executor lands
```
cjp256, line 1: craft-providers `Executor` landed first lol

craft_parts/plugins/base.py
cjp256: Might be worth just dropping this generalization and make each plugin specify the list explicitly.

cjp256 (approved):

craft_parts/plugins/base.py
sergiusens: the reason is legacy, it is nice to know what is consumed by the plugin and what is consumed by the rest of the machinery

sergiusens (approved):

craft_parts/plugins/base.py
cmatsuoka: I verified current v2 snapcraft plugins, `requirements` and `constraints` in the python plugin are the only properties that violate the name prefix rule. Craft-parts could establish that as a hard rule (and legacy cases could be pre-processed if needed?)

cjp256: Seems fair.  It looks like it's under control of the plugins anyhow so it should be OK even as-is?

PR #37, opened 2021-04-22
sequencer: add lifecycle logic

cmatsuoka:
Lifecycle planning is computed by the sequencer based on part specs and
the current state, resulting in a list of actions to be executed.

Lifecycle is implemented according to the following diagram:

Lifecycle logic

-----


tests/integration/lifecycle/test_simple.py
```diff
@@ -40,6 +40,25 @@
 def test_actions_simple(new_dir, mocker):
```
cjp256, line 40: Perhaps break this test up?  Less simple now :D

tests/integration/lifecycle/test_simple.py
```diff
@@ -40,6 +40,25 @@
...
+    ]
+
+    # foobar part depends on nothing
+    # command: prime foobar
+    lf = craft_parts.LifecycleManager(parts, application_name="test_demo")
+    actions = lf.plan(Step.PRIME, ["foobar"])
+    assert actions == [
+        # FIXME: this will be skipped after the executor writes state to disk
+        # Action("foobar", Step.PULL, action_type=ActionType.SKIP, reason="already ran"),
+        Action("foobar", Step.PULL, action_type=ActionType.RUN, reason=None),
```
cjp256, line 60: Do you specify action_type and reason just to prepare for the state being written?  Or is there something else special about this?

tests/integration/sequencer/test_sequencer.py
```diff
@@ -0,0 +1,280 @@
...
+        p1 = Part("foo", {"plugin": "nil", "after": ["bar"]})
+        p2 = Part("bar", {"plugin": "nil"})
+
+        seq = sequencer.Sequencer(
+            part_list=[p1, p2],
+            project_info=ProjectInfo(),
+        )
+
+        # pylint: disable=line-too-long
+        # fmt: off
```
cjp256, line 128: maybe worth expanding the line length?

cjp256 (approved):

tests/integration/lifecycle/test_simple.py
cmatsuoka: The only special thing is that this line will disappear and will be replaced by the line immediately above that's currently commented out. I just copied the line and replaced the values, it makes the line stand out and is an extra reminder that it should be removed later :)

tests/integration/lifecycle/test_simple.py
cmatsuoka: And it will get much more complex soon.

tests/integration/sequencer/test_sequencer.py
cmatsuoka: Hmm, apparently there's no inline configuration to change pylint's `max-line-length`.

sergiusens (approved):

tests/integration/lifecycle/test_simple.py
cjp256: figured, thanks!

PR #39, opened 2021-04-23
plugins: add plugin registry and management helpers

cmatsuoka:
Application-defined plugins must be registered to be recognized by
Craft Parts. This change implements the plugin registry and helpers
to register, unregister, and retrieve plugin classes and instances.

-----

### DIFF
--- a/craft_parts/plugins/plugins.py
+++ b/craft_parts/plugins/plugins.py
@@ -1,0 +1,107 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2021 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Definitions and helpers to handle plugins."""
+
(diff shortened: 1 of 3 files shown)


craft_parts/plugins/plugins.py
```diff
@@ -0,0 +1,107 @@
...
+
+# Plugin registry by plugin API version
+_BUILTIN_PLUGINS: Dict[str, PluginType] = {
+    "autotools": AutotoolsPlugin,
+    "dump": DumpPlugin,
+    "make": MakePlugin,
+    "nil": NilPlugin,
+}
+
+_PLUGINS = copy.deepcopy(_BUILTIN_PLUGINS)
```
cjp256, line 45: Perhaps consider (if you haven't already) an interface for the application to init craft-parts and receive a context to work with?  The plugin configuration would live in that context rather than a global var.   I imagine that would help for tests, etc.

craft_parts/plugins/plugins.py
```diff
@@ -0,0 +1,107 @@
...
+    """Unregister all user-registered plugins."""
+    global _PLUGINS  # pylint: disable=global-statement
+    _PLUGINS = copy.deepcopy(_BUILTIN_PLUGINS)
+
+
+def strip_plugin_properties(data: Dict[str, Any], *, plugin_name: str) -> None:
+    """Remove plugin-specific entries from part properties.
+
+    :param data: A dictionary containing all part properties.
+    :plugin_name: The name of the plugin.
```
cjp256, line 102:
```suggestion
    :param plugin_name: The name of the plugin.
```

cjp256 (approved):

craft_parts/plugins/plugins.py
cmatsuoka: That's an interesting idea, I'll think about it. Currently only action execution runs in a context.

craft_parts/plugins/plugins.py
cmatsuoka: Oops.

PR #42, opened 2021-04-26
states: ensure minimum interval between state writes

cmatsuoka:
Hosts with low timer resolution may create files with identical
timestamps if files are written at a rate higher than its tick
rate. Prevent that by enforcing a minimum time interval between
subsequent state writes.

-----

### DIFF
--- a/tests/unit/utils/test_os_utils.py
+++ b/tests/unit/utils/test_os_utils.py
@@ -15,10 +15,34 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import os
+from pathlib import Path
 
 import pytest
 
 from craft_parts.utils import os_utils
+
+
+class TestTimedWriter:
+    """Check if minimum interval ensured between writes."""
+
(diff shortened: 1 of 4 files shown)


cjp256 (approved):

tests/unit/utils/test_os_utils.py
```diff
@@ -15,12 +15,36 @@
...
 
 import pytest
 
 from craft_parts.utils import os_utils
 
 
+class TestTimedWriter:
+    """Check if minimum interval ensured between writes."""
+
+    @pytest.mark.usefixtures("new_dir")
```
sergiusens, line 28: these should ideally be 3 distinct tests, unless for organizational reasons, these don't need to be part of a class either.

sergiusens (approved):
With note on test strategy

tests/unit/utils/test_os_utils.py
cmatsuoka: Split into three tests, kept class for organization.

PR #43, opened 2021-04-26
packages: add support to apt cache operations

cmatsuoka:
Add infrastructure to retrieve packages using apt.

Packages/repo is planned to be moved into a separate project, where
it will be further refactored. Current code follows the snapcraft
implementation.

-----

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 PyYAML==5.4.1
+http://archive.ubuntu.com/ubuntu/pool/main/p/python-apt/python-apt_2.1.3ubuntu1.tar.xz; sys_platform == 'linux'
 progressbar==2.5
 pydantic==1.8.1
 pydantic-yaml==0.2.3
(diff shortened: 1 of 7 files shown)


cjp256 (approved):

requirements.txt
```diff
@@ -1,4 +1,5 @@
 PyYAML==5.4.1
+http://archive.ubuntu.com/ubuntu/pool/main/p/python-apt/python-apt_2.1.3ubuntu1.tar.xz; sys_platform == 'linux'
```
cjp256, line 2: As this really needs to come from the host, I'd just leave it at `python-apt` for now.

cmatsuoka: It seems that only a very old version (0.7.8) is available at PyPI, unsure if 2.1.3 available anywhere else.

cjp256: Ah yes, i mean that it needs to be `apt install`d.  It must match the host apt version.

sergiusens: this is not as easy from a python "library" point of view

sergiusens: I'd say we should look into hosting our own wheel

sergiusens (approved):

PR #46, opened 2021-04-27
tests: add fake snapd fixture

cmatsuoka:
The fake snapd fixture will be used to test snap package handling
in `craft-parts.packages`.

Packages/repo is planned to be moved into a separate project, where
it will be further refactored. Current code follows the snapcraft
implementation.

-----

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -15,9 +15,13 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import os
+import tempfile
+from unittest import mock
 
 import pytest
 import xdg  # type: ignore
+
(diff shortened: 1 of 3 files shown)


sergiusens (approved):
You will need to be careful where you use this as this will not work on Windows. One option would be to explore if the fixture setup can skip on itself if it cannot be used.

cmatsuoka:
> You will need to be careful where you use this as this will not work on Windows. One option would be to explore if the fixture setup can skip on itself if it cannot be used.

Will add a note, and check for windows compatibility issues in a separate round. I suspect there are more places in the codebase that need adjustments.

cjp256 (approved):

PR #50, opened 2021-04-29
sources: add source handler integration

cmatsuoka:
Provide calls to integrate source handlers with state manager and the
executor, and use it to check if sources are outdated during the
planning phase.

-----

### DIFF
--- a/craft_parts/sources/errors.py
+++ b/craft_parts/sources/errors.py
@@ -16,11 +16,55 @@
 
 """Source handler error definitions."""
 
+from typing import List
+
 from craft_parts import errors
+from craft_parts.utils import formatting_utils
 
 
 class SourceError(errors.PartsError):
     """Base class for source handler errors."""
+
+
+class InvalidSourceType(SourceError):
+    """Failed to determine a source type."""
+
+    def __init__(self, source: str):
+        self.source = source
+        brief = f"Failed to pull source: unable to determine source type of {source!r}."
+
+        super().__init__(brief=brief)
+
+
+class InvalidSourceOption(SourceError):
+    """A source option is not allowed for the given source type."""
+
+    def __init__(self, *, source_type: str, option: str):
+        self.source_type = source_type
+        self.option = option
+        brief = (
+            f"Failed to pull source: {option!r} cannot be used "
+            f"with a {source_type} source."
+        )
+        resolution = "Make sure sources are correctly specified."
+
+        super().__init__(brief=brief, resolution=resolution)
+
+
+class IncompatibleSourceOptions(SourceError):
+    """Source specified options that can't be used at the same time."""
+
+    def __init__(self, source_type: str, options: List[str]):
+        self.source_type = source_type
+        self.options = options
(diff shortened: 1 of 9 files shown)


craft_parts/sources/errors.py
```diff
@@ -16,13 +16,57 @@
...
+            f"Failed to pull source: {option!r} cannot be used "
+            f"with a {source_type} source."
+        )
+        resolution = "Make sure sources are correctly specified."
+
+        super().__init__(brief=brief, resolution=resolution)
+
+
+class IncompatibleSourceOptions(SourceError):
+    """Source specified options that can't be used at the same time."""
```
sergiusens, line 55:
```suggestion
    """Source specified options that cannot be used at the same time."""
```

craft_parts/sources/sources.py
```diff
@@ -71,16 +71,113 @@
...
+        dirs = ProjectDirs()
+
+    source_handler = None
+    if part.spec.source:
+        handler_class = _get_source_handler_class(
+            part.spec.source,
+            source_type=part.spec.source_type,
+        )
+        source_handler = handler_class(
+            application_name=application_name,
```
sergiusens, line 118: refresh my memory on why the application name is important here

craft_parts/sources/sources.py
```diff
@@ -71,16 +71,113 @@
...
+            source=part.spec.source,
+            part_src_dir=part.part_src_dir,
+            source_checksum=part.spec.source_checksum,
+            source_branch=part.spec.source_branch,
+            source_tag=part.spec.source_tag,
+            source_depth=part.spec.source_depth,
+            source_commit=part.spec.source_commit,
+            dirs=dirs,
```
sergiusens, line 126: looking at this, it seems that we need to make this var a bit more clear on why it's there with a better name

craft_parts/sources/sources.py
```diff
@@ -71,16 +71,113 @@
...
+        )
+
+    return source_handler
+
+
+def _get_source_handler_class(source, *, source_type: str = "") -> SourceHandlerType:
```
sergiusens, line 132: on load is source_type None or ""?

cmatsuoka: It's an empty string in the unmarshaled part if not specified.

craft_parts/sources/sources.py
cmatsuoka: It's used to compose the path to the downloaded file cache.

craft_parts/sources/sources.py
cmatsuoka: Renamed it to `project_dirs` to emphasize the relation to the `ProjectDirs` class.

cjp256 (approved):

craft_parts/sources/sources.py
sergiusens: Ah, maybe this can be global, or in other words, scoped to craft-parts and not specific applications.

sergiusens (approved):
Looks good, we might want to reassess the use of application for sources.

PR #53, opened 2021-05-01
executor: create environment for step execution

cmatsuoka:
Generate a list of environment variables to be set at step execution,
using built-in, plugin, and user-defined variables from the part
specification.

Environment generation logic ported from snapcraft.

-----

### DIFF
--- a/craft_parts/infos.py
+++ b/craft_parts/infos.py
@@ -24,6 +24,7 @@ from typing import Any, Dict, List, Optional
 from craft_parts import errors, utils
 from craft_parts.dirs import ProjectDirs
 from craft_parts.parts import Part
+from craft_parts.steps import Step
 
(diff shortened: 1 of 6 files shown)


craft_parts/infos.py
```diff
@@ -197,6 +198,24 @@ def part_state_dir(self) -> Path:
...
+
+    def __init__(
+        self,
+        part_info: PartInfo,
+        step: Step,
+    ):
+        self._part_info = part_info
+        self.step = step
+
+    def __getattr__(self, name):
```
cjp256, line 212: :cry: would prefer if the caller has to just reference `.part_info`

cjp256 (approved):

sergiusens (approved):

PR #59, opened 2021-05-04
executor: add the step handler

cmatsuoka:
The step handler takes care of the execution of a step, using either
a built-in set of actions to be taken, or executing a user-defined
script defined in the part specification. User-defined scripts may also
call the built-in handler for a step by invoking a control utility.

This class implements the built-in handlers and a FIFO-based mechanism
and API to be used by the external control utility to communicate with
the running instance.

The control API server is a direct port of the implementation used in
Snapcraft, and should be refactored later.

-----

### DIFF
--- a/craft_parts/executor/step_handler.py
+++ b/craft_parts/executor/step_handler.py
@@ -342,6 +342,7 @@ def _fix_pkg_config(
     :param pkg_config_file: The pkg-config file to process.
     :param prefix_trim: The initial path to remove from the configuration prefix.
     """
+    # FIXME: see https://bugs.launchpad.net/snapcraft/+bug/1916281
     pattern_trim = None
     if prefix_trim:
         pattern_trim = re.compile("^prefix={}(?P<prefix>.*)".format(prefix_trim))
--- a/craft_parts/executor/step_handler.py
+++ b/craft_parts/executor/step_handler.py
@@ -1,0 +1,361 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2016-2021 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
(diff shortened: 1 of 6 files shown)


craft_parts/executor/step_handler.py
```diff
@@ -0,0 +1,361 @@
...
+from craft_parts.parts import Part
+from craft_parts.plugins import Plugin
+from craft_parts.sources import SourceHandler
+from craft_parts.steps import Step
+from craft_parts.utils import file_utils
+
+from . import environment, filesets
+from .filesets import Fileset
+
+FilesAndDirs = namedtuple("FilesAndDirs", ["files", "dirs"])
```
cjp256, line 44:
```suggestion
@dataclasses.dataclass
class FilesAndDirs:
  files: Set[str]
  dirs: Set[str]
```

cjp256 (approved):

craft_parts/executor/step_handler.py
cmatsuoka: In this case we're simply organizing the tuple returned by handler functions, and adding the overhead of a full dataclass is probably not worth it.

cjp256: > In this case we're simply organizing the tuple returned by handler functions, and adding the overhead of a full dataclass is probably not worth it.

The advantage to the dataclass over namedtuple is type information.  Which could be helpful in development.  But not really critical.

craft_parts/executor/step_handler.py
```diff
@@ -0,0 +1,361 @@
...
+) -> None:
+    """Rewrite the prefix entry in a pkg-config file.
+
+    :param root: The root to add to the configuration prefix.
+    :param pkg_config_file: The pkg-config file to process.
+    :param prefix_trim: The initial path to remove from the configuration prefix.
+    """
+    pattern_trim = None
+    if prefix_trim:
+        pattern_trim = re.compile("^prefix={}(?P<prefix>.*)".format(prefix_trim))
```
sergiusens, line 347: maybe take the opportunity to fix https://bugs.launchpad.net/snapcraft/+bug/1916281 :-)

sergiusens (approved):
Looks good, not a fan of returning sets for pull and build unless we have a reason I am not aware of in the future (or maybe it is just for symmetry)

cmatsuoka:
> Looks good, not a fan of returning sets for pull and build unless we have a reason I am not aware of in the future (or maybe it is just for symmetry)

It was done this way to allow uniform call protocol for all steps, but it can be refactored later if needed.

craft_parts/executor/step_handler.py
cmatsuoka: Will add a note to the code and fix in a different PR.

PR #60, opened 2021-05-04
executor: add callbacks

cmatsuoka:
Callbacks are a mechanism to run application-specified functions at
certain points of the lifecycle execution, such as before or after
each step is processed, before the execution phase starts, or after
the execution phase ends.

-----

### DIFF
--- a/craft_parts/callbacks.py
+++ b/craft_parts/callbacks.py
@@ -1,0 +1,140 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2021 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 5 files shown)


craft_parts/callbacks.py
```diff
@@ -0,0 +1,140 @@
...
+
+_PROLOGUE_HOOKS: List[CallbackHook] = []
+_EPILOGUE_HOOKS: List[CallbackHook] = []
+_PRE_STEP_HOOKS: List[CallbackHook] = []
+_POST_STEP_HOOKS: List[CallbackHook] = []
+
+logger = logging.getLogger(__name__)
+
+
+def register_prologue(func: ExecutionCallback) -> None:
```
cjp256, line 42: All of these hooks should likely be combined into a context object and available as properties rather than being global.  Then the idea of "clear()" is no longer required.

cjp256 (approved):

craft_parts/callbacks.py
cmatsuoka: My initial idea was to have a callback manager instantiated by the lifecycle manager, and this callback manager would hold the lists. A similar approach could be used for plugins. However, the typical use case for both callbacks and plugins, if they're used at all, is to register them once in the caller application and never unregister -- so static registration seemed to simplify the API because callbacks and plugins could then be registered at any point of the code (e.g. at start of execution), without having to wait for the lifecycle manager to be created. I'll think about it again.

sergiusens (approved):

PR #66, opened 2021-05-10
executor: add part handler (1/2)

cmatsuoka:
Handle part run actions by fetching and unpacking packages, calling the
appropriate step handler, and generating the step state data.

-----

### DIFF
--- a/craft_parts/executor/part_handler.py
+++ b/craft_parts/executor/part_handler.py
@@ -42,7 +42,7 @@ logger = logging.getLogger(__name__)
 
 
 class PartHandler:
-    """Handle steps for a part using the appropriate plugins.
+    """Handle lifecycle steps for a part.
 
     :param part: The part being processed.
     :param part_info: Information about the part being processed.
--- a/craft_parts/executor/part_handler.py
+++ b/craft_parts/executor/part_handler.py
@@ -1,0 +1,387 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2017-2021 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Definitions and helpers for part handlers."""
+
+import logging
+import os
+import os.path
+import shutil
+from glob import iglob
+from pathlib import Path
+from typing import Any, Callable, Dict, List, Optional
+
+from craft_parts import callbacks, errors, packages, plugins, sources
+from craft_parts.actions import Action, ActionType
+from craft_parts.infos import PartInfo, StepInfo
+from craft_parts.packages import errors as packages_errors
+from craft_parts.parts import Part
+from craft_parts.plugins import Plugin
+from craft_parts.state_manager import states
+from craft_parts.state_manager.states import StepState
+from craft_parts.steps import Step
+from craft_parts.utils import os_utils
+
+from .organize import organize_files
+from .step_handler import StepContents, StepHandler
+
+logger = logging.getLogger(__name__)
+
+
+class PartHandler:
+    """Handle steps for a part using the appropriate plugins.
+
+    :param part: The part being processed.
+    :param part_info: Information about the part being processed.
+    :param part_list: A list containing all parts.
+    """
+
+    def __init__(
+        self,
+        part: Part,
+        *,
+        part_info: PartInfo,
+        part_list: List[Part],
+    ):
+        self._part = part
+        self._part_info = part_info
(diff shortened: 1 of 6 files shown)


craft_parts/executor/part_handler.py
```diff
@@ -0,0 +1,387 @@
...
+        """Run the scriptlet if overriding, otherwise run the built-in handler.
+
+        :param step_info: Information about the step to execute.
+        :param scriptlet_name: The name of this step's scriptlet.
+        :param work_dir: The path to run the scriptlet on.
+
+        :return: If step is Stage or Prime, return a tuple of sets containing
+            the step's file and directory artifacts.
+        """
+        if not step_info.step:
```
cjp256, line 233: the type definition of `StepInfo` does not allow for None step, seems an unnecessary check unless it is set to None somewhere (and shouldn't be), though harmless

cmatsuoka: This is just a consistency check to make sure nobody called us with a (handcrafted?) invalid parameter, and it also makes pyright happy.

craft_parts/executor/part_handler.py
```diff
@@ -0,0 +1,387 @@
...
+            handler = self._run_build
+        elif action.step == Step.STAGE:
+            handler = self._run_stage
+        elif action.step == Step.PRIME:
+            handler = self._run_prime
+        else:
+            raise RuntimeError("cannot run action for invalid step {action.step!r}")
+
+        callbacks.run_pre_step(step_info)
+        state = handler(step_info)
```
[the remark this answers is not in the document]
cmatsuoka: I assigned it to the `state` variable to be explicit about what `handler` is returning.

craft_parts/executor/part_handler.py
cjp256: Interesting, that's what I was wondering.  Why does pyright think it can be None?

craft_parts/infos.py
```diff
@@ -100,6 +102,11 @@ def target_arch(self) -> str:
         """Return the architecture used for debs, snaps and charms."""
         return self._machine["deb"]
 
+    @property
+    def base(self) -> str:
```
cjp256, line 106: Is this just required for repo?

cjp256 (approved):

craft_parts/infos.py
cmatsuoka: Yes, it's only used in `fetch_stage_packages` to filter package names for that base.

craft_parts/executor/part_handler.py
cmatsuoka: I just used the current set of linting tools after removing the check and nobody complained, so it seems that we can remove it now.

craft_parts/executor/part_handler.py
```diff
@@ -0,0 +1,384 @@
...
+from craft_parts.utils import os_utils
+
+from .organize import organize_files
+from .step_handler import StepContents, StepHandler
+
+logger = logging.getLogger(__name__)
+
+
+class PartHandler:
+    """Handle steps for a part using the appropriate plugins.
```
sergiusens, line 45: Is mentioning "using the appropriate plugins" necessary considering that you pass a Part (which encapsulates the plugin)?

```suggestion
    """Manage the life cycle steps for a Part.
```

I might be missing doc hints for sphinx here

craft_parts/executor/part_handler.py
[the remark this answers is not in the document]
sergiusens: to be really explicit this may need state to have the type explicitly set instead of auto set

craft_parts/executor/part_handler.py
```diff
@@ -0,0 +1,384 @@
...
+            raise RuntimeError("cannot run action for invalid step {action.step!r}")
+
+        callbacks.run_pre_step(step_info)
+        state = handler(step_info)
+        state_file = states.state_file_path(self._part, action.step)
+        state.write(state_file)
+        callbacks.run_post_step(step_info)
+
+    def _run_pull(self, step_info: StepInfo) -> StepState:
+        _remove(self._part.part_src_dir)
```
sergiusens, line 114: are we ever going to consider a reentrant "pull" that updates sources? (life git pull multiple times?)

craft_parts/executor/part_handler.py
```diff
@@ -0,0 +1,384 @@
...
+        )
+
+        self._part_properties = part.spec.marshal()
+        self._source_handler = sources.get_source_handler(
+            application_name=part_info.application_name,
+            part=part,
+            project_dirs=part_info.dirs,
+        )
+
+        self._build_packages = _get_build_packages(part=self._part, plugin=self._plugin)
```
sergiusens, line 76: these could be lazy loaded
https://docs.python.org/3.8/library/functools.html#functools.cached_property

Is there a reason this is done during __init__? If so, maybe code comment it so no one tries to optimize. 
The other option deviating from lazy loading with the same objective is to just call `_get_build_packages(part=self._part, plugin=self._plugin)` in the assets dict decl.

sergiusens (approved):
Looks good, I do have questions which may be missing some context

craft_parts/executor/part_handler.py
cmatsuoka: This could save time and bandwidth if e.g. git sources are being changed locally, but would also require manual cleaning when starting over. Currently we support updating for certain sources (local. git) which should cover the "locally changed sources" scenario. Update support is scheduled to be added in the forthcoming Part Handler (2/2) PR.

PR #68, opened 2021-05-13
executor: add part handler (2/2)

cmatsuoka:
Handle `ActionType.RERUN` and `ActionType.UPDATE` actions.

-----

### DIFF
--- a/craft_parts/executor/part_handler.py
+++ b/craft_parts/executor/part_handler.py
@@ -275,12 +275,14 @@ class PartHandler:
             )
             return
 
-        # consistency check
+        # the sequencer won't generate update actions for parts without
+        # source, but they can be created manually
         if not self._source_handler:
-            raise RuntimeError(
-                f"Update requested on part {self._part.name!r} without "
-                f"a source handler."
+            logger.warning(
+                "Update requested on part %r without a source handler.",
+                self._part.name,
             )
+            return
 
         # the update action is sequenced only if an update is required and the
         # source knows how to update
--- a/craft_parts/executor/part_handler.py
+++ b/craft_parts/executor/part_handler.py
@@ -517,7 +517,7 @@ def _clean_migrated_files(files: Set[str], dirs: Set[str], directory: Path) -> N
     """
     for each_file in files:
         try:
-            os.remove(os.path.join(directory, each_file))
+            Path(directory, each_file).unlink()
         except FileNotFoundError:
             logger.warning(
                 "Attempted to remove file %r, but it didn't exist. Skipping...",
--- a/craft_parts/executor/part_handler.py
+++ b/craft_parts/executor/part_handler.py
@@ -22,7 +22,7 @@ import os.path
 import shutil
 from glob import iglob
 from pathlib import Path
-from typing import Any, Callable, Dict, List, Optional
+from typing import Any, Callable, Dict, List, Optional, Set
 
 from craft_parts import callbacks, errors, packages, plugins, sources
 from craft_parts.actions import Action, ActionType
@@ -33,7 +33,7 @@ from craft_parts.plugins import Plugin
 from craft_parts.state_manager import states
 from craft_parts.state_manager.states import StepState
 from craft_parts.steps import Step
-from craft_parts.utils import os_utils
+from craft_parts.utils import file_utils, os_utils
 
 from .organize import organize_files
 from .step_handler import StepContents, StepHandler
@@ -84,12 +84,11 @@ class PartHandler:
         step_info = StepInfo(self._part_info, action.step)
 
         if action.action_type == ActionType.UPDATE:
-            # TODO: add update action handler
+            self._update_action(action, step_info=step_info)
             return
 
         if action.action_type == ActionType.RERUN:
-            # TODO: clean step
-            pass
+            self.clean_step(step=action.step)
 
         handler: Callable[[StepInfo], StepState]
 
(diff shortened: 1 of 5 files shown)


craft_parts/executor/part_handler.py
```diff
@@ -245,6 +244,131 @@ def _run_step(
...
+        if self._part.spec.override_pull:
+            self._run_step(
+                step_info=step_info,
+                scriptlet_name="override-pull",
+                work_dir=self._part.part_src_dir,
+            )
+            return
+
+        # consistency check
+        if not self._source_handler:
```
cmatsuoka, line 279: The sequencer won't generate inconsistent actions, but test to prevent problems caused by actions added by hand. This also keeps linters happy.

cjp256: That's because the spec for `get_source_handler` returns Optional source handler.  (that is, `source` was not specified`)

cjp256: Is that a problem for this?  Or does _update_pull() never get invoked if no source?

craft_parts/executor/part_handler.py
```diff
@@ -245,6 +244,131 @@ def _run_step(
...
+    def _update_build(self, step_info: StepInfo) -> None:
+        self._make_dirs()
+        self._unpack_stage_packages()
+        self._unpack_stage_snaps()
+
+        if not self._plugin.out_of_source_build:
+            # Use the local source to update. It's important to use
+            # file_utils.copy instead of link_or_copy, as the build process
+            # may modify these files
+            source = sources.LocalSource(
```
cjp256, line 300: Would it be reasonable to simply call the copy rather than attempt to use the local plugin?  Or is the `check_if_outdated()` call providing value?

craft_parts/executor/part_handler.py
cmatsuoka: Update events are not generated by craft-parts for parts without `source` (for example when using the `nil` plugin), but it's possible to create `Action("part_name", Step.PULL, ActionType.UPDATE)` for that part and send it to the action executor. In this case it will raise an exception (it's `RuntimeError` in the code, but I think it should be `InvalidAction`, will update accordingly).

cmatsuoka: Or perhaps we should just log the attempt and do nothing.

craft_parts/executor/part_handler.py
cmatsuoka: Yes, `check_if_outdated()` must be called before `update`. I'll add a note to the source code.

craft_parts/executor/part_handler.py
cjp256: What if there was a source for a part, then removed?

cmatsuoka: The part specification is re-evaluated at each run, so if source was removed it will be considered non-existent.

cjp256 (approved):

craft_parts/executor/part_handler.py
```diff
@@ -332,6 +458,71 @@ def _remove(filename: Path) -> None:
...
+
+    # Directories may not be ordered so that subdirectories come before
+    # parents, and we want to be able to remove directories if possible, so
+    # we'll sort them in reverse here to get subdirectories before parents.
+
+    for each_dir in sorted(dirs, reverse=True):
+        migrated_directory = os.path.join(directory, each_dir)
+        try:
+            if not os.listdir(migrated_directory):
+                os.rmdir(migrated_directory)
```
facundobatista, line 517: Why are we avoiding to remove directories? This allows the possibility of asking for "clean" and not getting it cleaned...

craft_parts/executor/part_handler.py
```diff
@@ -332,6 +458,71 @@ def _remove(filename: Path) -> None:
...
+def _clean_migrated_files(files: Set[str], dirs: Set[str], directory: Path) -> None:
+    """Remove files and directories migrated from part install to a common directory.
+
+    :param files: A set of files to remove.
+    :param dirs: A set of directories to remove.
+    :param directory: The path to remove files and directories from.
+    """
+    for each_file in files:
+        try:
+            os.remove(os.path.join(directory, each_file))
```
facundobatista, line 502: Other parts of this code work with "pathlib", shouldn't we standarize everything to it?

facundobatista (changes requested):
Good, thanks! Commented some details.

craft_parts/executor/part_handler.py
```diff
@@ -332,6 +458,71 @@ def _remove(filename: Path) -> None:
...
+    directories = state.directories
+
+    # We want to make sure we don't remove a file or directory that's
+    # being used by another part. So we'll examine the state for all parts
+    # in the project and leave any files or directories found to be in
+    # common.
+    for other_name, other_state in part_states.items():
+        if other_state and (other_name != part_name):
+            files -= other_state.files
+            directories -= other_state.directories
```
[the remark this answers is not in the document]
cmatsuoka: This step state data is loaded from disk only for this operation and is discarded afterwards, so changing them directly shouldn't be a problem.

craft_parts/executor/part_handler.py
cmatsuoka: We don't remove the directory if it contains files placed by a different part. Only the stage and prime steps clean migrated files and the directory is shared by all parts, so we should remove only the files and subdirs that belong to the part we're currently cleaning.

craft_parts/executor/part_handler.py
cmatsuoka: Yes. Fixing it, and there's a task already in place to standardize it across the whole project.

facundobatista (approved):
Wonderful!

PR #69, opened 2021-05-17
executor: add executor class and integrate with lifecycle manager (CRAFT-31)

cmatsuoka:
Add the calls to execute actions using the lifecycle manager.

Additional tests will be added in separate PRs.

-----

### DIFF
--- a/craft_parts/executor/executor.py
+++ b/craft_parts/executor/executor.py
@@ -177,16 +177,14 @@ class ExecutionContext:
         *,
         executor: Executor,
     ):
-        self._prologue = executor.prologue
-        self._epilogue = executor.epilogue
-        self._execute = executor.execute
+        self._executor = executor
 
     def __enter__(self) -> "ExecutionContext":
-        self._prologue()
+        self._executor.prologue()
         return self
 
     def __exit__(self, *exc):
-        self._epilogue()
+        self._executor.epilogue()
 
     def execute(self, actions: Union[Action, List[Action]]) -> None:
         """Execute the specified action or list of actions.
@@ -196,4 +194,4 @@ class ExecutionContext:
 
         :raises InvalidActionException: If the action parameters are invalid.
         """
-        self._execute(actions)
+        self._executor.execute(actions)
--- a/craft_parts/executor/executor.py
+++ b/craft_parts/executor/executor.py
@@ -1,0 +1,199 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2021 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Definitions and helpers for the action executor."""
+
+import logging
+from typing import Dict, List, Union
+
+from craft_parts import callbacks, packages, parts
+from craft_parts.actions import Action, ActionType
+from craft_parts.infos import PartInfo, ProjectInfo
+from craft_parts.parts import Part
+from craft_parts.steps import Step
+from craft_parts.utils import os_utils
+
+from .collisions import check_for_stage_collisions
+from .part_handler import PartHandler
+
+logger = logging.getLogger(__name__)
+
+
+class Executor:
+    """Execute lifecycle actions."""
+
+    def __init__(
+        self,
+        *,
+        part_list: List[Part],
+        project_info: ProjectInfo,
+        extra_build_packages: List[str] = None,
+        extra_build_snaps: List[str] = None,
+    ):
+        self._part_list = part_list
+        self._project_info = project_info
+        self._extra_build_packages = extra_build_packages
+        self._extra_build_snaps = extra_build_snaps
+        self._handler: Dict[str, PartHandler] = {}
+
+    def prologue(self):
+        """Prepare the execution environment.
+
+        This method is called before executing lifecycle actions.
+        """
+        self._install_build_packages()
+        self._install_build_snaps()
+
+        callbacks.run_prologue(self._project_info, part_list=self._part_list)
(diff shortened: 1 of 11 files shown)


craft_parts/executor/executor.py
```diff
@@ -0,0 +1,199 @@
...
+        selected_steps.reverse()
+
+        for part in selected_parts:
+            self._create_part_handler(part)
+            handler = self._handler[part.name]
+
+            for step in selected_steps:
+                handler.clean_step(step=step)
+
+    def _run_action(self, action: Action):
```
cjp256, line 103:
```suggestion
    def _run_action(self, action: Action) -> None:
```

craft_parts/executor/executor.py
```diff
@@ -0,0 +1,199 @@
...
+
+        if action.step == Step.STAGE:
+            check_for_stage_collisions(self._part_list)
+
+        self._create_part_handler(part)
+
+        handler = self._handler[part.name]
+        handler.run_action(action)
+
+    def _create_part_handler(self, part: Part):
```
cjp256, line 124:
```suggestion
    def _create_part_handler(self, part: Part) -> None:
```

But may be nicer if also returning the part.

craft_parts/executor/executor.py
```diff
@@ -0,0 +1,199 @@
...
+        """Instantiate a part handler for a new part."""
+        if part.name not in self._handler:
+            self._handler[part.name] = PartHandler(
+                part,
+                part_info=PartInfo(self._project_info, part),
+                part_list=self._part_list,
+            )
+
+    def _install_build_packages(self):
```
cjp256, line 133:
```suggestion
    def _install_build_packages(self) -> None:
```

craft_parts/executor/executor.py
```diff
@@ -0,0 +1,199 @@
...
+        """Prepare the execution environment.
+
+        This method is called before executing lifecycle actions.
+        """
+        self._install_build_packages()
+        self._install_build_snaps()
+
+        callbacks.run_prologue(self._project_info, part_list=self._part_list)
+
+    def epilogue(self):
```
cjp256, line 62:
```suggestion
    def epilogue(self) -> None:
```

craft_parts/executor/executor.py
```diff
@@ -0,0 +1,199 @@
...
+        extra_build_packages: List[str] = None,
+        extra_build_snaps: List[str] = None,
+    ):
+        self._part_list = part_list
+        self._project_info = project_info
+        self._extra_build_packages = extra_build_packages
+        self._extra_build_snaps = extra_build_snaps
+        self._handler: Dict[str, PartHandler] = {}
+
+    def prologue(self):
```
cjp256, line 52:
```suggestion
    def prologue(self) -> None:
```

craft_parts/executor/executor.py
```diff
@@ -0,0 +1,199 @@
...
+
+class ExecutionContext:
+    """A context manager to handle lifecycle action executions."""
+
+    def __init__(
+        self,
+        *,
+        executor: Executor,
+    ):
+        self._prologue = executor.prologue
```
cjp256, line 180: Perhaps just save the reference to executor instead of duplicating?

cjp256 (approved):

cmatsuoka:
Amended the commit message to mention to the task number as requested by @sergiusens

craft_parts/executor/part_handler.py
```diff
@@ -76,6 +76,25 @@ def __init__(
...
 
+    @property
+    def build_packages(self) -> List[str]:
+        """Return the list of build packages defined for this part.
+
+        The list of build packages include packages defined directly in
+        the parts specification, packages required by the source handler,
+        and packages required by the plugin.
+        """
+        return self._build_packages
```
facundobatista, line 87: What's the point of these properties? Can't we just make `self._build_packages` public?

Note that you're not protecting this attribute of being modified, as it's mutable, so there's no point in protecting it from somebody doing `part_handler.build_packages = []` when they still could do `part_handler.build_packages.clear()`.

craft_parts/executor/executor.py
```diff
@@ -0,0 +1,197 @@
...
+            specified, all parts will be cleaned.
+        """
+        selected_parts = parts.part_list_by_name(part_names, self._part_list)
+
+        selected_steps = [initial_step] + initial_step.next_steps()
+        selected_steps.reverse()
+
+        for part in selected_parts:
+            self._create_part_handler(part)
+            handler = self._handler[part.name]
```
facundobatista, line 98: it would be simpler if `_create_part_handler` would return the handler, it makes more sense to write here:

```
handler = self._create_part_handler(part)
```

craft_parts/executor/executor.py
```diff
@@ -0,0 +1,197 @@
...
+        if action.action_type == ActionType.SKIP:
+            logger.debug("Skip execution of %s (because %s)", action, action.reason)
+            return
+
+        if action.step == Step.STAGE:
+            check_for_stage_collisions(self._part_list)
+
+        self._create_part_handler(part)
+
+        handler = self._handler[part.name]
```
facundobatista, line 121: same here, just use a `handler` returned by creating function

craft_parts/executor/executor.py
```diff
@@ -0,0 +1,197 @@
...
+                part_info=PartInfo(self._project_info, part),
+                part_list=self._part_list,
+            )
+
+    def _install_build_packages(self) -> None:
+        for part in self._part_list:
+            self._create_part_handler(part)
+
+        build_packages = set()
+        for _, handler in self._handler.items():
```
facundobatista, line 138: If you will discard the keys, just use `self._handler.values()`.

craft_parts/executor/executor.py
```diff
@@ -0,0 +1,197 @@
...
+            build_packages.update(handler.build_packages)
+
+        if self._extra_build_packages:
+            build_packages.update(self._extra_build_packages)
+
+        packages.Repository.install_build_packages(sorted(build_packages))
+
+    def _install_build_snaps(self) -> None:
+        build_snaps = set()
+        for _, handler in self._handler.items():
```
facundobatista, line 148: `.values()`

facundobatista (changes requested):
Nice! Added some comments. Thanks!

craft_parts/executor/part_handler.py
cmatsuoka: Good point. Removing them.

craft_parts/executor/executor.py
cmatsuoka: Much better, thanks.

PR #73, opened 2021-05-19
tests: add stage snaps integration test (CRAFT-68)

cmatsuoka:
Verify if stage-snaps are correctly unpacked into the part install dir.

-----


tests/integration/lifecycle/test_stage_snaps.py
```diff
@@ -0,0 +1,95 @@
...
+        ctx.execute(actions[0])
+
+    snaps = list(Path("parts/foo/stage_snaps").glob("*.snap"))
+    assert len(snaps) == 1
+    assert snaps[0].name == "test-snap.snap"
+
+    ctx.execute(actions[1])
+
+    foo_install_dir = Path(new_dir / "parts" / "foo" / "install")
+    assert (foo_install_dir / "meta.basic" / "snap.yaml").is_file()
```
sergiusens, line 64: the file is test-snap and the name in snap.yaml is probably basic. For clarity, I would either:

- clarify
- rename the snap name
- rename the snap file

tests/integration/lifecycle/test_stage_snaps.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def test_stage_snap_error(new_dir, fake_snap_command):
```
sergiusens, line 67: can you clarify the type of error? Is this download error, unpack error? Do we need to have one for each of these?

sergiusens (approved):
Looks good, approved with comments.

facundobatista (approved):
Like it.

tests/integration/lifecycle/test_stage_snaps.py
cmatsuoka: It bothered me too, I'll rename the snap file to be consistent.

tests/integration/lifecycle/test_stage_snaps.py
cmatsuoka: Added separate tests for download and pull error.

PR #74, opened 2021-05-20
lifecycle: integrate executor clean and package list refresh (CRAFT-32)

cmatsuoka:
Add step and part cleaning and package list refresh methods to the
lifecycle manager.

-----

### DIFF
--- a/craft_parts/lifecycle_manager.py
+++ b/craft_parts/lifecycle_manager.py
@@ -16,11 +16,11 @@
 
 """The parts lifecycle manager."""
 
-from typing import Any, Dict, List, Sequence
+from typing import Any, Dict, List, Optional, Sequence
 
 from pydantic import ValidationError
 
-from craft_parts import errors, executor, plugins, sequencer
+from craft_parts import errors, executor, packages, plugins, sequencer
 from craft_parts.actions import Action
 from craft_parts.dirs import ProjectDirs
 from craft_parts.infos import ProjectInfo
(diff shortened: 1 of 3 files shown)


sergiusens (approved):

craft_parts/lifecycle_manager.py
```diff
@@ -103,6 +103,40 @@ def project_info(self) -> ProjectInfo:
...
+
+        Cleaning a step removes its state and all artifacts generated in that
+        step and subsequent steps for the specified parts.
+
+        :para step: The step to clean.
+        :param part_names: The list of part names to clean. If not specified,
+            all parts will be cleaned and work directories will be removed.
+        """
+        if not step:
+            step = Step.PULL
```
facundobatista, line 119: How weird would be to put this default in the signature above?

facundobatista (approved):
Approved, but see a comment, thanks!

craft_parts/lifecycle_manager.py
```diff
@@ -103,6 +103,40 @@ def project_info(self) -> ProjectInfo:
...
+        it shouldn't be updated between planning and execution.
+
+        :param system: Also refresh the list of available build packages to
+            install on the host system.
+        """
+        packages.Repository.refresh_stage_packages_list(
+            application_name=self._application_name, target_arch=self._target_arch
+        )
+
+        if system:
```
cjp256, line 137: Is it expected to be called more than once with a conditional?  Maybe they should just be separate?

craft_parts/lifecycle_manager.py
cmatsuoka: Good idea, will do that.

craft_parts/lifecycle_manager.py
cmatsuoka: We'll also need to update the overlay package list and my idea was to consolidate all three in a single call to simplify the API. Let's see how the overlay discussion goes, we can refactor later if needed.

PR #75, opened 2021-05-25
ctl: add ctl protocol client (CRAFT-35)

cmatsuoka:
The control protocol client allows a user scriptlet to execute the
default handler for a step in the running application context, or set
the value of a custom variable previously passed as an argument to
`LifecycleManager`.
    
The `craft_parts.ctl` module can be imported by application-defined
clients, or invoked as a utility from the command line using the
supplied `partsctl` tool.
    

-----

### DIFF
--- a/craft_parts/ctl.py
+++ b/craft_parts/ctl.py
@@ -68,7 +68,6 @@ def client(cmd: str, args: List[str]):
 
     with open(call_fifo, "w") as fifo:
         fifo.write(json.dumps(data))
-        fifo.flush()
 
     with open(feedback_fifo, "r") as fifo:
         feedback = fifo.readline().strip()
--- a/craft_parts/ctl.py
+++ b/craft_parts/ctl.py
@@ -1,0 +1,70 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2017-2021 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Helpers to invoke step execution handlers from the command line."""
+
+import json
+import os
+import sys
+from typing import List
+
+
+def main():
+    """Run the ctl client cli."""
+    if len(sys.argv) < 2:
+        print(f"usage: {sys.argv[0]} <command>")
+        sys.exit(1)
+
+    cmd, param = sys.argv[1], sys.argv[2:]
+    try:
+        client(cmd, param)
+    except RuntimeError as err:
+        print(f"Error: {err}.", file=sys.stderr)
+        sys.exit(1)
+
+
+def client(cmd: str, args: List[str]):
+    """Execute a command in the running step processor.
+
+    :param cmd: The function to execute in the step processor.
+    :param args: Optional arguments.
+
+    :raise RuntimeError: If the command is invalid.
+    """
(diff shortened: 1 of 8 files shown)


craft_parts/ctl.py
```diff
@@ -0,0 +1,75 @@
...
+    """Run the ctl client cli."""
+    if len(sys.argv) < 2:
+        print(f"usage: {sys.argv[0]} <command>")
+        sys.exit(1)
+
+    cmd, param = sys.argv[1], sys.argv[2:]
+    try:
+        client(cmd, param)
+    except RuntimeError as err:
+        print(f"Error: {err}.", file=sys.stderr)
```
cjp256, line 35: I don't think it is desirable to pollute output with this extra text.
```suggestion
        logger.debug("ctl error: %r", err)
```

cjp256 (approved):
LGTM with exception of the output

craft_parts/ctl.py
cmatsuoka: I'll set the logger level to info so the error message is actually shown if partsctl is executed from the command line. This should be replaced by the appropriate standard messages call later.
```
$ partsctl pull
ctl error: 'PARTS_CALL_FIFO' environment variable must be defined.
Note that this utility is designed for use only in part scriptlets.
```

cjp256: Thanks!

craft_parts/ctl.py
```diff
@@ -0,0 +1,78 @@
...
+import sys
+from typing import List
+
+logger = logging.getLogger(__name__)
+
+
+def main():
+    """Run the ctl client cli."""
+    if len(sys.argv) < 2:
+        print(f"usage: {sys.argv[0]} <command>")
```
facundobatista, line 31: This should expose that there may be parameters for the command.

craft_parts/ctl.py
```diff
@@ -0,0 +1,78 @@
...
+        sys.exit(1)
+
+    cmd, param = sys.argv[1], sys.argv[2:]
+    try:
+        client(cmd, param)
+    except RuntimeError as err:
+        logger.warning("ctl error: %s", err)
```
facundobatista, line 38: why not `logger.error`?

craft_parts/ctl.py
```diff
@@ -0,0 +1,78 @@
...
+        raise RuntimeError(
+            "{!s} environment variable must be defined.\nNote that this "
+            "utility is designed for use only in part scriptlets.".format(err)
+        ) from err
+
+    data = {"function": cmd, "args": args}
+
+    with open(call_fifo, "w") as fifo:
+        fifo.write(json.dumps(data))
+        fifo.flush()
```
facundobatista, line 71: This flush is not needed, it will happen when the file is closed.

craft_parts/ctl.py
```diff
@@ -0,0 +1,78 @@
...
+
+    with open(feedback_fifo, "r") as fifo:
+        feedback = fifo.readline().strip()
```
facundobatista, line 74: May it happen to be more than one line?

craft_parts/executor/step_handler.py
```diff
@@ -267,7 +264,43 @@ def _handle_control_api(self, scriptlet_name, function_call) -> None:
...
+            self._handle_step_function(function_name)
+        elif function_name == "set":
+            if len(function_args) != 2:
+                raise invalid_control_api_call(
+                    message=(
+                        f"invalid number of arguments to function {function_name!r}"
+                    ),
+                )
+            name = function_args[0]
+            value = function_args[1]
```
facundobatista, line 289: you already know that length is 2, so much simpler and readable to do `name, value = function_args`

facundobatista (changes requested):
Nice! I added some comments inline, thanks.

craft_parts/ctl.py
cmatsuoka: Good point. It's an error, after all.

craft_parts/ctl.py
cmatsuoka: It could, but the control protocol used in snapcraft assumes that only the first line has significant content.

facundobatista: Ack.

facundobatista (approved):
Wonderful, thanks!

PR #76, opened 2021-05-26
packages: normalize unpacked stage packages (CRAFT-33)

cmatsuoka:
Packages unpacked in the stage area may contain files that need to be
patched to work correctly. Patches include replacing absolute pathnames
to the python interpreter in script shebangs, converting absolute
symlinks to relative, cleaning suid/guid bits, and changing path names
in pkg-config and xml2/xslt config files to reflect the new installation
root.

The implementation of package normalization and tests is ported from
Snapcraft, and will be replaced by an upcoming stand-alone package
containing repository and package operations.

-----

### DIFF
--- a/craft_parts/packages/deb.py
+++ b/craft_parts/packages/deb.py
@@ -34,6 +34,7 @@ from craft_parts.utils import file_utils, os_utils
 from . import errors
 from .base import BaseRepository, get_pkg_name_parts, mark_origin_stage_package
 from .deb_package import DebPackage
+from .normalize import normalize
 
 if sys.platform == "linux":
     # Ensure importing works on non-Linux.
@@ -500,8 +501,7 @@ class Ubuntu(BaseRepository):
                 mark_origin_stage_package(extract_dir, marked_name)
                 # Stage files to install_dir.
                 file_utils.link_or_copy_tree(extract_dir, install_path.as_posix())
-        # XXX: normalization of stage packages done by the application
-        # cls.normalize(str(install_path))
+        normalize(str(install_path), repository=cls)
 
     @classmethod
     def is_package_installed(cls, package_name) -> bool:
--- a/craft_parts/packages/deb.py
+++ b/craft_parts/packages/deb.py
@@ -505,7 +505,7 @@ class Ubuntu(BaseRepository):
                 file_utils.link_or_copy_tree(extract_dir, install_path.as_posix())
 
         if pkg_path:
-            normalize(str(install_path), repository=cls)
+            normalize(install_path, repository=cls)
 
     @classmethod
     def is_package_installed(cls, package_name) -> bool:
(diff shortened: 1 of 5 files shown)


cjp256 (approved):
LGTM

craft_parts/packages/deb.py
```diff
@@ -489,7 +490,12 @@ def unpack_stage_packages(
         cls, *, stage_packages_path: pathlib.Path, install_path: pathlib.Path
     ) -> None:
         """Unpack stage packages to install_path."""
-        for pkg_path in stage_packages_path.glob("*.deb"):
+        stage_packages = stage_packages_path.glob("*.deb")
+
+        if not stage_packages:
```
facundobatista, line 495: `stage_packages` is a generator, it's `bool` will always be True. You need to consume it.

craft_parts/packages/deb.py
```diff
@@ -500,8 +506,8 @@ def unpack_stage_packages(
                 mark_origin_stage_package(extract_dir, marked_name)
                 # Stage files to install_dir.
                 file_utils.link_or_copy_tree(extract_dir, install_path.as_posix())
-        # XXX: normalization of stage packages done by the application
-        # cls.normalize(str(install_path))
+
+        normalize(str(install_path), repository=cls)
```
facundobatista, line 510: `normalize` should also handle Pathlib. Otherwise we'd have half of the code with string paths, the other half with pathlib.Path, and that's no good.

craft_parts/packages/normalize.py
```diff
@@ -0,0 +1,263 @@
...
+from typing import TYPE_CHECKING, Optional, Pattern, Union
+
+if TYPE_CHECKING:
+    from .base import RepositoryType
+
+
+logger = logging.getLogger(__name__)
+
+
+def normalize(unpack_dir: str, *, repository: "RepositoryType") -> None:
```
facundobatista, line 38: This is what I mentioned above, `unpack_dir` should be pathlib.Path.

facundobatista (changes requested):
Nice! I annotated a  couple of details and a big pathlib/str issue. Thanks!

craft_parts/packages/deb.py
cmatsuoka: Good catch! This line actually comes from snapcore/snapcraft/pull/3529 and it was added considering the behavior of `glob.glob()` but not `Path.glob()`. @sergiusens may want to fix it there as well.

cmatsuoka: Fixed, and also opened a PR for snapcraft.

craft_parts/packages/deb.py
cmatsuoka: Refactored the normalize module to use pathlib.

craft_parts/packages/normalize.py
cmatsuoka: Refactored the normalize module to use pathlib.

facundobatista (approved):
Awesome, t
hanks!

PR #78, opened 2021-05-29
plugins: add go plugin (CRAFT-48)

cmatsuoka:
Add a plugin for go projects using go.mod.  Plugin build definitions
are based on the Snapcraft V2 go plugin.
    
Co-authored-by: Sergio Schvezov <sergio.schvezov@canonical.com>

-----

### DIFF
--- a/craft_parts/plugins/go_plugin.py
+++ b/craft_parts/plugins/go_plugin.py
@@ -123,13 +123,11 @@ class GoPlugin(Plugin):
 
     def get_build_packages(self) -> Set[str]:
         """Return a set of required packages to install in the build environment."""
-        return {"gcc"}
+        return set()
 
     def get_build_environment(self) -> Dict[str, str]:
         """Return a dictionary with the environment to use in the build step."""
         return {
-            "PARTS_GO_LDFLAGS": "-ldflags -linkmode=external",
-            "CGO_ENABLED": "1",
             "GOBIN": "{}/bin".format(self._part_info.part_install_dir),
         }
 
@@ -144,8 +142,5 @@ class GoPlugin(Plugin):
 
         return [
             "go mod download",
-            'go install -p "{}" {} ${{PARTS_GO_LDFLAGS}} ./...'.format(
-                self._part_info.parallel_build_count,
-                tags,
-            ),
+            f'go install -p "{self._part_info.parallel_build_count}" {tags} ./...',
         ]
--- a/craft_parts/plugins/go_plugin.py
+++ b/craft_parts/plugins/go_plugin.py
@@ -22,9 +22,9 @@ from typing import Any, Dict, List, Optional, Set, cast
 
 from craft_parts import errors
 
+from . import validator
 from .base import Plugin, PluginModel, extract_plugin_properties
 from .properties import PluginProperties
-from .validator import PluginEnvironmentValidator
 
 logger = logging.getLogger(__name__)
 
@@ -53,7 +53,7 @@ class GoPluginProperties(PluginProperties, PluginModel):
         return cls(**plugin_data)
 
 
-class GoPluginEnvironmentValidator(PluginEnvironmentValidator):
+class GoPluginEnvironmentValidator(validator.PluginEnvironmentValidator):
     """Check the execution environment for the Go plugin.
 
     :param part_name: The part whose build environment is being validated.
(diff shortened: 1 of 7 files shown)


craft_parts/plugins/go_plugin.py
```diff
@@ -0,0 +1,102 @@
...
+class GoPlugin(Plugin):
+    """The go plugin can be used for go projects using go.mod.
+
+    This plugin uses the common plugin keywords as well as those for "sources".
+    For more information check the 'plugins' topic for the former and the
+    'sources' topic for the latter.
+
+    Additionally, this plugin uses the following plugin-specific keywords:
+
+    - ``go-channel``
```
sergiusens, line 59: I am going to deviate from this PR a bit and ask the question. what if instead we add a method to verify that the "required" dependencies are satisfied and error with a suggestion to solve the missing bits?

cmatsuoka:
Converted to draft to experiment with environment validation.

craft_parts/plugins/go_plugin.py
cjp256: I like that approach. :+1:

craft_parts/plugins/go_plugin.py
```diff
@@ -0,0 +1,102 @@
...
+      (string, default: latest/stable)
```
cjp256, line 60: I forgot that go-channel is required in v2.  Interesting...

cjp256 (approved):

craft_parts/plugins/go_plugin.py
cmatsuoka: Plugin environment validation implemented in PR #150, and updated the go plugin to use it.

craft_parts/plugins/go_plugin.py
cmatsuoka: This was removed now that we don't require the go snap.

craft_parts/plugins/go_plugin.py
```diff
@@ -0,0 +1,151 @@
...
+
+    - ``go-buildtags``
+      (list of strings)
+      Tags to use during the go build. Default is not to use any build tags.
+    """
+
+    properties_class = GoPluginProperties
+    validator_class = GoPluginEnvironmentValidator
+
+    def get_build_snaps(self) -> Set[str]:
```
sergiusens, line 120: we forgot to remove this from the API

craft_parts/plugins/go_plugin.py
```diff
@@ -0,0 +1,151 @@
...
+        """Return a set of required snaps to install in the build environment."""
+        return set()
+
+    def get_build_packages(self) -> Set[str]:
+        """Return a set of required packages to install in the build environment."""
+        return {"gcc"}
```
sergiusens, line 126: This should not be here, pure go binaries don't require gcc.

craft_parts/plugins/go_plugin.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def get_build_environment(self) -> Dict[str, str]:
+        """Return a dictionary with the environment to use in the build step."""
+        return {
+            "PARTS_GO_LDFLAGS": "-ldflags -linkmode=external",
+            "CGO_ENABLED": "1",
```
sergiusens, line 132: We should probably remove these two. The first one is a Snapcraft workaround for classic confinement which is leaking into the pureness of this library :-)

The second one is more debatable and maybe the reason for gcc in the former.

craft_parts/plugins/validator.py
```diff
@@ -33,8 +33,14 @@ class PluginEnvironmentValidator:
     deb package, a snap, built from sources or even built by a different
     part. Plugin environment validators allow a plugin to ensure its
     execution environment is correct before building a part.
+
+    :param part_name: The part whose build environment is being validated.
+    :param env: A string containing the build step environment setup.
     """
 
+    COMMAND_NOT_FOUND = 127
```
sergiusens, line 41: is this here due to a missing rebase or something?

craft_parts/plugins/go_plugin.py
cmatsuoka: I will remove the method from the base, implementations and tests in a follow-up PR.

craft_parts/plugins/go_plugin.py
cmatsuoka: We can change the plugin to be pure go and address issues (possibly in a different way) as they appear.

craft_parts/plugins/go_plugin.py
cmatsuoka: Removing to let the plugin work with pure go.

craft_parts/plugins/validator.py
cmatsuoka: I moved the definition from each validator implementation to the base class, but perhaps it would be better to have it defined in package level.

sergiusens (approved):

PR #81, opened 2021-06-11
packages: add test to ensure unnecessary normalize not called (CRAFT-221)

cmatsuoka:
When unpacking stage packages, it's not necessary to normalize unpacked
files if nothing was unpacked. This test verifies if no such call is
made in this case.

This is a follow-up to PR #76, and adds the normalization call test included in
https://github.com/snapcore/snapcraft/pull/3534.

-----

### DIFF
--- a/tests/unit/packages/test_deb.py
+++ b/tests/unit/packages/test_deb.py
@@ -191,6 +191,22 @@ class TestPackages:
             )
         assert raised.value.message == "foo"
 
+    def test_unpack_stage_packages_dont_normalize(self, tmpdir, mocker):
+        packages_path = Path(tmpdir, "pkg")
+        install_path = Path(tmpdir, "install")
+
+        mock_normalize = mocker.patch("craft_parts.packages.normalize.normalize")
+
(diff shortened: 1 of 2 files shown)


cjp256 (approved):

tests/unit/packages/test_deb.py
```diff
@@ -191,6 +191,22 @@ def test_get_package_fetch_error(self, tmpdir, fake_apt_cache):
             )
         assert raised.value.message == "foo"
 
+    def test_unpack_stage_packages_dont_normalize(self, tmpdir, mocker):
+        packages_path = Path(tmpdir, "pkg")
```
sergiusens, line 195: maybe add a not that an empty directory for stage-packages is being created

sergiusens (approved):
Approved, with a minor request for notes

tests/unit/packages/test_deb.py
cmatsuoka: There's a note about it being empty right before the actual directory creation in lines 200-201.

PR #82, opened 2021-06-14
executor: set step environment variables (CRAFT-272)

cmatsuoka:
Add variables with project and part-related content to the step execution
environment. The variables are a subset of the variables set by Snapcraft
and are prefixed by the name of the application using craft-parts.
Additional application-specific environment variables can be defined by
the application in the pre-step callback.

The application_name string must be non-empty and contain only letters,
digits or underscore characters. All other characters will be converted
to underscores. If the application name starts with a digit, an
underscore will be prepended to the step environment variable names.

-----


craft_parts/lifecycle_manager.py
```diff
@@ -65,7 +67,10 @@ def __init__(
         extra_build_packages: List[str] = None,
         **custom_args,  # custom passthrough args
     ):
-        # TODO: validate or slugify application name
+        if not application_name:
+            raise ValueError("application name cannot be empty")
+
+        application_name = re.sub("[^0-9A-Za-z_]+", "_", application_name)
```
facundobatista, line 73: Nice!

facundobatista (approved):
Wonderful!

craft_parts/executor/environment.py
```diff
@@ -140,6 +140,35 @@ def _basic_environment_for_part(part: Part, *, step_info: StepInfo) -> Dict[str,
     return part_environment
 
 
+def _get_step_environment(step_info: StepInfo) -> Dict[str, str]:
+    """Set environment variables from step information.
+
+    Variable names are prefixed by the application name and follows the
+    list of environment variables set by Snapcraft.
```
sergiusens, line 147:
```suggestion
    A list of environment variables prefixed by the application name.
```

Declare them if you want too :-)

craft_parts/executor/environment.py
```diff
@@ -140,6 +140,35 @@ def _basic_environment_for_part(part: Part, *, step_info: StepInfo) -> Dict[str,
...
+
+    :param step_info: Information about the current step.
+
+    :return: A dictionary containing environment variables and values.
+    """
+    # The application name contains only alphanumeric characters or underscore.
```
sergiusens, line 153: is this being checked for?

craft_parts/lifecycle_manager.py
sergiusens: should we not match instead and raise? If not, maybe instead mention this normalization process in the doc string

sergiusens (approved):
Looks good, just a couple of suggestions to look into before merging

craft_parts/executor/environment.py
sergiusens: it is, further bellow, maybe this requires a new check as the attribute is mutable?

cmatsuoka:
Agree with @sergiusens, it's better to validate and raise instead of automatically transforming. Auto-transforming could lead to a number of potentially bad situations, like different original values transforming to the same string. Also it could be convenient to allow the user to optionally specify the variable prefix, with the default being the application name. Will refactor accordingly.

craft_parts/lifecycle_manager.py
cmatsuoka: The transformation magic is cool but I agree that we shouldn't change what the application supplies. That could create problems at runtime.

sergiusens (approved):

PR #83, opened 2021-06-22
sequencer: improve dirty state report message (CRAFT-302)

cmatsuoka:
Clarify the message to include the dependency step name in the reason
why the step is being re-executed.

-----

### DIFF
--- a/craft_parts/state_manager/reports.py
+++ b/craft_parts/state_manager/reports.py
@@ -136,7 +136,7 @@ class DirtyReport:
             else:
                 part_name = self.changed_dependencies[0].part_name
                 step_name = self.changed_dependencies[0].step.name.lower()
-                reasons.append(f"part {part_name!r} {step_name}")
+                reasons.append(f"{step_name} for part {part_name!r}")
 
         if not reasons:
             return ""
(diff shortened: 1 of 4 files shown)


craft_parts/state_manager/reports.py
```diff
@@ -134,7 +134,9 @@ def reason(self) -> str:
             if reasons_count > 1 or len(self.changed_dependencies) > 1:
                 reasons.append("dependencies")
             else:
-                reasons.append(f"{self.changed_dependencies[0].part_name!r}")
+                part_name = self.changed_dependencies[0].part_name
+                step_name = self.changed_dependencies[0].step.name.lower()
+                reasons.append(f"part {part_name!r} {step_name}")
```
sergiusens, line 139: Is there anything else aside from "changed"?

```suggestion
                reasons.append(f"{step_name} for part {part_name!r}")
```

sergiusens (approved):
Looks better, I added another suggestion.

facundobatista (approved):
Nice!

craft_parts/state_manager/reports.py
cmatsuoka: It's only that for now, we can improve it in future iterations.

PR #85, opened 2021-06-23
api: pass cache directory to lifecycle manager (CRAFT-308)

cmatsuoka:
Instead of using a path based on the application name, let the
application choose the full path for file caching. This gives the application
control over the location of the file cache, allowing it to set the path
according to its own policies and making it easier to remove the
cache if needed.

-----


craft_parts/infos.py
```diff
@@ -47,16 +52,25 @@ def __init__(
...
         arch: str = "",
         base: str = "",
         parallel_build_count: int = 1,
         project_dirs: ProjectDirs = None,
         **custom_args,  # custom passthrough args
     ):
+        if cache_dir:
+            cache_path = Path(cache_dir)
+        else:
+            cache_path = BaseDirectory.save_cache_path(
```
sergiusens, line 65: why not just fail or require it?

craft_parts/lifecycle_manager.py
```diff
@@ -40,11 +41,13 @@ class LifecycleManager:
...
         output generated by PyYAML's ``yaml.load``.
     :param application_name: A unique non-empty identifier for the application
-        using Craft Parts. This string will be used as segregated directory path
-        when creating persistent data that shouldn't be shared with other
-        applications and as a prefix to step environment variable names. Valid
-        application names contain upper and lower case letters, underscores
-        or numbers, and must start with a letter.
+        using Craft Parts. The application name is used as a prefix to step
+        environment variables, and as a component of the cache directory path if
+        it's not set by the application. Valid application names contain upper
```
sergiusens, line 46: the fall back is nice for backwards compatibility cases, but we don't have that. I would personally just remove it.

sergiusens (approved):
Nice, this looks like a partial migration to pathlib :-)

cjp256 (approved):
Nice!

craft_parts/infos.py
cjp256: I think it'd best to require it too. Maybe set a default parameter?

At the very least, use `None` instead of `""` for optional, if being kept.  :D

craft_parts/lifecycle_manager.py
cmatsuoka: Removed.

PR #87, opened 2021-06-25
platform: compatibility fixes for python 3.7 and 3.9 (CRAFT-315)

cmatsuoka:
Make craft-parts compatible with Python 3.7. Also fix tests to accept
error messages that changed in Python 3.9, and update tox configuration.

-----

### DIFF
--- a/tests/unit/sources/test_base.py
+++ b/tests/unit/sources/test_base.py
@@ -14,6 +14,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import re
 from pathlib import Path
 
 import pytest
@@ -69,8 +70,13 @@ class TestSourceHandler:
             FaultySource(  # type: ignore
                 source=Path(), part_src_dir=Path(), cache_dir=Path()
             )
-        assert str(raised.value) == (
-            "Can't instantiate abstract class FaultySource with abstract methods pull"
+        assert (
+            re.match(
+                "^Can't instantiate abstract class FaultySource with "
+                "abstract methods? pull$",
+                str(raised.value),
+            )
+            is not None
         )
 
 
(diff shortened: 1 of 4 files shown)


sergiusens (approved):

tests/unit/sources/test_base.py
```diff
@@ -69,8 +70,13 @@ class FaultySource(SourceHandler):
             FaultySource(  # type: ignore
                 source=Path(), part_src_dir=Path(), cache_dir=Path()
             )
-        assert str(raised.value) == (
-            "Can't instantiate abstract class FaultySource with abstract methods pull"
+        assert (
```
facundobatista, line 73: The `pytest.raises` function has a `match` keyword exactly for this use. So, you could do:

```
expected = "^Can't instantiate abstract class FaultySource with abstract methods? pull$"
with pytest.raises(TypeError, match=expected) as raised:
    ...
```

tests/unit/sources/test_base.py
```diff
@@ -226,7 +232,11 @@ class FaultyFileSource(FileSourceHandler):
             FaultyFileSource(
                 source=None, part_src_dir=None, cache_dir=Path()  # type: ignore
             )
-        assert str(raised.value) == (
-            "Can't instantiate abstract class FaultyFileSource with abstract "
-            "methods provision"
+        assert (
```
facundobatista, line 235: Same than above.

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py38
+envlist = py37,py38,py39
```
facundobatista, line 2: Awesome.

facundobatista (approved):
Nice! Take a look at a suggestion inside.

tests/unit/sources/test_base.py
cmatsuoka: Nice! Thanks

PR #89, opened 2021-07-08
packaging: update setup script (CRAFT-281)

cmatsuoka:
Update the contents of setup.py, following the pattern used in
craft-providers.

-----


requirements-focal.txt
```diff
@@ -0,0 +1 @@
+http://archive.ubuntu.com/ubuntu/pool/main/p/python-apt/python-apt_2.0.0ubuntu0.20.04.5.tar.xz; sys_platform == 'linux'
```
cjp256, line 1: IMO, I wouldn't pin this req at all.  Depend on the host for it (python3-apt).

cjp256: Separately if craft-parts is to support non-Ubuntu bases, it shouldn't be a hard requirement either unless on deb/ubuntu.

cmatsuoka: Using python3-apt should make things easier.

sergiusens: this is requirements-focal.txt, so not really about not Ubuntu. When we support other distros the "import apt" should be lazy loaded,  catched and set to a stub for the non compatible platforms, or we can build the apt libs ourselves (at the application level).

python3-apt from debs would require setting the venv to also look at dist-packages.

There's a slight clarification to be made here too, it is not that python-apt is tied to a given Ubuntu libapt but to a given version of the library, if we built it, it could move out of band from the base we use (for the snap case or any other foreign os case)

setup.cfg
```diff
@@ -16,7 +16,7 @@ search = __version__ = "{current_version}"
 replace = __version__ = "{new_version}"
 
 [bdist_wheel]
-universal = 1
+universal = 0
```
sergiusens, line 19: interesting, what would the platform definitions be to contain a potential binary wheel to a given release?

sergiusens (approved):

cjp256 (approved):

PR #91, opened 2021-07-09
many: send craft-parts output to logger (CRAFT-365)

cmatsuoka:
Craft-parts shouldn't pollute the application output with its own
messages. This is an interim solution, send the output to the log
until the user message handling library is available.

Output example ("Execute" messages are printed by the cli utility):
```
$ python3 -mcraft_parts -fparts-mpg123.yaml 
WARNING:craft_parts.packages.apt_cache:libasound-dev is a virtual package, use non-virtual packages for deterministic results.
Execute: Pull mpg123
Execute: Build mpg123
Execute: Stage mpg123
Execute: Prime mpg123
```

-----

### DIFF
--- a/craft_parts/lifecycle_manager.py
+++ b/craft_parts/lifecycle_manager.py
@@ -65,7 +65,7 @@ class LifecycleManager:
         *,
         application_name: str,
         cache_dir: Union[Path, str],
-        work_dir: str = ".",
+        work_dir: Union[Path, str] = ".",
         arch: str = "",
         base: str = "",
         parallel_build_count: int = 1,
--- a/craft_parts/packages/apt_cache.py
+++ b/craft_parts/packages/apt_cache.py
@@ -235,14 +235,16 @@ class AptCache(ContextDecorator):
                 continue
 
             try:
-                dl_path = self.fetch_binary(package.candidate, download_path)
+                dl_path = package.candidate.fetch_binary(
+                    str(download_path), progress=self.progress
+                )
             except apt.package.FetchError as err:
                 raise errors.PackageFetchError(str(err))
 
             if package.candidate is None:
                 raise errors.PackageNotFound(package.name)
 
-            downloaded.append((package.name, package.candidate.version, dl_path))
+            downloaded.append((package.name, package.candidate.version, Path(dl_path)))
         return downloaded
 
     def get_installed_packages(self) -> Dict[str, str]:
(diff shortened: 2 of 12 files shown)


craft_parts/packages/apt_cache.py
```diff
@@ -43,6 +43,36 @@
...
+    def __init__(self):
+        self._id = 1
+
+    def fail(self, item: apt.apt_pkg.AcquireItemDesc) -> None:
+        """Handle failed item."""
+        apt.progress.base.AcquireProgress.fail(self, item)
+        if item.owner.status == item.owner.STAT_DONE:
+            logger.debug("Ign %s", item.description)
+        else:
+            logger.error("Err %s", item.description)
```
cjp256, line 58: Consider logging it all to debug and use the exception as the interface to provide the error message back to the app. Otherwise may get a little tricky to filter.

craft_parts/packages/apt_cache.py
```diff
@@ -339,6 +365,46 @@ def update(self) -> None:
 
     # pylint: enable=attribute-defined-outside-init
 
+    def fetch_binary(self, package_candidate, destination: Path) -> Path:
```
cjp256, line 368: Can we pass the custom progress logger to the real fetch_binary() and drop this?

cmatsuoka: Yes, this was a leftover from a previous attempt to reduce package installation verbosity.

craft_parts/lifecycle_manager.py
```diff
@@ -65,7 +65,7 @@ def __init__(
         *,
         application_name: str,
         cache_dir: Union[Path, str],
-        work_dir: str = ".",
+        work_dir: Union[Path, str] = ".",
```
facundobatista, line 68: Maybe to be improved in a next branch: we shouldn't have a duality Path/str for paths. If we want to use Path, let's use Path everywhere. Wherever we need to pass them or receive them as strings, let's convert them as soon as possible, and not leak them as strings in the rest of the code.

craft_parts/utils/os_utils.py
```diff
@@ -302,3 +302,22 @@ def version_codename(self) -> str:
...
+        command,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.STDOUT,
+        universal_newlines=True,
+        **kwargs,
+    ) as proc:
+        if not proc.stdout:
+            return
+        for line in iter(proc.stdout.readline, ""):
+            log_func(line.strip())
```
facundobatista, line 319: Why not log here in debug (without receiving the level from outside) and just use this function in the other modules, instead of defining two other intermediate functions?

I mean, everything will be simpler in how I propose it, why have it more complex? which is the benefit?

facundobatista: Maybe we should prefix each line with something, so visually we understand that it's the output of the command? I used `:: ` in other codes and it's super clear.

facundobatista (changes requested):
Annotated some details, thanks!

craft_parts/utils/os_utils.py
cmatsuoka: If call debug directly in `os_utils`, everything will be logged as originating from the `utils.os_util` package, not from the package where the code is actually running (`executor.step_handler`, `packages.deb`). This is temporary code anway, the final implementation should drop `process_run()` entirely and allow us to set stdout, stderr directly from `subprocess.run()`.

cmatsuoka: Will add a prefix.

craft_parts/lifecycle_manager.py
cmatsuoka: Will normalize to use Path everywhere in CRAFT-43.

sergiusens (approved):

facundobatista (approved):
Wonderful, thanks!

PR #92, opened 2021-07-12
parts: add validators for stage and prime filesets (CRAFT-366)

cmatsuoka:
Validate stage and prime to detect absolute paths on unmarshal, instead
of doing it at execution time when the filter is used.

-----

### DIFF
--- a/craft_parts/parts.py
+++ b/craft_parts/parts.py
@@ -67,7 +67,7 @@ class PartSpec(BaseModel):
     @validator("stage_files", "prime_files", each_item=True)
     def validate_relative_path_list(cls, item):
         """Check if the list does not contain empty of absolute paths."""
-        assert item != "", f"{item!r} must be a relative path (cannot be empty)"
+        assert item != "", "path cannot be empty"
         assert (
             item[0] != "/"
         ), f"{item!r} must be a relative path (cannot start with '/')"
--- a/craft_parts/parts.py
+++ b/craft_parts/parts.py
@@ -19,7 +19,7 @@
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Sequence, Set
 
-from pydantic import BaseModel, Field, ValidationError
+from pydantic import BaseModel, Field, ValidationError, validator
 
 from craft_parts import errors
 from craft_parts.dirs import ProjectDirs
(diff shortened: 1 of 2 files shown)


cmatsuoka:
Output example when running an application with an absolute path in the prime files list:
```
$ charmcraft pack --destructive-mode
Part 'charm' validation failed.
'prime',0: '/foo' must be a relative path (cannot start with '/')
Review part 'charm' and make sure it's correct. (full execution logs in '/tmp/charmcraft-log-18qg2t4l')
```

cjp256 (approved):

craft_parts/parts.py
```diff
@@ -63,6 +63,18 @@ class Config:
         allow_mutation = False
         alias_generator = lambda s: s.replace("_", "-")  # noqa: E731
 
+    # pylint: disable=no-self-argument,no-self-use
+    @validator("stage_files", "prime_files", each_item=True)
+    def validate_relative_path_list(cls, item):
+        """Check if the list does not contain empty of absolute paths."""
+        assert item != "", f"{item!r} must be a relative path (cannot be empty)"
```
facundobatista, line 70: The text between parentesis should be to explain/exemplify the other part. In this case they are two different reasons (both valid). What about "{} must be a relative path and cannot be empty", or "{} must be a relative non-empty path"? Or maybe, as below it's tested the "relativeness" itself, what about "{} path cannot be empty"?

facundobatista (approved):
+1, but please check the comment about the message.

facundobatista:
> Output example when running an application with an absolute path in the prime files list:
> 
> ```
> $ charmcraft pack --destructive-mode
> Part 'charm' validation failed.
> 'prime',0: '/foo' must be a relative path (cannot start with '/')
> Review part 'charm' and make sure it's correct. (full execution logs in '/tmp/charmcraft-log-18qg2t4l')
> ```
> 
> The pydantic validation error message will be reformatted in a separate PR to be more human-readable.

I approved this, but something bothers me: the error here has a different style than the other charmcraft config errors. This is what you say that you will fix later?

cmatsuoka:
The actual error message 

> > Output example when running an application with an absolute path in the prime files list:
> > ```
> > $ charmcraft pack --destructive-mode
> > Part 'charm' validation failed.
> > 'prime',0: '/foo' must be a relative path (cannot start with '/')
> > Review part 'charm' and make sure it's correct. (full execution logs in '/tmp/charmcraft-log-18qg2t4l')
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The pydantic validation error message will be reformatted in a separate PR to be more human-readable.
> 
> I approved this, but something bothers me: the error here has a different style than the other charmcraft config errors. This is what you say that you will fix later?

Yes, we must adjust the error message style. I already have a PR to change the validation error formatting, or we can rewrap the error to handle brief/details/resolution error parts if needed.

PR #93, opened 2021-07-13
lifecycle: add part validator (CRAFT-367)

cmatsuoka:
Validate the given part data against the common parts model and the
part's plugin model before creating the lifecycle manager. This allows
the validation to happen on the host while the lifecycle manager is
created later in a managed environment.

-----


cmatsuoka:
Validator was simplified and moved to the application, no need to add this to craft-parts for now.

PR #94, opened 2021-07-13
errors: reformat pydantic validation error (CRAFT-368)

cmatsuoka:
Change formatting of pydantic errors to be more human-readable. The
new format is similar to the one used in Charmcraft. Specifically,
this changes

```
'foo',1: some error happened
'bar': field required
'baz': extra fields not permitted
```
to
```
- some error happened in field 'foo[1]'
- field 'bar' is required'
- extra field 'baz' not permitted
```
    
Co-authored-by: Chris Patterson <chris.patterson@canonical.com>

-----

### DIFF
--- a/craft_parts/errors.py
+++ b/craft_parts/errors.py
@@ -128,10 +128,36 @@ class PartSpecificationError(PartsError):
             if not (loc and msg) or not isinstance(loc, tuple):
                 continue
 
-            fields = ",".join([repr(entry) for entry in loc])
-            formatted_errors.append(f"{fields}: {msg}")
+            field = cls._format_loc(loc)
+            if msg == "field required":
+                formatted_errors.append(f"- field {field!r} is required")
+            elif msg == "extra fields not permitted":
+                formatted_errors.append(f"- extra field {field!r} not permitted")
+            else:
+                formatted_errors.append(f"- {msg} in field {field!r}")
 
         return cls(part_name=part_name, message="\n".join(formatted_errors))
+
+    @classmethod
+    def _format_loc(cls, loc):
+        """Format location."""
+        loc_parts = []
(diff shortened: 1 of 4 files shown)


cjp256 (approved):
LGTM

facundobatista (commented):
Code here looks ok. What bothers me is that we have TWO places where messages are formatted. Here and in the code that uses this lib. Shouldn't the lib just raise the error with proper components and solve the formatting in ONE place? Think of the case of needing to change the formatting, for example, what would happen when this lib is used by several apps?

sergiusens (approved):
There is nothing wrong with providing a nice error message by default using our current constructs, but I agree with @facundobatista but @cmatsuoka verified that this not intended to be used by Charmcraft itself, just when using craft-parts in isolation

cmatsuoka:
> Code here looks ok. What bothers me is that we have TWO places where messages are formatted. Here and in the code that uses this lib. Shouldn't the lib just raise the error with proper components and solve the formatting in ONE place? Think of the case of needing to change the formatting, for example, what would happen when this lib is used by several apps?

I changed validation handling in the charmcraft craft-parts integration so that charmcraft gets raw errors instead of pre-formatted pydantic validation errors, so now all formatting in the initial validation is performed by charmcraft. The changes in this PR will be used by craft-parts to present errors when running stand-alone or if a validation error happens during the initialization of the lifecycle manager (this should never happen because charmcraft already validated parts).

I'll place this PR on hold until we propose the final craft-parts integration so we can be sure that no errors from craft-parts are leaking during the initial validation (it's harder to verify it with this PR merged because the error messages are in the same style).

facundobatista:
Wonderful
, thanks!

PR #96, opened 2021-07-16
many: fix source ignore patterns (CRAFT-375)

cmatsuoka:
Work directories and files generated by the application in the current
directory (which may be a local source for plugins) should be ignored when
pulling sources. This change fixes a number of issues related to file
patterns to ignore:

- Don't pull local sources matching the patterns to ignore, otherwise the
  newly generated application file may be also considered a source file.
- Allow the application to configure ignore patterns to match files it
  generated (such as `*.snap` for snapcraft, or `*.charm` for charmcraft).
- When a work directory is set, ignore it instead of the standard parts,
  prime, stage directories.

-----

### DIFF
--- a/craft_parts/sources/local_source.py
+++ b/craft_parts/sources/local_source.py
@@ -16,14 +16,19 @@
 
 """The local source handler and helpers."""
 
+import contextlib
 import functools
 import glob
+import logging
 import os
+from pathlib import Path
 from typing import List, Optional
 
 from craft_parts.utils import file_utils
 
 from .base import SourceHandler
+
+logger = logging.getLogger(__name__)
 
 # TODO: change file operations to use pathlib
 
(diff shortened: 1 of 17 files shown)


craft_parts/sources/local_source.py
```diff
@@ -36,14 +41,19 @@ def __init__(self, *args, copy_function=file_utils.link_or_copy, **kwargs):
         self.source_abspath = os.path.abspath(self.source)
         self.copy_function = copy_function
 
-        ignore_patterns = [
-            self._dirs.parts_dir.name,
-            self._dirs.stage_dir.name,
-            self._dirs.prime_dir.name,
-            "*.snap",  # FIXME: this should be specified by the application
-        ]
+        if self._dirs.work_dir.resolve() == Path.cwd():
```
cjp256, line 44: Would this break if cwd is not the project dir?  Perhaps this should instead check to see if self.source_abspath matches workdir?

cmatsuoka: Good catch! Changed to check if workdir matches source dir, and fixed an inconsistent test which placed workdir outside source dir but expected it to be inside.

sergiusens (approved):
Nice, this could eventually be exposed as `source-ignore` or similar :-)

craft_parts/sources/local_source.py
```diff
@@ -41,13 +41,16 @@ def __init__(self, *args, copy_function=file_utils.link_or_copy, **kwargs):
...
+        if self._dirs.work_dir.resolve() == Path(self.source_abspath):
+            # ignore parts/stage/dir if source dir matches workdir
             self._ignore_patterns.append(self._dirs.parts_dir.name)
             self._ignore_patterns.append(self._dirs.stage_dir.name)
             self._ignore_patterns.append(self._dirs.prime_dir.name)
         else:
+            # otherwise check if work_dir inside source dir
             with contextlib.suppress(ValueError):
-                rel_work_dir = self._dirs.work_dir.relative_to(Path.cwd())
+                rel_work_dir = self._dirs.work_dir.relative_to(self.source_abspath)
```
cjp256, line 52: I know I suggested source_abspath, but upon further consideration should it be resolve()d for consistency too?  I think you recently changed it to resolve?

cjp256 (approved):

PR #97, opened 2021-07-19
utils: fix snap environment detection and add tests (CRAFT-376)

cmatsuoka:
Snap environment detection returned false positives and was not being
properly tested. Fix snap environment detection and add tests for
`is_snap()` and `is_inside_container()`.

-----

### DIFF
--- a/craft_parts/utils/os_utils.py
+++ b/craft_parts/utils/os_utils.py
@@ -172,13 +172,9 @@ def is_snap(*, application_name: Optional[str] = None) -> bool:
     if application_name:
         res = snap_name == application_name
     else:
-        res = snap_name is not None
+        res = snap_name != ""
 
-    logger.debug(
-        "craft_parts is is snap: %s, SNAP_NAME set to %s",
-        res,
-        snap_name,
-    )
+    logger.debug("is_snap: %s, SNAP_NAME set to %s", res, snap_name)
 
     return res
 
--- a/craft_parts/utils/os_utils.py
+++ b/craft_parts/utils/os_utils.py
@@ -162,17 +162,14 @@ def is_dumb_terminal() -> bool:
     return not is_stdout_tty or is_term_dumb
 
 
-def is_snap(*, application_name: Optional[str] = None) -> bool:
+def is_snap(application_name: str) -> bool:
     """Verify whether we're running as a snap.
 
     :application_name: The snap application name. If provided, check
         if it matches the snap name.
     """
-    snap_name = os.environ.get("SNAP_NAME", "")
-    if application_name:
-        res = snap_name == application_name
-    else:
-        res = snap_name != ""
+    snap_name = os.environ.get("SNAP_NAME")
+    res = snap_name == application_name
 
     logger.debug("is_snap: %s, SNAP_NAME set to %s", res, snap_name)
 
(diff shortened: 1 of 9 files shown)


craft_parts/utils/os_utils.py
```diff
@@ -172,13 +172,9 @@ def is_snap(*, application_name: Optional[str] = None) -> bool:
     if application_name:
         res = snap_name == application_name
     else:
-        res = snap_name is not None
+        res = snap_name != ""
```
cjp256, line 175: Generally i'd suggest using the default of None to distinguish if envvar is actually set to empty string.

```
$ FOO="" python -c 'import os; print(repr(os.environ.get("FOO")))'
''
$ python -c 'import os; print(os.environ.get("FOO"))'
None
```

But in this case it's non-consequential, so I'm ok as-is.

cjp256 (approved):

craft_parts/utils/os_utils.py
cmatsuoka: I used the empty string to avoid an extra test before comparing to `application_name`.

craft_parts/utils/os_utils.py
```diff
@@ -172,13 +172,9 @@ def is_snap(*, application_name: Optional[str] = None) -> bool:
     if application_name:
         res = snap_name == application_name
```
sergiusens, line 173: Does this not work?
```suggestion
    is_snap = application_name == snap_name:
```

craft_parts/utils/os_utils.py
```diff
@@ -172,13 +172,9 @@ def is_snap(*, application_name: Optional[str] = None) -> bool:
...
     else:
-        res = snap_name is not None
+        res = snap_name != ""
 
-    logger.debug(
-        "craft_parts is is snap: %s, SNAP_NAME set to %s",
-        res,
-        snap_name,
-    )
+    logger.debug("is_snap: %s, SNAP_NAME set to %s", res, snap_name)
```
sergiusens, line 177:
```suggestion
    logger.debug("is_snap: %s, SNAP_NAME set to %s ", is_snap, snap_name)
```

craft_parts/utils/os_utils.py
sergiusens: what if SNAP_NAME is set but does not match application name?

- Launch emacs/vscode snap
- start vterm
- env will have SNAP_NAME, but not the SNAP_NAME we want

we don't care if SNAP_NAME is set, only from the point of view of it matching the application name, I believe my simpler check does the trick.

sergiusens (changes requested):

craft_parts/utils/os_utils.py
cmatsuoka: This is an interesting issue because craft-parts doesn't know the snap name of the application it's running under, so perhaps it should check if SNAP_NAME is set to something, regardless of the name.

sergiusens:
Closing in favor of #99

cmatsuoka:
Updated to use `application_name` to check if we're running as a snap. A few notes on this change:
- Updated `is_snap()` to require an application name passed, it's not optional anymore.
- Changed package manager configuration to be executed only once (when the lifecycle manager is created).
- Added an optional `application_package_name` parameter which can be used if `application_name` can't match the snap name.

sergiusens (approved):

PR #98, opened 2021-07-21
utils: provide uname call with non-gnu options (CRAFT-382)

cmatsuoka:
This fixes the uname call in MacOS. Problem found during charmcraft tests.

-----

### DIFF
--- a/craft_parts/utils/os_utils.py
+++ b/craft_parts/utils/os_utils.py
@@ -20,6 +20,7 @@ import contextlib
 import logging
 import os
 import subprocess
+import sys
 import time
 from pathlib import Path
 from typing import Callable, Dict, List, Optional, Union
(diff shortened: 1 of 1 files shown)


craft_parts/utils/os_utils.py
```diff
@@ -212,7 +213,10 @@ def get_system_info() -> str:
                 "--hardware-platform",
                 "--operating-system",
             ]
-        )
+        else:
+            uname_cmd = ["uname", "-s", "-r", "-v", "-m", "-p"]
```
facundobatista, line 217: It's sad that in MacOS `uname` does not accept `-i` or `-o`.

facundobatista (approved):
Great!

sergiusens (approved):

PR #99, opened 2021-07-22
packages: interface to configure apt (CRAFT-376)

cjp256:
Refactor apt_cache() to expose all performed configuration for
apt when using stage-packages in an extensible manner.  Set
defaults that match the current current behavior.

Future work may change the defaults and remove the unnecessary
re-configuration of apt (needs only be done once per-process).

Signed-off-by: Chris Patterson <chris.patterson@canonical.com>

-----

### DIFF
--- a/craft_parts/packages/apt_cache.py
+++ b/craft_parts/packages/apt_cache.py
@@ -18,6 +18,7 @@
 
 from __future__ import annotations
 
+import dataclasses
 import logging
 import os
 import re
@@ -32,8 +33,6 @@ import apt.package
 import apt.progress
 import apt.progress.base
 
-from craft_parts.utils import os_utils
-
 from . import errors
 from .base import get_pkg_name_parts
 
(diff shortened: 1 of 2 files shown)


sergiusens (changes requested):
I think the point was to make this independent of $SNAP as that can lead to spurious errors, the defaults could just be None and it is up to the application to change that. If we still want to use and environment variable, we should use a different one, something like CRAFT_PARTS_REPO_ROOT that the application should set

cjp256:
> I think the point was to make this independent of $SNAP as that can lead to spurious errors, the defaults could just be None and it is up to the application to change that. If we still want to use and environment variable, we should use a different one, something like CRAFT_PARTS_REPO_ROOT that the application should set

I constrained this PR to match existing behavior by default (with some sanity checks) so it is evident that this refactoring doesn't change behavior.  I plan a short follow up to selectively change defaults as needed.  However, each application should define their own AptStagePackageOptions.

The last bit on my TODO is to decide whether to make the default AptStagePackageOptions global, static to AptCache, or provide an interface to set/get the default (e.g. via lifecycle manager).  I'd like to have a quick chat about it first before making that change.

cmatsuoka:
I'm not a big fan of using an environment variable because it would be the only configuration done that way, with all other configuration parameters passed when creating the lifecycle manager. `PARTS_*` environment variables do exist, but they're mostly used internally (e.g. to set the ctl fifo path).

cmatsuoka:
Also it's interesting to create this configuration API considering that repo/packages will be made into a separate craft project at some point.

sergiusens:
We are going with the application_name concept, we can revisit this when looking closer at craft-repo

cmatsuoka:
Oops. Reopened the wrong PR.

PR #100, opened 2021-07-23
packages: remove sudo calls (CRAFT-384)

cmatsuoka:
Remove sudo calls from apt and snap install calls, and fail instead
if permission to run is insufficient. Sudo is not necessary in the
provider environment because the application already runs as root.

-----


sergiusens (approved):

sergiusens:
We will need a root check in the future to fail gracefully, but looks good for now

PR #102, opened 2021-07-29
tests: add tar source integration test (CRAFT-197)

cmatsuoka:
Make sure we can correctly read a valid tar source.

-----

### DIFF
--- a/tests/integration/sources/test_tar.py
+++ b/tests/integration/sources/test_tar.py
@@ -1,1 +1,74 @@
-# TODO: add tar source integration test after executor lands
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2021 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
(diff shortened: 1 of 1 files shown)


tests/integration/sources/test_tar.py
```diff
@@ -1 +1,75 @@
...
+        """\
+        parts:
+          foo:
+            plugin: make
+            source: foobar.tar.gz
+        """
+    )
+
+    parts = yaml.safe_load(_parts_yaml)
+    src = _LOCAL_DIR / "data" / "foobar.tar.gz"
```
cjp256, line 42: Might be nicer to create a tar as part of the test (or as fixture) rather than track the blob

cjp256 (approved):

sergiusens (approved):
I agree with Chris, this might also be a good reason to ammend the commit instead of pushing a new one :-)

tests/integration/sources/test_tar.py
cmatsuoka: Changed to create the tar file programmatically in the test.

cmatsuoka:
Tar file created programmatically in the test instead of using a binary asset.

PR #103, opened 2021-08-02
docs: update package documentation (CRAFT-37)

cmatsuoka:
Reorganize documentation layout to place a simple usage tutorial at
the start. Also improve docstrings and add lifecycle and class
layout diagrams.

Further documentation updates and fixes expected as package development
continues. Application plugin tutorial/example to be added after adding
environment verification to the plugin API.

-----


docs/examples.rst
```diff
@@ -0,0 +1,59 @@
+*****************************************
+Adding parts processing to an application
+*****************************************
+
+A simple example
+================
+
+To add parts processing to an application, one needs to instantiate the
```
sergiusens, line 8: This section should probably be imperative (no room for creativity) for reproduceable results.

sergiusens (approved):
Nice work, just a comment about the tutorial/example section

docs/examples.rst
cmatsuoka: Agree, change wording to use strict values instead of suggesting them.

PR #105, opened 2021-08-06
sources: verify deleted files in outdated check (CRAFT-440)

cmatsuoka:
When verifying if a pulled source is outdated, also check for files
or directories that might have been deleted from the source directory.
If this is the case, also remove the files or directories from the
part source directory.

-----


cjp256 (approved):
LGTM

sergiusens (approved):

PR #106, opened 2021-08-06
executor: fix rerun action to clean subsequent steps (CRAFT-441)

cmatsuoka:
Rerunning a part step should invalidate not only its state but also
the state of subsequent steps in the same part. This is correctly
implemented in action planning but not in execution. Fix the action
executor to clean all the necessary steps and not only the step being
re-executed.

-----


cjp256 (commented):
LGTM

sergiusens (approved):

cjp256 (approved):

PR #111, opened 2021-08-18
overlay: add overlayfs helpers (CRAFT-462)

cmatsuoka:
Add helpers to mount and unmount overlayfs filesystems, and to
identify special files and directories used in filesystem layering.

-----

### DIFF
--- a/craft_parts/overlays/errors.py
+++ b/craft_parts/overlays/errors.py
@@ -1,0 +1,53 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2021 Canonical Ltd.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License version 3 as published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Overlay error definitions."""
+
+from craft_parts import errors
+
+
+class OverlayError(errors.PartsError):
+    """Base class for overlay handler errors."""
+
+
+class OverlayMountError(OverlayError):
+    """Failed to mount an overlay filesystem.
+
+    :param mountpoint: The filesystem mount point.
+    :param message: The error message.
+    """
+
+    def __init__(self, mountpoint: str, message: str):
+        self.mountpoint = mountpoint
+        self.message = message
+        brief = f"Failed to mount overlay on {mountpoint}: {message}"
+
+        super().__init__(brief=brief)
+
+
+class OverlayUnmountError(OverlayError):
+    """Failed to unmount an overlay filesystem.
+
+    :param mountpoint: The filesystem mount point.
+    :param message: The error message.
+    """
+
+    def __init__(self, mountpoint: str, message: str):
+        self.mountpoint = mountpoint
+        self.message = message
+        brief = f"Failed to unmount {mountpoint}: {message}"
+
+        super().__init__(brief=brief)
--- a/craft_parts/overlays/overlay_fs.py
+++ b/craft_parts/overlays/overlay_fs.py
@@ -1,0 +1,109 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2021 Canonical Ltd.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License version 3 as published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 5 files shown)


craft_parts/overlays/overlay_fs.py
```diff
@@ -0,0 +1,115 @@
...
+logger = logging.getLogger(__name__)
+
+
+class OverlayFS:
+    """Linux overlayfs operations."""
+
+    def __init__(
+        self, *, lower_dir: Union[Path, List[Path]], upper_dir: Path, work_dir: Path
+    ):
+        if not isinstance(lower_dir, list):
```
cjp256, line 38: Consider making it `lower_dirs` and always require a list so this isn't responsible for the conversion (and to keep the interface a touch simpler).

cjp256 (changes requested):

craft_parts/overlays/overlay_fs.py
```diff
@@ -0,0 +1,115 @@
...
+    def unmount(self) -> None:
+        """Umount an overlayfs."""
+        if not self._mountpoint:
+            return
+
+        logger.debug("unmount overlayfs from %s", self._mountpoint)
+        try:
+            os_utils.umount(str(self._mountpoint))
+        except CalledProcessError as err:
+            raise errors.OverlayMountError(str(self._mountpoint), message=str(err))
```
cjp256, line 74:
```suggestion
            raise errors.OverlayUnmountError(str(self._mountpoint), message=str(err))
```

tests/unit/overlays/test_errors.py
```diff
@@ -0,0 +1,37 @@
...
+    )
+    assert err.details is None
+    assert err.resolution is None
+
+
+def test_overlay_unmount_error():
+    err = errors.OverlayUnmountError("/mountpoint", "something wrong happened")
+    assert err.mountpoint == "/mountpoint"
+    assert err.message == "something wrong happened"
+    assert err.brief == ("Failed to unmount /mountpoint: something wrong happened")
```
cjp256, line 35:
```suggestion
    assert err.brief == "Failed to unmount /mountpoint: something wrong happened"
```

craft_parts/overlays/overlay_fs.py
cmatsuoka: Oops. Copy-and-paste error detected.

tests/unit/overlays/test_errors.py
cmatsuoka: This is a strange construct. I wonder if black added it somehow.

craft_parts/overlays/overlay_fs.py
cmatsuoka: Changed to always receive a list.

tests/unit/overlays/test_errors.py
sergiusens: it was probably multiline with parens and black re condensed it to one but did not remove the parens

craft_parts/overlays/errors.py
```diff
@@ -0,0 +1,53 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Overlay error definitions."""
+
+from craft_parts import errors
+
+
+class OverlayError(errors.PartsError):
+    """Base class for overlay handler errors."""
+
```
sergiusens, line 24: How about moving most of the logic in Overlay{Unm|M}ountError to the parent?

```suggestion

    def __init__(self, action: str, mountpoint: str, message: str):
        self.action = action
        self.mountpoint = mountpoint
        self.message = message
        
        brief = f"Failed to {action} overly at {mountpoint}: {message}"
        
        super().__init__(brief=brief)
```

sergiusens: this may mean you don't need two distinct errors as .action would hold the operation (.action is not a final name, could be .operation :-))

craft_parts/overlays/overlay_fs.py
```diff
@@ -0,0 +1,118 @@
...
+            os_utils.mount(
+                "overlay",
+                str(mountpoint),
+                "-toverlay",
+                "-olowerdir={!s},upperdir={!s},workdir={!s}".format(
+                    lower_dir, self._upper_dir, self._work_dir
+                ),
+            )
+        except CalledProcessError as err:
+            raise errors.OverlayMountError(str(mountpoint), message=str(err))
```
sergiusens, line 61: It would be useful to have lower_dir and upper_dir if lower_dir and upper_dir can influence the error and are dynamically determined

craft_parts/overlays/errors.py
cmatsuoka: I have the feeling that this type of construct is very close to creating a generic exception only containing a message, given that something failing and the failure happening in overlay are implied if this is an overlay error. Since the general pattern is to have more specific exceptions, I would prefer to have separate mount/unmount errors in this case.

craft_parts/overlays/overlay_fs.py
cmatsuoka: The failed command (including upper, lower, work) is included in `CalledProcessError` so it will appear in the final error message.

cjp256, line 61: Consider chaining the exceptions, even though you're including the string.

```suggestion
            raise errors.OverlayMountError(str(mountpoint), message=str(err)) from error
```

cmatsuoka: Interesting, one of the linters used to suggest that but apparently not anymore?

sergiusens (approved):

cjp256 (approved):
Thanks!

PR #113, opened 2021-08-19
overlay: add overlay support to parts (CRAFT-470)

cmatsuoka:
Define the overlay work directory. Add part properties to retrieve
overlay-related subdirectories and identify if a part has overlay
parameters or overlay visibility.

-----

### DIFF
--- a/craft_parts/dirs.py
+++ b/craft_parts/dirs.py
@@ -25,6 +25,15 @@ class ProjectDirs:
 
     :param work_dir: The parent directory containing the parts, prime and stage
         subdirectories. If not specified, the current directory will be used.
+
+    :ivar work_dir: The root of the work directories used for project processing.
+    :ivar parts_dir: The directory containing work subdirectories for each part.
+    :ivar overlay_dir: The directory containing work subdirectories for overlays.
+    :ivar overlay_mount_dir: The mountpoint for the overlay filesystem.
+    :ivar overlay_packages_dir: The cache directory for overlay packages.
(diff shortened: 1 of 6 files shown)


cjp256 (approved):
LGTM

craft_parts/dirs.py
```diff
@@ -35,6 +35,26 @@ def work_dir(self) -> Path:
         """Return the root of the work directories used for project processing."""
         return self._work_dir
 
+    @property
+    def overlay_dir(self) -> Path:
+        """Return the directory containing work subdirectories for overlays."""
+        return self._work_dir / "overlay"
```
facundobatista, line 41: Do these need to be properties?

Why not keep them simple, and just, in `__init__`, do `self.overlay_dir = self._work_dir / "overlay"` ?

facundobatista (changes requested):
General idea looks great, but I'd go for a more simple implementation if possible.

craft_parts/dirs.py
cmatsuoka: `ProjectDirs` is minimally exposed in the codebase so we may want to change all its properties to attributes.

cmatsuoka: Done.

cmatsuoka:
`Final` typing hint not available in python 3.7, so not using it for now.

sergiusens (approved):

facundobatista (approved):
Great, thanks!

PR #114, opened 2021-08-20
states: refactor step state class, add migration state (CRAFT-471)

cmatsuoka:
Refactor step state and add a migration state superclass to be used
when migrating overlay files.

-----

### DIFF
--- a/craft_parts/state_manager/states.py
+++ b/craft_parts/state_manager/states.py
@@ -45,7 +45,7 @@ def load_step_state(part: Part, step: Step) -> Optional[StepState]:
 
     :raise RuntimeError: If step is invalid.
     """
-    filename = step_state_path(part, step)
+    filename = get_step_state_path(part, step)
     if not filename.is_file():
         return None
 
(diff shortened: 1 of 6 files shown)


craft_parts/state_manager/states.py
```diff
@@ -80,6 +99,17 @@ def remove(part: Part, step: Step) -> None:
         state_file.unlink()
 
 
-def state_file_path(part: Part, step: Step) -> Path:
-    """Return the path to the state file for the give part and step."""
+def step_state_path(part: Part, step: Step) -> Path:
```
cjp256, line 102: jw if we should prefix these methods with get_ or similar, but your call :D

cjp256 (approved):
LGTM

craft_parts/state_manager/states.py
sergiusens: that would be nice even if the operation is cheap

cmatsuoka: Renamed `step_state_path()` and `overlay_migration_state_path()` to `get_step_state_path()` and `get_overlay_migration_state_path()`.

facundobatista (approved):
Looks great.

PR #119, opened 2021-08-30
overlay: add overlay hash to build and stage steps (CRAFT-489)

cmatsuoka:
List the overlay verification hash in the build and stage steps to
ensure proper invalidation of steps depending on overlay data. This
hash will be verified in the build step of parts that can see overlay
contents, and in the stage step of parts declaring overlay parameters.

-----

### DIFF
--- a/craft_parts/executor/part_handler.py
+++ b/craft_parts/executor/part_handler.py
@@ -27,6 +27,7 @@ from typing import Any, Callable, Dict, List, Optional
 from craft_parts import callbacks, errors, packages, plugins, sources
 from craft_parts.actions import Action, ActionType
 from craft_parts.infos import PartInfo, StepInfo
+from craft_parts.overlays import LayerHash
 from craft_parts.packages import errors as packages_errors
 from craft_parts.parts import Part
 from craft_parts.plugins import Plugin
@@ -57,10 +58,12 @@ class PartHandler:
         part_info: PartInfo,
         part_list: List[Part],
         ignore_patterns: Optional[List[str]] = None,
+        base_layer_hash: Optional[LayerHash] = None,
     ):
         self._part = part
         self._part_info = part_info
         self._part_list = part_list
+        self._base_layer_hash = base_layer_hash
 
         self._plugin = plugins.get_plugin(
             part=part,
@@ -180,10 +183,13 @@ class PartHandler:
         }
         assets.update(_get_machine_manifest())
 
+        overlay_hash = self._compute_layer_hash(for_all_parts=True)
+
         state = states.BuildState(
             part_properties=self._part_properties,
             project_options=step_info.project_options,
             assets=assets,
+            overlay_hash=overlay_hash.hex(),
         )
         return state
 
@@ -196,11 +202,14 @@ class PartHandler:
             work_dir=self._part.stage_dir,
         )
 
+        overlay_hash = self._compute_layer_hash(for_all_parts=True)
+
         state = states.StageState(
             part_properties=self._part_properties,
             project_options=step_info.project_options,
             files=contents.files,
             directories=contents.dirs,
+            overlay_hash=overlay_hash.hex(),
         )
         return state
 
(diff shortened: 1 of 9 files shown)


craft_parts/executor/part_handler.py
```diff
@@ -180,10 +183,13 @@ def _run_build(self, step_info: StepInfo, *, update=False) -> StepState:
         }
         assets.update(_get_machine_manifest())
 
+        overlay_hash = self._compute_layer_hash(for_all_parts=True)
```
sergiusens, line 186: can you add a code comment on why for_all_parts and what the objective of this piece of code is? Not so much the what, but why.

craft_parts/executor/part_handler.py
```diff
@@ -196,11 +202,14 @@ def _run_stage(self, step_info: StepInfo) -> StepState:
             work_dir=self._part.stage_dir,
         )
 
+        overlay_hash = self._compute_layer_hash(for_all_parts=True)
```
sergiusens, line 205: Similar to above, the reason is probably different

craft_parts/executor/part_handler.py
```diff
@@ -248,6 +257,19 @@ def _run_step(
 
         return step_handler.run_builtin()
 
+    def _compute_layer_hash(self, *, for_all_parts: bool = False) -> LayerHash:
```
sergiusens, line 260: you seem to always set this to True, is there a reason for this parameter?

craft_parts/overlays/layers.py
```diff
@@ -49,7 +51,8 @@ def for_part(cls, part: Part, *, previous_layer_hash: "LayerHash") -> "LayerHash
             to the given part.
         """
         hasher = hashlib.sha1()
-        hasher.update(previous_layer_hash.hash_bytes)
+        if previous_layer_hash:
+            hasher.update(previous_layer_hash.hash_bytes)
```
facundobatista, line 55: The `hash_bytes` attribute in `LayerHash` is super-misleading. 

It looks like you're updating here with the **bytes** of the previous hash layer, but you're only using the **digest** of the previous hash layer. 

May this naming be improved? thanks!

craft_parts/state_manager/build_state.py
```diff
@@ -25,6 +27,14 @@ class BuildState(StepState):
     """Context information for the build step."""
 
     assets: Dict[str, Any] = {}
+    overlay_hash: Optional[str] = None
+
+    @validator("overlay_hash")
+    @classmethod
+    def _validate_hex_string(cls, value):
```
facundobatista, line 34: please add a docstring here (and type hints, if you want); thanks!

craft_parts/state_manager/stage_state.py
```diff
@@ -16,14 +16,25 @@
...
 
 
 class StageState(StepState):
     """Context information for the stage step."""
 
+    overlay_hash: Optional[str] = None
+
+    @validator("overlay_hash")
+    @classmethod
+    def _validate_hex_string(cls, value):
```
facundobatista, line 33: there's one exactly like this one defined also above...

facundobatista (changes requested):
Added some questions and comments, thanks!!

craft_parts/executor/part_handler.py
cmatsuoka: The overlay step handler uses the validation hash for layers up to the step being processed (but this handler is not implemented yet).

craft_parts/executor/part_handler.py
cmatsuoka: Added a comment explaining the reason.

craft_parts/executor/part_handler.py
cmatsuoka: Also added an explanatory comment. The reason is similar except that in build we verify the overlay hash to invalidate parts with overlay visibility (so these parts are rebuilt) and in stage we verify it to invalidate parts declaring overlay parameters (so parts are restaged).

craft_parts/executor/part_handler.py
cmatsuoka: Added a docstring for `_compute_layer_hash()`.

craft_parts/state_manager/build_state.py
cmatsuoka: Added docstring.

craft_parts/state_manager/stage_state.py
cmatsuoka: Reusing the validator.

craft_parts/overlays/layers.py
cmatsuoka: The name refers to the hash of the previous layer, expressed in bytes (as opposed to hexadecimal string). Since it can potentially confuse readers, I'm renaming to `digest`.

sergiusens (approved):
My only remaining comment would be to remove the default for `for_all_parts` and maybe even remove the `for_`

facundobatista (approved):
great, thanks!!!

PR #122, opened 2021-09-06
overlay: add layer state management helper (CRAFT-506)

cmatsuoka:
The sequencer must keep track of layer state and changes in order to
emit actions related to the overlay step (such as layer reapplications
or re-execution of steps invalidated by changes in layers). Add a
layer state management helper to be used during the planning phase
to store the layer hash corresponding to each part and compute the
expected values to ensure overlay consistency.

-----

### DIFF
--- a/craft_parts/overlays/layers.py
+++ b/craft_parts/overlays/layers.py
@@ -114,7 +114,7 @@ class LayerStateManager:
         self._part_list = part_list
         self._base_layer_hash = base_layer_hash
 
-        self._layer_hash: Dict[str, Optional[LayerHash]] = dict()
+        self._layer_hash: Dict[str, Optional[LayerHash]] = {}
         for part in part_list:
             self.set_layer_hash(part, LayerHash.load(part))
 
--- a/craft_parts/overlays/layers.py
+++ b/craft_parts/overlays/layers.py
@@ -107,6 +107,7 @@ class LayerStateManager:
     """An in-memory layer state management helper for action planning.
 
     :param part_list: The list of parts in the project.
+    :param base_layer_hash: The verification hash of the overlay base layer.
     """
 
     def __init__(self, part_list: List[Part], base_layer_hash: Optional[LayerHash]):
--- a/craft_parts/overlays/layers.py
+++ b/craft_parts/overlays/layers.py
@@ -18,7 +18,7 @@
 
 import hashlib
 import logging
-from typing import Optional
+from typing import Dict, List, Optional
 
 from craft_parts.parts import Part
 
@@ -35,6 +35,9 @@ class LayerHash:
         return self.hex()
 
     def __eq__(self, other):
+        if not isinstance(other, LayerHash):
+            return False
+
         return self.digest == other.digest
 
     @classmethod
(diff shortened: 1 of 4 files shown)


craft_parts/overlays/layers.py
```diff
@@ -98,3 +101,51 @@ def save(self, part: Part) -> None:
     def hex(self) -> str:
         """Return the current hash value as a hexadecimal string."""
         return self.digest.hex()
+
+
+class LayerStateManager:
+    """An in-memory layer state management helper for action planning.
+
+    :param part_list: The list of parts in the project.
```
sergiusens, line 109: missing base_layer_hash

tests/unit/overlays/test_layers.py
```diff
@@ -95,3 +95,111 @@ def test_save(self, new_dir):
...
+class TestLayerStateManager:
+    """Verify in-memory layer state management operations."""
+
+    def test_get_layer_hash(self):
+        p1 = Part("p1", {})
+        p2 = Part("p2", {})
+        base_layer_hash = LayerHash(b"base hash value")
+
+        Path("parts/p1/state").mkdir(parents=True)
+        lh = LayerHash(bytes.fromhex("a42a1d8ac7fdcfc4752e28aba0b0ee905e7cf96f"))
```
sergiusens, line 110: bytes are cheap, can we call this var layer_hash and similar for others?

craft_parts/overlays/layers.py
```diff
@@ -98,3 +101,51 @@ def save(self, part: Part) -> None:
...
+
+    def __init__(self, part_list: List[Part], base_layer_hash: Optional[LayerHash]):
+        self._part_list = part_list
+        self._base_layer_hash = base_layer_hash
+
+        self._layer_hash: Dict[str, Optional[LayerHash]] = dict()
+        for part in part_list:
+            self.set_layer_hash(part, LayerHash.load(part))
+
+    def get_layer_hash(self, part: Part) -> Optional[LayerHash]:
```
sergiusens, line 120: why not require part_name instead

craft_parts/overlays/layers.py
```diff
@@ -98,3 +101,51 @@ def save(self, part: Part) -> None:
     def hex(self) -> str:
         """Return the current hash value as a hexadecimal string."""
         return self.digest.hex()
+
+
+class LayerStateManager:
```
sergiusens, line 106: maybe convert this into https://docs.python.org/3/library/collections.html#userdict-objects ?

And instead of passign parts, pass part_names, the index in compute_layer_hash can be derived from the part name.

craft_parts/overlays/layers.py
```diff
@@ -98,3 +101,51 @@ def save(self, part: Part) -> None:
...
+    """An in-memory layer state management helper for action planning.
+
+    :param part_list: The list of parts in the project.
+    """
+
+    def __init__(self, part_list: List[Part], base_layer_hash: Optional[LayerHash]):
+        self._part_list = part_list
+        self._base_layer_hash = base_layer_hash
+
+        self._layer_hash: Dict[str, Optional[LayerHash]] = dict()
```
facundobatista, line 116: Create the empty dict with `{}`.

craft_parts/sequencer.py
```diff
@@ -190,30 +193,26 @@ def _run_step(
...
                 project_options=self._project_info.project_options,
-                assets={},  # TODO: obtain pull assets
             )
 
         elif step == Step.BUILD:
             state = states.BuildState(
                 part_properties=part_properties,
                 project_options=self._project_info.project_options,
-                assets={},  # TODO: obtain build assets
+                overlay_hash=self._layer_state.get_overlay_hash().hex(),
```
facundobatista, line 202: What happened with `assets`?

craft_parts/sequencer.py
```diff
@@ -190,30 +193,26 @@ def _run_step(
...
             )
 
         elif step == Step.STAGE:
             state = states.StageState(
                 part_properties=part_properties,
                 project_options=self._project_info.project_options,
-                files=set(),
-                directories=set(),
+                overlay_hash=self._layer_state.get_overlay_hash().hex(),
```
facundobatista, line 209: `overlay_hash` is not specified in the States; should it be?

facundobatista (changes requested):
Nice! Added some questions inside...

craft_parts/overlays/layers.py
cmatsuoka: Added documentation for `base_layer_hash`.

craft_parts/overlays/layers.py
cmatsuoka: Indeed, this could be the part name. Changing to use it instead.

craft_parts/overlays/layers.py
cmatsuoka: I was passing the part directory to `compute_layer_hash` to prevent some inefficiency in taking the name from a part and convert it back to a part. We can rebase it on UserDict, but I'd prefer to do it in a refactoring round after overlays are implemented in sequencer and executor so we can evaluate the interaction of all components in context.

craft_parts/sequencer.py
cmatsuoka: This was originally based on Snapcraft states and craft-parts doesn't use them for lifecycle planning. The need for asset listing will be re-evaluated during integration with next-gen snapcraft, when we can reinstate it or drop it if not needed

craft_parts/sequencer.py
cmatsuoka: `overlay_hash` is specific to the Build and Stage states, which are the steps affected by changes in the overlay verification hash.

craft_parts/overlays/layers.py
cmatsuoka: It seems that `dict()` could lead to gratuitous inefficiency, so replacing it with `{}`.

tests/unit/overlays/test_layers.py
cmatsuoka: Sure, renamed to `layer_hash` (and `p1_layer_hash`, etc).

craft_parts/sequencer.py
facundobatista: But it's not specified in `BuildState` either, in `craft_parts/state_manager/build_state.py`....

craft_parts/overlays/layers.py
sergiusens: I really dislike this reality as `dict()` is much easier to quickly gloss over than `{}`.

craft_parts/sequencer.py
cmatsuoka: It is there:
```
class BuildState(StepState):
    """Context information for the build step."""

    assets: Dict[str, Any] = {}
    overlay_hash: Optional[str] = None
```
This was added in 96274b8

sergiusens (approved):

craft_parts/sequencer.py
facundobatista: Sorry, I didn't have the local copy updated :/

facundobatista (approved):
Thanks!

PR #125, opened 2021-09-10
main: add details to the parts specification error message (CRAFT-512)

cmatsuoka:
The parts specification error when running craft-parts from the CLI
tool entry point lacks information, add details on what cause the
reported error.

With this fix, parts errors will look like:
```
$  python -mcraft_parts
Error: invalid parts specification: Part 'p1' validation failed.                                                                                                                                
'source' is required by the dump plugin                                                         
Review part 'p1' and make sure it's correct.
```

-----

### DIFF
--- a/craft_parts/main.py
+++ b/craft_parts/main.py
@@ -58,8 +58,8 @@ def main():
             msg = f"{err.filename}: {msg}"
         print(f"Error: {msg}.", file=sys.stderr)
         sys.exit(1)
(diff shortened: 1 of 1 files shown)


sergiusens (approved):

facundobatista (approved):
+1, but please normalize having the dot at the end (or not) in the different `print(f"Error".....` calls.

PR #127, opened 2021-09-10
overlay: add overlay manager mount operations (CRAFT-80)

cmatsuoka:
The overlay manager is used to mount the overlay stack during the execution
of the overlay step (or the build step for parts with overlay visibility) and perform
overlay package installation. This commit implements mount operations, package
installation will be added in a separate changeset.

-----


craft_parts/overlays/overlay_manager.py
```diff
@@ -0,0 +1,114 @@
...
+        self._base_layer_dir = base_layer_dir
+
+    def mount_layer(self, part: Part, *, pkg_cache: bool = False) -> None:
+        """Mount the overlay step layer stack up to the given part.
+
+        :param part: The part corresponding to the topmost layer to mount.
+        :param pkg cache: Whether the package cache layer is enabled.
+        """
+        if not self._base_layer_dir:
+            return
```
facundobatista, line 61: Here (and similar below), what should happen if not _base_layer_dir? I don't like this function silently doing nothing in this case. 

Is this a valid real case? I see from the signature and tests that `_base_layer_dir` may be None. My point is if instantiating `OverlayManager` with `base_layer_dir=None` will make all these to not do anything, what's the point of instantiating it?

facundobatista (commented):
Added a question to understand better the usage. Thanks!

craft_parts/overlays/overlay_manager.py
cmatsuoka: This is a consistency check to make sure the base layer is defined when the overlayfs mount is requested (the requests happen during overlay package installation, overlay script execution, and building of parts with overlay visibility). The method should never be called without the base dir and I'll raise runtime errors instead if that happens.

craft_parts/overlays/overlay_manager.py
```diff
@@ -0,0 +1,114 @@
...
+        self,
+        *,
+        project_info: ProjectInfo,
+        part_list: List[Part],
+        base_layer_dir: Optional[Path]
+    ):
+        self._project_info = project_info
+        self._part_list = part_list
+        self._layer_dirs = [p.part_layer_dir for p in part_list]
+        self._overlay_mount_dir = project_info.overlay_mount_dir
```
sergiusens, line 50: unused attribute

craft_parts/overlays/overlay_manager.py
```diff
@@ -0,0 +1,114 @@
...
+            upper_dir=self._project_info.overlay_packages_dir,
+            work_dir=self._project_info.overlay_work_dir,
+        )
+
+        self._overlay_fs.mount(self._project_info.overlay_mount_dir)
+
+    def unmount(self) -> None:
+        """Unmount the overlay step layer stack."""
+        if not self._overlay_fs:
+            logger.warning("overlay filesystem not mounted")
```
sergiusens, line 101: why is this just a warning?

craft_parts/overlays/overlay_manager.py
```diff
@@ -0,0 +1,114 @@
...
+    :param base_layer_dir: The directory containing the overlay base, or None
+        if the project doesn't use overlay parameters.
+    """
+
+    def __init__(
+        self,
+        *,
+        project_info: ProjectInfo,
+        part_list: List[Part],
+        base_layer_dir: Optional[Path]
```
sergiusens, line 45: why is this optional?

craft_parts/overlays/overlay_manager.py
cmatsuoka: Removed.

craft_parts/overlays/overlay_manager.py
cmatsuoka: The overlay manager is currently instantiated on the `Executor` class and passed to each part handler, even if the base layer directory is not specified by the user. This can probably be refactored to make the entire overlay manager instance to be optional in this case, but let's do that in a final refactoring round after all the overlay subsystem pieces are in place so we can better evaluate it in context.

sergiusens (approved):

PR #129, opened 2021-09-15
sources: fix removed file updates (CRAFT-532)

cmatsuoka:
Source handlers are used to pull sources from the location specified
in the part definition into the part source work directory. They may
provide a method to update specific files that have been modified
without re-pulling the entire source tree.

Changes were introduced in 5d56713 to allow removal of pulled files
that have been deleted in the part source.

The local source handler, however, is not used only for the primary
source handler purpose, it's also used to migrate contents from the
source work directory to the build work directory, regardless of the
type of the part's source. (We should investigate how to refactor this
to use the migration mechanisms directly without reusing the source
handler.)

As a consequence of the way deleted file removal was implemented and
the reuse of the local source handler for build migration, artifacts
created during the build process were removed in a source update.
Deleted file updates should cover only files removed from the original
source without affecting build artifacts.

Refactor source handlers to preserve build artifacts by removing
only entries present in the source work directory that are not
in the original source from the build work directory. In the long
term part handling should be refactored to be decoupled from the
local source handler.

-----

### DIFF
--- a/craft_parts/sources/local_source.py
+++ b/craft_parts/sources/local_source.py
@@ -157,13 +157,24 @@ class LocalSource(SourceHandler):
                 os.path.join(self.part_src_dir, file_path),
             )
 
-        # And remove deleted entries
+    def source_cleanup(self):
+        """Remove pulled entries not present in the actual part source."""
         for entry in self._deleted_entries:
             path = Path(self.part_src_dir, entry)
-            if path.is_dir():
-                shutil.rmtree(str(path))
-            else:
-                path.unlink()
+            _delete_entry(path)
+
+            if self.part_build_dir:
+                path = self.part_build_dir / entry
+                _delete_entry(path)
+
+
+def _delete_entry(path: Path) -> None:
+    if path.is_dir():
+        logger.debug("delete dir: %s", path)
+        shutil.rmtree(str(path))
+    else:
+        logger.debug("delete file: %s", path)
+        path.unlink()
 
 
 def _check_deleted_sources(srcdir: Path, destdir: Path) -> List[Path]:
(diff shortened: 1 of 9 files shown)


cmatsuoka:
This is the wrong solution for the problem. It solves the build artifacts issue, but new files can be created in `part_src_dir` as a result of `override-pull`. For deleted file updates to work correctly, we'll probably need to keep a manifest of pulled files including anything created by the user. We also should revert 5d56713, and if the user deletes local source files they need to repull.

craft_parts/sources/local_source.py
```diff
@@ -157,13 +157,24 @@ def update(self):
...
+            _delete_entry(path)
+
+            if self.part_build_dir:
+                path = self.part_build_dir / entry
+                _delete_entry(path)
+
+
+def _delete_entry(path: Path) -> None:
+    if path.is_dir():
+        logger.debug("delete dir: %s", path)
```
facundobatista, line 173: Better %r, or as we don in charmcraft, %r of str(path), to not be confused by spaces but also don't get the ugly "Pathlib" thing

craft_parts/sources/local_source.py
```diff
@@ -157,13 +157,24 @@ def update(self):
...
-            else:
-                path.unlink()
+            _delete_entry(path)
+
+            if self.part_build_dir:
+                path = self.part_build_dir / entry
+                _delete_entry(path)
+
+
+def _delete_entry(path: Path) -> None:
```
facundobatista, line 171: Please add a docstring.

facundobatista (approved):
+1, but commented about two minor details.

PR #130, opened 2021-09-15
packages: fix apt_pkg imports (CRAFT-534)

cmatsuoka:
Pyright warns that `apt_pkg` is not exported from module `apt`. Use
the toplevel `apt_pkg` module directly instead.

-----

### DIFF
--- a/craft_parts/packages/apt_cache.py
+++ b/craft_parts/packages/apt_cache.py
@@ -31,6 +31,7 @@ import apt.cache
 import apt.package
 import apt.progress
 import apt.progress.base
+import apt_pkg
 
 from craft_parts.utils import os_utils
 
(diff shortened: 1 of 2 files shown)


facundobatista (commented):
Change is fine. But if we're using `apt_pkg` directly, shouldn't we also request it in the dependencies list?

I mean, if `apt` is not exporting it, maybe in the future it could stop including it...

cmatsuoka:
Python-apt is handled separately with a dependency specific to each OS release. Since it's a wrapper for libapt-pkg, I don't think this will change in the foreseeable future, but we'll keep an eye on any changes. Also it seems that the non-exported module issue always existed, and only now pyright started to notice it.

sergiusens (approved):

facundobatista (approved):
Thanks for the explanation!

PR #132, opened 2021-09-16
sources: revert updating deleted files in outdated check (CRAFT-532)

cmatsuoka:
Commit 5d56713 is the wrong solution for the problem as it removes build
artifacts (because LocalSource is reused to migrate from `part_src_dir`
to `part_build_dir`) and any files created by the user in override-pull.

Updating deleted files can be revisited later. We should disable deleted
file synchronization during the migration to preserve build artifacts,
and be sure that the pull update handler is aware of new files created
as a part of the pull step.

For now, if the user deletes local source files they need to repull.

-----


cmatsuoka:
This replaces the partial solution proposed in PR #129.

facundobatista (approved):
+1

sergiusens (approved):

PR #134, opened 2021-09-21
overlay: add overlay package install (CRAFT-81)

cmatsuoka:
Overlay packages are installed on the layer corresponding to the
declaring part. This layer is configured as the upper directory of
the overlay stack. The installation runs on a chrooted environment
so that maintainer scripts are executed and the installed packages
database is updated correctly.
    
-----

### DIFF
--- a/tests/unit/overlays/test_overlay_manager.py
+++ b/tests/unit/overlays/test_overlay_manager.py
@@ -156,7 +156,7 @@ class TestPackageManagement:
     """Verify package installation on mounted overlayfs."""
 
     @pytest.fixture(autouse=True)
-    def setup_method_fixture(self, new_dir):
+    def setup_method_fixture(self, mocker, new_dir):
         # pylint: disable=attribute-defined-outside-init
         info = ProjectInfo(application_name="test", cache_dir=new_dir)
         self.p1 = Part("p1", {"plugin": "nil"})
@@ -168,32 +168,30 @@ class TestPackageManagement:
             part_list=[self.p1, self.p2],
             base_layer_dir=base_layer_dir,
         )
+        self.mock_mount = mocker.patch("craft_parts.utils.os_utils.mount")
+        self.mock_umount = mocker.patch("craft_parts.utils.os_utils.umount")
+        self.mock_chroot = mocker.patch("pychroot.Chroot")
+        self.mock_refresh_packages_list = mocker.patch(
+            "craft_parts.packages.Repository.refresh_packages_list"
+        )
         # pylint: enable=attribute-defined-outside-init
 
     def test_refresh_packages_list(self, mocker, new_dir):
-        mock_mount = mocker.patch("craft_parts.utils.os_utils.mount")
-        mock_chroot = mocker.patch("pychroot.Chroot")
-        mock_refresh_packages_list = mocker.patch(
-            "craft_parts.packages.Repository.refresh_packages_list"
-        )
-
         self.om.mkdirs()
         self.om.mount_pkg_cache()
         self.om.refresh_packages_list()
 
-        mock_mount.assert_called_once_with(
+        self.mock_mount.assert_called_once_with(
             "overlay",
             str(new_dir / "overlay/overlay"),
             "-toverlay",
             f"-olowerdir=base_dir,upperdir={new_dir}/overlay/packages,"
             f"workdir={new_dir}/overlay/work",
         )
-        mock_chroot.assert_called_once_with(new_dir / "overlay/overlay")
-        mock_refresh_packages_list.called_once_with()
+        self.mock_chroot.assert_called_once_with(new_dir / "overlay/overlay")
+        self.mock_refresh_packages_list.called_once_with()
 
     def test_download_packages(self, mocker, new_dir):
-        mock_mount = mocker.patch("craft_parts.utils.os_utils.mount")
-        mock_chroot = mocker.patch("pychroot.Chroot")
         mock_download_packages = mocker.patch(
             "craft_parts.packages.Repository.download_packages"
         )
@@ -202,19 +200,17 @@ class TestPackageManagement:
         self.om.mount_pkg_cache()
         self.om.download_packages(["pkg1", "pkg2"])
 
-        mock_mount.assert_called_once_with(
+        self.mock_mount.assert_called_once_with(
             "overlay",
             str(new_dir / "overlay/overlay"),
             "-toverlay",
             f"-olowerdir=base_dir,upperdir={new_dir}/overlay/packages,"
             f"workdir={new_dir}/overlay/work",
         )
-        mock_chroot.assert_called_once_with(new_dir / "overlay/overlay")
+        self.mock_chroot.assert_called_once_with(new_dir / "overlay/overlay")
         mock_download_packages.called_once_with(["pkg1", "pkg2"])
 
     def test_install_packages(self, mocker, new_dir):
-        mock_mount = mocker.patch("craft_parts.utils.os_utils.mount")
-        mock_chroot = mocker.patch("pychroot.Chroot")
         mock_install_packages = mocker.patch(
             "craft_parts.packages.Repository.install_packages"
         )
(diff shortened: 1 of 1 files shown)


craft_parts/lifecycle_manager.py
```diff
@@ -151,7 +151,7 @@ def refresh_packages_list(self) -> None:  # pylint: disable=no-self-use
         sequence of actions to take. To ensure consistency between the scenarios,
         it shouldn't be updated between planning and execution.
         """
-        packages.Repository.refresh_build_packages_list()
+        packages.Repository.refresh_packages_list()
```
cmatsuoka, line 154: Renamed because it's also used for overlay packages.

craft_parts/overlays/overlay_manager.py
```diff
@@ -110,3 +117,122 @@ def mkdirs(self) -> None:
...
+        if not self._overlay_fs:
+            raise RuntimeError("overlay filesystem not mounted")
+
+        mount_dir = self._project_info.overlay_mount_dir
+        with contextlib.suppress(SystemExit), pychroot.Chroot(mount_dir):
+            packages.Repository.install_packages(package_names)
+            shutil.rmtree("/var/cache", ignore_errors=True)
+
+    def fix_resolv_conf(self) -> None:
+        """Work around problems with pychroot when resolv.conf a symlink."""
```
facundobatista, line 156: "is a symlink"

craft_parts/overlays/overlay_manager.py
```diff
@@ -110,3 +117,122 @@ def mkdirs(self) -> None:
...
+        resolv = self._project_info.overlay_mount_dir / "etc" / "resolv.conf"
+        if resolv.is_symlink():
+            resolv.unlink()
+            resolv.touch()
```
facundobatista, line 160: When it could be a symlink? And a symlink to what? 

Maybe instead of removing it and leaving an empty file in place, would be better to copy the symlinked file?

craft_parts/overlays/overlay_manager.py
```diff
@@ -110,3 +117,122 @@ def mkdirs(self) -> None:
...
+            raise RuntimeError("overlay filesystem not mounted")
+
+        mount_dir = self._project_info.overlay_mount_dir
+        with contextlib.suppress(SystemExit), pychroot.Chroot(mount_dir):
+            packages.Repository.install_packages(package_names)
+            shutil.rmtree("/var/cache", ignore_errors=True)
+
+    def fix_resolv_conf(self) -> None:
+        """Work around problems with pychroot when resolv.conf a symlink."""
+        resolv = self._project_info.overlay_mount_dir / "etc" / "resolv.conf"
```
facundobatista, line 157: Here you're not checking if the overlay fs was mounted...

setup.py
```diff
@@ -34,7 +34,7 @@ def is_ubuntu() -> bool:
 
 install_requires = [
     "PyYAML",
-    "progressbar",
+    "pychroot==0.10.3",  # 0.10.4 drops Python 3.7 support
```
facundobatista, line 37: What happened to `progressbar`?

facundobatista (changes requested):
Annotated some details inside, but there is a mix in the design that I don't end up feeling comfortable.

From outside, I will do something like the following...

```
>>> om = OverlayManager
>>> with LayerMount(om, ...) as layer_mount:
       ...
```

An `LayerMount` will call `OverlayManager.mount_layer`, do the context manager's code, and then **exit the process** if it's not the original one, and `unmount`.

I guess that the context manager's code will do `layer_mount.install_packages`, which will execute `OverlayManager.install_packages`, that will create a "chroot" (running other process) to install the packages, in the context manager of lines 151 to 153. 

When that context manager ends it will suppress a SystemExit, `OverlayManager.install_packages` will finish, `LayerMount.install_packages` will finish, and then the outer context manager will exit, running `__exit__` at line 191, which is the one that will raise SystemExit, outside of the context manager that is suppressing it??

I don't know if I expressed well my confusion here, let's talk if you want. Thanks!!

craft_parts/overlays/overlay_manager.py
cmatsuoka: Some images (such as cloudimgs) have this symlink in place:
```
$ ls -l etc/resolv.conf
lrwxrwxrwx 1 root root 39 Mar 11  2021 etc/resolv.conf -> ../run/systemd/resolve/stub-resolv.conf
```
Pychroot needs `resolv.conf` to be a regular file, otherwise it fails to install its own resolver configuration and name resolution will fail inside the chroot.

cmatsuoka:
In the process of wrapping the pychroot context manager we learned that it does too much magic (inspired by [withhacks](https://pypi.org/project/withhacks/)) and that's not something we really want in this project. Writing a replacement with all the functionality we need can take some time so we'll keep pychroot for now, but we should replace it with something cleaner and less hacky when time permits.

setup.py
cmatsuoka: Progress display was dropped a few releases ago but the package dependency was not removed.

craft_parts/overlays/overlay_manager.py
cmatsuoka: This method is only called by the `*Mount` context managers right after the filesystem is mounted. It is another workaround for bad pychroot behavior, and should go away when we replace pychroot.

craft_parts/overlays/overlay_manager.py
facundobatista: What about copying it instead of leave an empty file?

facundobatista (approved):
Approving, but please consider copying the original file (or add a comment explaining why it doesn't matter and it's safe to have an empty one). Thanks!

craft_parts/overlays/overlay_manager.py
```diff
@@ -110,3 +117,125 @@ def mkdirs(self) -> None:
             self._project_info.overlay_work_dir,
         ]:
             overlay_dir.mkdir(parents=True, exist_ok=True)
+
+    def refresh_packages_list(self) -> None:
+        """Update the list of available packages in the overlay system."""
+        if not self._overlay_fs:
+            raise RuntimeError("overlay filesystem not mounted")
```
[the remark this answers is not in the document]
cmatsuoka: This test should never fail if the code is correct. We can revise/adjust this after the entire subsystem is in place.

tests/unit/overlays/test_overlay_manager.py
```diff
@@ -150,3 +150,184 @@ def test_mkdirs(self, new_dir):
...
+            base_layer_dir=base_layer_dir,
+        )
+        # pylint: enable=attribute-defined-outside-init
+
+    def test_refresh_packages_list(self, mocker, new_dir):
+        mock_mount = mocker.patch("craft_parts.utils.os_utils.mount")
+        mock_chroot = mocker.patch("pychroot.Chroot")
+        mock_refresh_packages_list = mocker.patch(
+            "craft_parts.packages.Repository.refresh_packages_list"
+        )
```
sergiusens, line 178: can't these all be patched in the setup_method_fixture? If refresh is patched when install is to be used shouldn't matter much and at least you don't accidentally trigger the wrong thing and break your env :-)

sergiusens (approved):

tests/unit/overlays/test_overlay_manager.py
cmatsuoka: Moved common overlay manager test mocks to test class setup.

cmatsuoka:
> Approving, but please consider copying the original file (or add a comment explaining why it doesn't matter and it's safe to have an empty one). Thanks!

Added additional clarification. In short, it's safe to use an empty file because the layer it's created on is discarded after the overlay step, and it's not included in the migrated payload.

craft_parts/overlays/overlay_manager.py
cmatsuoka: Pychroot only needs a file to bind-mount it's own file on. Our replacement is done on the package cache overlayfs layer, and this layer is discarded after the overlay step execution. The empty file is not migrated or included in the final payload.

Added additional clarification to the source code.

PR #137, opened 2021-09-29
many: add overlay step (CRAFT-84)

cmatsuoka:
The overlay step is used to process filesystem mutations defined for
each part. It can include package installation, modification of existing
file data or metadata, file renaming or removal, etc.

**Note to reviewers:** this change is rather large because many subsystems
need to be updated to handle the new step, and tests need to be updated to
account for the actions corresponding to step changes. Additional integration
tests will be added after the outdated overlay verification tests land, covering
the usage scenarios described in the feature specification.
    
### DIFF
--- a/craft_parts/lifecycle_manager.py
+++ b/craft_parts/lifecycle_manager.py
@@ -124,9 +124,9 @@ class LifecycleManager:
         # a base layer is mandatory if overlays are in use
         if self._has_overlay:
             if not base_layer_dir:
-                raise ValueError("using overlays and base_layer_dir not specified")
+                raise ValueError("base_layer_dir must be specified if using overlays")
             if not base_layer_hash:
-                raise ValueError("using overlays and base_layer hash not specified")
+                raise ValueError("base_layer_hash must be specified if using overlays")
         else:
             base_layer_dir = None
 
--- a/craft_parts/lifecycle_manager.py
+++ b/craft_parts/lifecycle_manager.py
@@ -26,6 +26,7 @@ from craft_parts import errors, executor, packages, plugins, sequencer
 from craft_parts.actions import Action
 from craft_parts.dirs import ProjectDirs
 from craft_parts.infos import ProjectInfo
+from craft_parts.overlays import LayerHash
 from craft_parts.parts import Part
 from craft_parts.steps import Step
 
@@ -59,6 +60,10 @@ class LifecycleManager:
         by the package manager used by the platform. Defaults to the application name.
     :param ignore_local_sources: A list of local source patterns to ignore.
     :param extra_build_packages: A list of additional build packages to install.
+    :param base_layer_dir: The path to the overlay base layer, if using overlays.
+    :param base_layer_hash: The validation hash of the overlay base image, if using
+        overlays. The validation hash should be constant for a given image, and should
+        change if a different base image is used.
     :param custom_args: Any additional arguments that will be passed directly
         to :ref:`callbacks<callbacks>`.
     """
@@ -76,8 +81,12 @@ class LifecycleManager:
         application_package_name: Optional[str] = None,
         ignore_local_sources: Optional[List[str]] = None,
         extra_build_packages: Optional[List[str]] = None,
+        base_layer_dir: Optional[Path] = None,
+        base_layer_hash: Optional[bytes] = None,
         **custom_args,  # custom passthrough args
     ):
+        # pylint: disable=too-many-locals
+
         if not re.match("^[A-Za-z][0-9A-Za-z_]*$", application_name):
             raise errors.InvalidApplicationName(application_name)
 
@@ -110,6 +119,22 @@ class LifecycleManager:
         for name, spec in parts_data.items():
             part_list.append(_build_part(name, spec, project_dirs))
 
+        self._has_overlay = any(p.has_overlay for p in part_list)
+
+        # a base layer is mandatory if overlays are in use
+        if self._has_overlay:
+            if not base_layer_dir:
+                raise ValueError("using overlays and base_layer_dir not specified")
+            if not base_layer_hash:
+                raise ValueError("using overlays and base_layer hash not specified")
+        else:
+            base_layer_dir = None
+
+        if base_layer_hash:
+            layer_hash: Optional[LayerHash] = LayerHash(base_layer_hash)
+        else:
+            layer_hash = None
+
         self._part_list = part_list
         self._application_name = application_name
         self._target_arch = project_info.target_arch
@@ -117,14 +142,18 @@ class LifecycleManager:
             part_list=self._part_list,
             project_info=project_info,
             ignore_outdated=ignore_local_sources,
+            base_layer_hash=layer_hash,
         )
         self._executor = executor.Executor(
             part_list=self._part_list,
             project_info=project_info,
             ignore_patterns=ignore_local_sources,
             extra_build_packages=extra_build_packages,
+            base_layer_dir=base_layer_dir,
+            base_layer_hash=layer_hash,
         )
         self._project_info = project_info
+        # pylint: enable=too-many-locals
 
     @property
     def project_info(self) -> ProjectInfo:
--- a/craft_parts/main.py
+++ b/craft_parts/main.py
@@ -80,12 +80,19 @@ def _process_parts(options: argparse.Namespace) -> None:
     if not cache_dir:
         cache_dir = BaseDirectory.save_cache_path("craft-parts")
 
+    if options.overlay_base:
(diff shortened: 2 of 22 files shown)


craft_parts/sequencer.py
```diff
@@ -181,6 +191,29 @@ def _run_step(
...
 
+        if step == Step.OVERLAY:
+            # Make sure all previous layers are in place before we add a new
+            # layer to the overlay stack,
+            layer_hash = self._ensure_overlay_consistency(
+                part,
+                reason=f"required to overlay {part.name!r}",
+                skip_last=True,
+            )
+            self._layer_state.set_layer_hash(part, layer_hash)
```
cmatsuoka, line 202: Here we obtain the verification hash for the layer being processed.

craft_parts/sequencer.py
```diff
@@ -181,6 +191,29 @@ def _run_step(
...
+        ):
+            # The overlay step for all parts should run before we build a part
+            # with overlay visibility or before we stage a part that declares
+            # overlay parameters.
+            last_part = self._part_list[-1]
+            verb = _step_verb[step]
+            self._ensure_overlay_consistency(
+                last_part,
+                reason=f"required to {verb} {part.name!r}",
+            )
```
cmatsuoka, line 215: Later, in the build or stage steps, we verify if the layer stack is consistent based on the verification hash.

tests/integration/executor/test_callbacks.py
```diff
@@ -41,6 +41,12 @@ def teardown_module():
     callbacks.unregister_all()
 
 
+@pytest.fixture(autouse=True)
+def mock_mount_unmount(mocker):
+    mocker.patch("craft_parts.utils.os_utils.mount")
+    mocker.patch("craft_parts.utils.os_utils.umount")
```
cmatsuoka, line 47: Overlay operations require overlayfs mounting and umounting. We mock these in the regular unit and integration tests, but a new set of "privileged" tests will be added after the implementation is complete to actually mount overlay, perform chroot, etc.

craft_parts/main.py
```diff
@@ -80,12 +80,19 @@ def _process_parts(options: argparse.Namespace) -> None:
     if not cache_dir:
         cache_dir = BaseDirectory.save_cache_path("craft-parts")
 
+    if options.overlay_base:
+        base_layer_hash = options.overlay_base.encode()
```
cmatsuoka, line 84: The base layer hash algorithm is not specified and could be anything that remains constant for a given base. The CLI tool just uses the path to the base for simplicity, but real applications can (and probably should) use an actual digest.

cmatsuoka: Added this note as a comment in the code.

craft_parts/lifecycle_manager.py
```diff
@@ -110,21 +119,41 @@ def __init__(
...
             part_list.append(_build_part(name, spec, project_dirs))
 
+        self._has_overlay = any(p.has_overlay for p in part_list)
+
+        # a base layer is mandatory if overlays are in use
+        if self._has_overlay:
+            if not base_layer_dir:
+                raise ValueError("using overlays and base_layer_dir not specified")
+            if not base_layer_hash:
+                raise ValueError("using overlays and base_layer hash not specified")
```
facundobatista, line 129: The error message looks weird starting with "using overlays" (because that's not an error).

I propose to turn both messages around, in the line of "The base_layer_dir must be specified when using overlays.".

What do you think?

craft_parts/lifecycle_manager.py
```diff
@@ -110,21 +119,41 @@ def __init__(
...
+        else:
+            base_layer_dir = None
+
+        if base_layer_hash:
+            layer_hash: Optional[LayerHash] = LayerHash(base_layer_hash)
+        else:
+            layer_hash = None
```
facundobatista, line 136: Having base_layer_dir and/or layer_hash is ok if NOT using overlays?

Also, if one of base_layer_dir or layer_hash is present, the other must be present too?

craft_parts/sequencer.py
```diff
@@ -247,9 +290,56 @@ def _add_action(
...
+                    part_names=[part.name],
+                    reason=reason,
+                )
+                self._layer_state.set_layer_hash(part, layer_hash)
+
+            if part.name == top_part.name:
+                return layer_hash
+
+        # execution should never reach this line
+        raise RuntimeError(f"part {top_part!r} not in parts list")
```
facundobatista, line 337: I agree that code should never get here, but the reason is wrong (as it's checked that in line 314/315).

Then I started to dislike having two returns in the for loop.

So I tried to think how to reorder code so we don't have a "should not get here" situation, and came up with the following (untested, and just a suggestion):

```
        useful_part_list = []
        for part in self._part_list:
            useful_part_list.append(part)
            if top_part.name == part.name:
                if skip_last:
                    del useful_part_list[-1]
                break
        else:
            raise ValueError(f"part {top_part!r} not in parts list")

        # NOTE!! here is more evident what happens when there's an empty "useful part list",
        # which should be managed (it can happen when "top part is the first part and skip_last")
        # In the original code the situation is there, but more difficult to see, and it's 
        # returning a layer_hash that should NOT be returning.

        for part in useful_part_list:
            layer_hash = self._layer_state.compute_layer_hash(part)
            state_layer_hash = self._layer_state.get_layer_hash(part)

            if layer_hash != state_layer_hash:
                self._add_all_actions(
                    target_step=Step.OVERLAY,
                    part_names=[part.name],
                    reason=reason,
                )
                self._layer_state.set_layer_hash(part, layer_hash)

        return layer_hash

```

What do you think?

facundobatista (changes requested):
Great work, annotated some details inside, thanks!

craft_parts/lifecycle_manager.py
cmatsuoka: Rephrased as suggested.

craft_parts/lifecycle_manager.py
cmatsuoka: If not using overlays, the base layer is not mounted and the overlay verification hash will not be computed, so the parameters are ignored.

craft_parts/sequencer.py
cmatsuoka: We want to return the hash of the top layer, even if it's the first layer and verification against state is skipped (note that the skip_last parameter only skips the consistency check but the hash is still computed). I fixed the docstring to make this intent clearer, and slightly refactored the logic to remove one of the return points.

If we pre-build the useful part list and it ends up empty (because the top part is the first part and we skip last) we're not returning the top part's hash.

If I manage to further refactor the logic improve clarity I'll propose a follow-up PR.

cmatsuoka: One more change: removed the initial test for part presence. It only fails in case of a programming error, and in this case we let the final guard to act (so it's not an "impossible" situation anymore).

craft_parts/lifecycle_manager.py
facundobatista: I see, so we're "safe". My question, though, was coming from the point of view of passing this information.

So, if I pass A and not B or C, instead of part getting ignored, shouldn't it raise an error?

craft_parts/sequencer.py
facundobatista: I see, thanks for all the clarifications.

One further detail, though: do we expect for all these hashes calculations and verifications to be super fast, or super slow?

Because if they will take a significative amount of time, maybe we should validate if the top_part is in the part list before doing everything else. Something similar to:

```
        useful_part_list = [] 
        for part in self._part_list: 
            useful_part_list.append(part) 
            if top_part.name == part.name: 
                break 
        else: 
            raise ValueError(f"part {top_part!r} not in parts list")

        for part in useful_part_list:
            layer_hash = self._layer_state.compute_layer_hash(part)

            # run the overlay step if the layer hash doesn't match the existing
            # state (unless we're in the top part and skipping the consistency check)
            if not (skip_last and part.name == top_part.name):
                state_layer_hash = self._layer_state.get_layer_hash(part)

                if layer_hash != state_layer_hash:
                    self._add_all_actions(
                        target_step=Step.OVERLAY,
                        part_names=[part.name],
                        reason=reason,
                    )
                    self._layer_state.set_layer_hash(part, layer_hash)
```

I just don't want to fall into those situations where the user makes a mistake, and the system takes a lot only to crash. It's an UX detail. But if it's fast, I'm happy with your code.

facundobatista (changes requested):

facundobatista (approved):
Actually approving this, as depending on the response to the questions, all is great.

cmatsuoka:
Thanks. The execution should be very fast, even if many parts with overlay are used.

PR #138, opened 2021-09-30
many: handle layer reapplication (CRAFT-83, CRAFT-85)

cmatsuoka:
If the overlay step is re-executed for a part, all layers sitting on top the of layer corresponding to that part need to be repopulated. This process is called "reapplication" of a layer. This is needed to keep the overlayfs stack consistent regarding modified files, including whiteouts and opaque directories.


craft_parts/executor/executor.py
```diff
@@ -63,7 +63,7 @@ def __init__(
         base_layer_dir: Optional[Path] = None,
         base_layer_hash: Optional[LayerHash] = None,
     ):
-        self._part_list = part_list
+        self._part_list = sort_parts(part_list)
```
cmatsuoka, line 66: Part ordering in executor was missing, caught in the updated set of tests.

craft_parts/executor/part_handler.py
```diff
@@ -475,6 +482,21 @@ def _update_build(self, step_info: StepInfo) -> None:
...
+            step_name = action.step.name.lower()
+            raise errors.InvalidAction(
+                f"cannot reapply step {step_name!r} of {self._part.name!r}"
+            )
+
+    def _reapply_overlay(self, step_info) -> None:
+        """Clean and repopulate the current part's layer, keeping its state."""
+        shutil.rmtree(self._part.part_layer_dir)
+        self._run_overlay(step_info)
+
```
cmatsuoka, line 499: This is the actual layer reapplication on disk: all previous contents are removed and the layer is completely re-populated.

craft_parts/sequencer.py
```diff
@@ -273,10 +276,21 @@ def _rerun_step(
...
 
+    def _reapply_layer(
+        self, part: Part, layer_hash: LayerHash, *, reason: Optional[str] = None
+    ):
+        """Update the layer hash without changing the step state."""
+        logger.debug("reapply layer %s: hash=%s", part.name, layer_hash)
+        self._layer_state.set_layer_hash(part, layer_hash)
+        self._add_action(
+            part, Step.OVERLAY, action_type=ActionType.REAPPLY, reason=reason
+        )
```
cmatsuoka, line 292: This is the layer reapplication during planning. It only changes state in memory and doesn't change the filesystem.

craft_parts/sequencer.py
```diff
@@ -331,6 +345,44 @@ def _ensure_overlay_consistency(
...
         raise RuntimeError(f"part {top_part!r} not in parts list")
 
+    def _check_if_dirty_on_overlay(self, part: Part, step: Step) -> bool:
+        """Verify whether the step is dirty because the overlay changed."""
+        if step == Step.OVERLAY:
+            # Layers depend on the integrity of its validation hash
+            current_layer_hash = self._layer_state.compute_layer_hash(part)
+            state_layer_hash = self._layer_state.get_layer_hash(part)
+            if current_layer_hash != state_layer_hash:
+                logger.debug("%s:%s changed layer hash", part.name, step)
```
facundobatista, line 355: Here and in the other two `logger.debug` below: does it make sense to replace `step` in it? I mean, you already know which step it is, wouldn't it be simpler to do `logger.debug("%s:OVERLAY changed layer hash", part.name)`?

Also, I searched this PR to see how this line would look like IRL, but it seems it's missing from the tests?

craft_parts/state_manager/state_manager.py
```diff
@@ -381,10 +388,39 @@ def check_if_dirty(self, part: Part, step: Step) -> Optional[DirtyReport]:
...
+            raise ValueError("only BUILD and STAGE states have overlay hash")
+
+        stw = self._state_db.get(part_name=part.name, step=step)
+        if not stw:
+            # step didn't run yet
+            return b""
+
+        state = stw.state
+
+        overlay_hash = state.overlay_hash  # type: ignore
```
facundobatista, line 418: silly detail: just colapse these two lines nito `overlay_hash = stw.state.overlay_hash`

facundobatista (approved):
+1, but please take a look at the comments and fix/add tests if needed; thanks!

craft_parts/state_manager/state_manager.py
cmatsuoka: Refactored as suggested.

craft_parts/sequencer.py
cmatsuoka: Added unit tests for this method. I kept the debug message so we can always have consistent step names even if we change the enumeration item name (impact should be negligible).

PR #139, opened 2021-10-01
overlays: replace pychroot with non-magic chroot (CRAFT-558)

cmatsuoka:
Add a simple, non-magic replacement for pychroot. Pychroot has numerous
issues such as being too hackish, being restricted to certain interpreter
versions and platforms, and not working in the CI environment. This
implementation is much simpler and we can actually understand it.
    
-----

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -103,3 +103,63 @@ jobs:
         run: |
           source ${HOME}/.venv/bin/activate
           make test-integrations
+      - name: Run overlay smoke test
+        run: |
+          source ${HOME}/.venv/bin/activate
+          wget -q https://cloud-images.ubuntu.com/focal/current/focal-server-cloudimg-amd64.squashfs
+          mkdir base
+          sudo mount -t squashfs focal-server-cloudimg-amd64.squashfs base/
+          cat <<-EOF > parts.yaml
+          parts:
+            foo:
+              plugin: nil
+              overlay-packages: [hello]
+              overlay-script: |
+                chroot "\$CRAFT_PARTS_OVERLAY" /bin/bash -c "hello > hello.txt"
+                echo 1234 > "\$CRAFT_PARTS_OVERLAY/etc/issue"
+              overlay:
+                - "*.txt"
+                - etc
+                - usr/bin
+            bar:
+              plugin: nil
+              after: [foo]
+              overlay-packages: [bison]
+              overlay-script: |
+                chroot "\$CRAFT_PARTS_OVERLAY" /bin/bash -c "rm -Rf /etc/shells /etc/ssh && mkdir /etc/ssh"
+                echo 5678 > "\$CRAFT_PARTS_OVERLAY/etc/issue"
+              overlay:
+                - -usr/share
+              override-build: |
+                rev "\$CRAFT_PARTS_OVERLAY/hello.txt" > "\$CRAFT_PARTS_PART_INSTALL/olleh.txt"
+          EOF
+          cat <<-EOF > run.sh
+            source ${HOME}/.venv/bin/activate
+            echo "--- overlay foo"
+            python -mcraft_parts --work-dir=work --trace --overlay-base=base --refresh overlay foo
+            echo "--- next actions plan"
+            python -mcraft_parts --work-dir=work --overlay-base=base --dry-run --show-skipped
+            echo "--- next actions execution"
+            python -mcraft_parts --work-dir=work --trace --overlay-base=base
+          EOF
+          echo "--- parts.yaml:"
+          cat parts.yaml
+          sudo bash run.sh
+          echo "* Check if hello executables installed"
+          test -x work/prime/usr/bin/hello && echo "hello"
+          test -x work/prime/usr/bin/bison && echo "bison"
+          test -x work/prime/usr/bin/m4 && echo "m4"
+          echo "* Check if /usr/bin/yacc is a symbolic link"
+          test -L work/prime/usr/bin/yacc && ls -l work/prime/usr/bin/yacc
+          echo "* Check if docs excluded"
+          test ! -d work/usr/share/doc/hello && echo "hello"
+          test ! -d work/usr/share/doc/bison && echo "bison"
+          echo "* Check created files"
+          grep "!dlrow ,olleH" work/prime/olleh.txt
+          grep "Hello, world!" work/prime/hello.txt
+          grep 5678 work/prime/etc/issue
+          echo "* Check whiteout files"
+          test -f work/prime/etc/.wh.shells && echo "/etc/shells"
+          test -f work/prime/etc/ssh/.wh..wh.opq && echo "/etc/ssh"
+          echo "--- prime files:"
+          find work/prime
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -130,6 +130,7 @@ jobs:
                 echo 5678 > "\$CRAFT_PARTS_OVERLAY/etc/issue"
               overlay:
                 - -usr/share
+                - -var/cache
               override-build: |
                 rev "\$CRAFT_PARTS_OVERLAY/hello.txt" > "\$CRAFT_PARTS_PART_INSTALL/olleh.txt"
           EOF
--- a/craft_parts/executor/part_handler.py
+++ b/craft_parts/executor/part_handler.py
@@ -165,19 +165,15 @@ class PartHandler:
         self._make_dirs()
 
         if self._part.has_overlay:
+            # install overlay packages
             overlay_packages = self._part.spec.overlay_packages
             if overlay_packages:
-                # must be in a separate context manager to avoid pychroot process leak
                 with overlays.LayerMount(
                     self._overlay_manager, top_part=self._part
                 ) as ctx:
                     ctx.install_packages(overlay_packages)
 
-            # remove temporary files and package cache from layer
-            shutil.rmtree(self._part.part_layer_dir / "var/cache", ignore_errors=True)
-            shutil.rmtree(self._part.part_layer_dir / "var/tmp", ignore_errors=True)
-            shutil.rmtree(self._part.part_layer_dir / "tmp", ignore_errors=True)
-
+            # execute overlay script
             with overlays.LayerMount(self._overlay_manager, top_part=self._part):
                 contents = self._run_step(
                     step_info=step_info,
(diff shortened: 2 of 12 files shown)


.github/workflows/tests.yaml
```diff
@@ -103,3 +103,63 @@ jobs:
         run: |
           source ${HOME}/.venv/bin/activate
           make test-integrations
+      - name: Run overlay smoke test
```
cmatsuoka, line 106: This is a test using two parts with overlays that runs on ci to make sure the basics are working. This test failed with pychroot with multiple process leaks and unmount errors.

craft_parts/executor/part_handler.py
```diff
@@ -173,6 +173,11 @@ def _run_overlay(self, step_info: StepInfo) -> StepState:
                 ) as ctx:
                     ctx.install_packages(overlay_packages)
 
+            # remove temporary files and package cache from layer
+            shutil.rmtree(self._part.part_layer_dir / "var/cache", ignore_errors=True)
+            shutil.rmtree(self._part.part_layer_dir / "var/tmp", ignore_errors=True)
+            shutil.rmtree(self._part.part_layer_dir / "tmp", ignore_errors=True)
+
```
sergiusens, line 180: this should not be done by craft-parts but by the user instead. This can break layering.

craft_parts/overlays/chroot.py
```diff
@@ -0,0 +1,148 @@
...
+
+    conn.send((res, None))
+    sys.exit()
+
+
+def _setup_chroot(path: Path) -> None:
+    """Prepare the chroot environment before executing the target function."""
+    logger.debug("setup chroot: %r", path)
+    if sys.platform == "linux":
+        _setup_chroot_linux(path)
```
sergiusens, line 88:
```suggestion
    if sys.platform != "linux":
        raise ....

    _setup_chroot_linux(path)
```

craft_parts/overlays/chroot.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def _cleanup_chroot(path: Path) -> None:
+    """Clean the chroot environment after executing the target function."""
+    logger.debug("cleanup chroot: %r", path)
+    if sys.platform == "linux":
```
sergiusens, line 94: similar to setup

craft_parts/overlays/chroot.py
sergiusens: the easy way to guard from this at a higher level is to make the overlay keywords error if not on linux, but this will pass silently (nothing happens effect)

cmatsuoka: Yes, we need a higher level guard.

craft_parts/executor/part_handler.py
cmatsuoka: I don't think this can break layering, we're removing temporary files directly from the topmost layer and exposing content that previous layers might have placed there. Since this is performed for all layers, these directories are never part of any layer and the base layer contents are always visible. That said, we could leave this cleanup operation to the user or to the application, but it's something that should be performed at some point before packing the primed payload.

craft_parts/overlays/chroot.py
cmatsuoka: Will address this issue in a follow-up PR.

craft_parts/overlays/chroot.py
cmatsuoka: Will address this issue in a follow-up PR.

craft_parts/executor/part_handler.py
cmatsuoka: Filesystem cleanup calls removed.

sergiusens (approved):

cmatsuoka:
Close/reopen to restart tests.

craft_parts/overlays/chroot.py
```diff
@@ -0,0 +1,147 @@
...
+_linux_mounts: List[_Mount] = [
+    _Mount(None, "/etc/resolv.conf", "/etc/resolv.conf", "--bind"),
+    _Mount("proc", "proc", "/proc", None),
+    _Mount("sysfs", "sysfs", "/sys", None),
+    _Mount(None, "/dev", "/dev", "--bind"),
+    _Mount("tmpfs", "tmpfs", "/dev/shm", None),
+]
+
+
+def _setup_chroot_linux(path: Path) -> None:
```
facundobatista, line 108: Please add a docstring.

craft_parts/overlays/chroot.py
```diff
@@ -0,0 +1,147 @@
...
+        if mountpoint.exists():
+            logger.debug("[pid=%d] mount %r on chroot", os.getpid(), str(mountpoint))
+            os_utils.mount(entry.src, str(mountpoint), *args)
+        else:
+            logger.debug(
+                "[pid=%d] mountpoint %r does not exist", os.getpid(), str(mountpoint)
+            )
+
+
+def _cleanup_chroot_linux(path: Path) -> None:
```
facundobatista, line 141: Please add a docstring.

craft_parts/overlays/chroot.py
```diff
@@ -0,0 +1,147 @@
...
+        # ``/run/systemd/resolve/stub-resolv.conf``. We want resolv.conf to be
+        # a regular file to bind-mount the host resolver configuration on.
+        #
+        # There's no need to restore the file to its original conditions because
+        # this operation happens on a temporary filesystem layer.
+        if entry.src == "/etc/resolv.conf":
+            if mountpoint.is_symlink():
+                mountpoint.unlink()
+                mountpoint.touch()
+            elif not mountpoint.exists() and Path(path, "etc").is_dir():
```
facundobatista, line 128: Instead of building the path again, kind of arbitrarily using "etc", why not just do `mountpoint.parent.is_dir()`?

craft_parts/overlays/chroot.py
```diff
@@ -0,0 +1,147 @@
...
+
+        mountpoint = path / entry.mountpoint.lstrip("/")
+
+        # Some images (such as cloudimgs) symlink ``/etc/resolv.conf`` to
+        # ``/run/systemd/resolve/stub-resolv.conf``. We want resolv.conf to be
+        # a regular file to bind-mount the host resolver configuration on.
+        #
+        # There's no need to restore the file to its original conditions because
+        # this operation happens on a temporary filesystem layer.
+        if entry.src == "/etc/resolv.conf":
```
facundobatista, line 124: I don't like this special case here. I understand that resolv.conf may need extra work, but why not make it generic? Like adding a parameter in `_Mount` above?

IOW, this looks like instead of current lines 111 and 112 (which I like), you'd have there:

```
if entry.src == "/etc/resolv.conf" or entry.src == "/dev":
    args.append("--bind")
```

facundobatista (changes requested):
Great work!! Added some questions and comments inline, thanks!!

craft_parts/overlays/chroot.py
cmatsuoka: Docstring added.

craft_parts/overlays/chroot.py
cmatsuoka: Docstring added.

craft_parts/overlays/chroot.py
```diff
@@ -0,0 +1,147 @@
...
+                "[pid=%d] mountpoint %r does not exist", os.getpid(), str(mountpoint)
+            )
+
+
+def _cleanup_chroot_linux(path: Path) -> None:
+    for entry in reversed(_linux_mounts):
+        mountpoint = path / entry.mountpoint.lstrip("/")
+
+        if mountpoint.exists():
+            logger.debug("[pid=%d] umount: %r", os.getpid(), str(mountpoint))
```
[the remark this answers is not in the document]
cmatsuoka: Done as suggested.

craft_parts/overlays/chroot.py
```diff
@@ -0,0 +1,147 @@
...
+        _cleanup_chroot_linux(path)
+
+
+_Mount = namedtuple("_Mount", ["fstype", "src", "mountpoint", "option"])
+
+_linux_mounts: List[_Mount] = [
+    _Mount(None, "/etc/resolv.conf", "/etc/resolv.conf", "--bind"),
+    _Mount("proc", "proc", "/proc", None),
+    _Mount("sysfs", "sysfs", "/sys", None),
+    _Mount(None, "/dev", "/dev", "--bind"),
```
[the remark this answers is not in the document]
sergiusens: proc is a virtual filesystem

```
$ mount |grep ^proc
proc on /proc type proc (rw,nosuid,nodev,noexec,relatime)
```

same for sysfs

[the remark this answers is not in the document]
cmatsuoka: In the `/dev` case we're bind-mounting the `/dev` directory to the `/dev` mountpoint inside the chroot. In the `proc` case we're mounting a filesystem with an arbitrary name (which happens to be `proc`) to the `/proc` mountpoint inside the chroot (so `/dev` is a directory, hence the slash, but `proc` and `sysfs` are names given to the devices that will be mounted)

craft_parts/overlays/chroot.py
cmatsuoka: Changed as suggested.

craft_parts/overlays/chroot.py
```diff
@@ -0,0 +1,147 @@
...
+def _cleanup_chroot(path: Path) -> None:
+    """Clean the chroot environment after executing the target function."""
+    logger.debug("cleanup chroot: %r", path)
+    if sys.platform == "linux":
+        _cleanup_chroot_linux(path)
+
+
+_Mount = namedtuple("_Mount", ["fstype", "src", "mountpoint", "option"])
+
+_linux_mounts: List[_Mount] = [
```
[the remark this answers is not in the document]
cmatsuoka: Added a comment about those being necessary to have basic funcionality inside the chroot environment including name resolution. I don't think there's formal documentation for this procedure, it's just common knowledge on how to make it work.

craft_parts/overlays/chroot.py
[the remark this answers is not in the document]
facundobatista: thanks

craft_parts/overlays/chroot.py
cmatsuoka: This is a special case handling that can be done separately, keeping the mount loop uniform. Changed the code to perform special case verification before entering the mount list processing.

facundobatista (approved):
Wonderful, thanks!!

PR #140, opened 2021-10-06
executor: use bash as script interpreter instead of /bin/sh (CRAFT-567)

cmatsuoka:
Snapcraft uses /bin/sh as the script interpreter when using plugins
V1 and /bin/bash when using plugins V2. To maintain compatibility,
change craft-parts to use /bin/bash as the script interpreter.

-----


facundobatista (approved):
Thanks!

PR #141, opened 2021-10-18
packages: load a stub AptCache on ImportError (CRAFT-574)

sergiusens:
The goal is to support non at enabled environments from working with
LXD or Multpass when running from a non snap in applications (like
Homebrew Linux).

Signed-off-by: Sergio Schvezov <sergio.schvezov@canonical.com>

-------

-----

### DIFF
--- a/craft_parts/packages/deb.py
+++ b/craft_parts/packages/deb.py
@@ -35,9 +35,16 @@ from .base import BaseRepository, get_pkg_name_parts, mark_origin_stage_package
 from .deb_package import DebPackage
 from .normalize import normalize
 
-if sys.platform == "linux":
-    # Ensure importing works on non-Linux.
+# Ensure importing works on non-Linux or linux without apt.
+# If the import fails provide a stub that appropriately fails
+# at runtime.
+try:
     from .apt_cache import AptCache
+
+    _apt_cache_available = True
+except ImportError:
+    _apt_cache_available = False
+
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


craft_parts/packages/deb.py
```diff
@@ -35,10 +35,17 @@
 from .deb_package import DebPackage
 from .normalize import normalize
 
-if sys.platform == "linux":
-    # Ensure importing works on non-Linux.
+# Ensure importing works on non-Linux or linux without apt.
+# If the import fails provide a stub that appropriately fails
```
facundobatista, line 39: What happens "if the import fails" is misleading (probably remained there from a previous implementation)

craft_parts/packages/deb.py
```diff
@@ -202,6 +209,18 @@
 }
 
 
+def _apt_cache_wrapper(method):
+    """Decorate a method to handle apt availability."""
+
+    @functools.wraps(method)
+    def wrapped(self, *args, **kwargs):
```
facundobatista, line 216: Signature here does not correspond to the decorated methods (which are class methods, receive `cls`, not `self`) -- probably just leaving `*args, **kwargs` is good enough.

craft_parts/packages/errors.py
```diff
@@ -26,6 +26,16 @@ class PackagesError(PartsError):
     """Base class for package handler errors."""
 
 
+class PackageBackendNotSupported(PartsError):
+    """Requested package resolved not supported on this host."""
+
+    def __init__(self, backend: str) -> None:
+        self.backend = backend
+        super().__init__(
+            brief=f"Package backed {backend!r} is not supported on this environment.",
```
facundobatista, line 35: s/backed/backend/

tests/unit/packages/test_errors.py
```diff
@@ -17,6 +17,14 @@
 from craft_parts.packages import errors
 
 
+def test_package_backend_not_supported():
+    err = errors.PackageBackendNotSupported("apt")
+    assert err.backend == "apt"
+    assert err.brief == ("Package backed 'apt' is not supported on this environment.")
```
facundobatista, line 23: Parenthesis not needed here.

craft_parts/packages/errors.py
sergiusens: fixed in 65ca1ee2bf780ef7a31e10e01738c14c34ae9d40

PR #150, opened 2021-11-01
plugins: add plugin environment validation (CRAFT-611)

cmatsuoka:
Plugins may require certain environment elements to be present in
order to build a part, regardless of how these elements were installed
on the system. For example, a compiler may have been installed from a
deb package, a snap, built from sources or even built by a different
part. This commit adds plugin environment validators, allowing plugins
to verify if the system environment is correct before building a part.

**Note to reviewers**: see the integration test for use cases.

-----

### DIFF
--- a/craft_parts/errors.py
+++ b/craft_parts/errors.py
@@ -395,6 +395,20 @@ class StageFilesConflict(PartsError):
         super().__init__(brief=brief, details=details)
 
 
+class PluginEnvironmentValidationError(PartsError):
+    """Plugin build script failed at runtime.
+
+    :param part_name: The name of the part being processed.
+    """
+
+    def __init__(self, *, part_name: str, reason: str):
+        self.part_name = part_name
+        self.reason = reason
+        brief = f"Environment validation failed for part {part_name!r}: {reason}."
+
+        super().__init__(brief=brief)
+
+
 class PluginBuildError(PartsError):
     """Plugin build script failed at runtime.
 
--- a/craft_parts/errors.py
+++ b/craft_parts/errors.py
@@ -396,7 +396,7 @@ class StageFilesConflict(PartsError):
 
 
 class PluginEnvironmentValidationError(PartsError):
-    """Plugin build script failed at runtime.
+    """Plugin environment validation failed at runtime.
 
     :param part_name: The name of the part being processed.
     """
--- a/craft_parts/executor/environment.py
+++ b/craft_parts/executor/environment.py
@@ -29,7 +29,7 @@ from craft_parts.utils import os_utils
 logger = logging.getLogger(__name__)
 
 
-def generate_part_environment(
+def generate_step_environment(
     *, part: Part, plugin: Plugin, step_info: StepInfo
 ) -> str:
     """Generate an environment to use during step execution.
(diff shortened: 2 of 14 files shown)


cmatsuoka:
Environment variable handling needs to be reworked, the different parts can't be handled in a single dictionary (due to the `PATH=$PATH:...` scenario).

craft_parts/executor/environment.py
```diff
@@ -80,6 +80,33 @@ def generate_part_environment(
...
+    :return: The environment to use when executing the step.
+    """
+    # Craft parts' say.
+    build_environment = _basic_environment_for_part(part=part, step_info=step_info)
+
+    step = step_info.step
+
+    # Plugin's say.
+    if step == Step.BUILD:
+        build_environment.update(plugin.get_build_environment())
```
cmatsuoka, line 101: We can't handle env vars using a single dictionary.

craft_parts/errors.py
```diff
@@ -395,6 +395,20 @@ def __init__(self, *, part_name: str, conflicting_files: List[str]):
         super().__init__(brief=brief, details=details)
 
 
+class PluginEnvironmentValidationError(PartsError):
+    """Plugin build script failed at runtime.
```
sergiusens, line 399:
```suggestion
    """Plugin validation script failed at runtime.
```

tests/integration/plugins/test_validate_environment.py
```diff
@@ -0,0 +1,281 @@
...
+        """
+        try:
+            output = self._execute("mytool")
+            if output.strip() != "ok":
+                raise errors.PluginEnvironmentValidationError(
+                    part_name=self._part_name,
+                    reason="mytool is expected to print ok",
+                )
+        except subprocess.CalledProcessError as err:
+            if err.returncode != 127:
```
sergiusens, line 80: code comment for 127 please

tests/integration/plugins/test_validate_environment.py
```diff
@@ -0,0 +1,281 @@
...
+    )
+
+    # plugins act on the build step
+    actions = lf.plan(Step.BUILD)
+
+    with lf.action_executor() as exe:
+        exe.execute(actions)
+
+
+def test_validate_plugin_created_in_part(new_dir):
```
sergiusens, line 165:
```suggestion
def test_validate_plugin_satisfied_with_part(new_dir):
```

craft_parts/plugins/validator.py
```diff
@@ -0,0 +1,87 @@
...
+            supplied by another part when the method is called from the execution
+            prologue. If the validation fails and list is empty, the error is
+            final (the missing elements can't be supplied by a different part).
+            The plugin may require a specific part name as a hint that the
+            part will attempt to supply missing environment elements.
+
+        :raises PluginEnvironmentValidationError: If the environment is invalid.
+        """
+
+    def _execute(self, cmd: str) -> str:
```
sergiusens, line 65: if pylint were enabled, the recommendation would be to make this public :-)

sergiusens:
Looks pretty good, I will do a local pass tomorrow

craft_parts/errors.py
cmatsuoka: Copy-and-paste strikes again.

tests/integration/plugins/test_validate_environment.py
cmatsuoka: I'll use a constant to replace the magic number.

tests/integration/plugins/test_validate_environment.py
cmatsuoka: Renaming as suggested.

craft_parts/plugins/validator.py
cmatsuoka: I run pylint locally, it actually doesn't care about this name in particular.

craft_parts/executor/part_handler.py
```diff
@@ -340,12 +341,25 @@ def _run_step(
         :return: If step is Stage or Prime, return a tuple of sets containing
             the step's file and directory artifacts.
         """
+        step_env = generate_step_environment(
+            part=self._part, plugin=self._plugin, step_info=step_info
+        )
+
+        # Validate build environment
```
sergiusens, line 348: add comment about differences with the one at the executor level

sergiusens (approved):

craft_parts/executor/part_handler.py
cmatsuoka: Added clarification.

craft_parts/plugins/validator.py
cmatsuoka: To clarify: this is a "protected" method implementing an execution helper intended to be used only by plugin validators. It executes a command in the same environment used to run the build step.

PR #153, opened 2021-11-22
sources: use pathlib paths in source provision calls

cmatsuoka:
Refactor source provision calls to use pathlib.Path instead of strings
and add missing type annotations for source and part_src_dir. Source
is typed as string because it can also be a URL.

-----

### DIFF
--- a/craft_parts/sources/git_source.py
+++ b/craft_parts/sources/git_source.py
@@ -16,7 +16,6 @@
 
 """Implement the git source handler."""
 
-import os
 import re
 import subprocess
 import sys
@@ -54,7 +53,7 @@ class GitSource(SourceHandler):
         return True
 
     @classmethod
-    def generate_version(cls, *, part_src_dir=None) -> str:
+    def generate_version(cls, *, part_src_dir: Optional[Path] = None) -> str:
         """Return the latest git tag from PWD or defined part_src_dir.
 
         The output depends on the use of annotated tags and will return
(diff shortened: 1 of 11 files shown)


craft_parts/sources/git_source.py
```diff
@@ -220,22 +219,23 @@ def _clone_new(self):
...
-        self._run(command + [self.source, self.part_src_dir])
+        self._run(command + [self.source, str(self.part_src_dir)])
 
         if self.source_commit:
             self._fetch_origin_commit()
             command = [
                 self.command,
                 "-C",
-                self.part_src_dir,
+                str(self.part_src_dir),
```
sergiusens, line 229: you should not need to str things passed to subprocess, right?

craft_parts/sources/snap_source.py
```diff
@@ -96,33 +96,39 @@ def provision(
...
         # -dest <pathname>: unsquash to <pathname>
-        with tempfile.TemporaryDirectory(prefix=snap_dir) as temp_dir:
-            extract_command = ["unsquashfs", "-force", "-dest", temp_dir, snap_file]
+        with tempfile.TemporaryDirectory(prefix=str(snap_file.parent)) as temp_dir:
+            extract_command = [
+                "unsquashfs",
+                "-force",
+                "-dest",
+                temp_dir,
+                str(snap_file),
```
sergiusens, line 120: this goes to subprocess too

craft_parts/sources/git_source.py
cmatsuoka: Yes, subprocess should do the right thing when pathlib.Path is used.

craft_parts/sources/snap_source.py
cmatsuoka: In this case the list goes to `_run_output` which expects `List[str]`.

cmatsuoka: Hmm, it seems that it will require more changes, so I'll leave this as is and fix it later in a different os.path-to-pathlib refactoring round.

cmatsuoka: Changed `_run_output` argument type to `Sequence` and fixed a couple of tests so that arguments can be passed as subprocess expects them.

cmatsuoka:
Amended small fix in PullError argument typing.

sergiusens (approved):

PR #154, opened 2021-11-25
python: unpin dependencies (CRAFT-663)

sergiusens:
- Remove requirements to always test with the latest
- Separate the docs target into its own requirements for readthedocs
- Ensure the right dependencies are used to create the documentation
  assets
- Remove black's file filter which ignores gitignore
- Use 20.04 as the only base with a spread of python versions
- Ensure the sudo context of the GHA for overlays runs the correct
  python interpreter

Signed-off-by: Sergio Schvezov <sergio.schvezov@canonical.com>

-------

-----

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -14,54 +14,43 @@ jobs:
         uses: actions/checkout@v2
         with:
           fetch-depth: 0
-      - name: Add local bin to PATH
-        run: |
-          echo "${HOME}/.local/bin" >> $GITHUB_PATH
       - name: Install python packages and dependencies
         run: |
           sudo apt update
-          sudo apt install -y python3-pip python3-venv libapt-pkg-dev
-          python3 -m venv ${HOME}/.venv
-          source ${HOME}/.venv/bin/activate
-          pip install -U pip wheel setuptools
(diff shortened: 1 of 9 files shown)


.github/workflows/tests.yaml
```diff
@@ -14,98 +14,74 @@ jobs:
...
           sudo apt update
-          sudo apt install -y python3-pip python3-venv libapt-pkg-dev
-          python3 -m venv ${HOME}/.venv
-          source ${HOME}/.venv/bin/activate
-          pip install -U pip wheel setuptools
-          pip install -U -r requirements.txt -r requirements-dev.txt -r requirements-focal.txt
-          pip install -e .
+          sudo apt install -y libapt-pkg-dev
+          pip install -U \
+                         https://launchpad.net/ubuntu/+archive/primary/+sourcefiles/python-distutils-extra/2.43/python-distutils-extra_2.43.tar.xz \
```
cmatsuoka, line 22: distutils-extra wasn't in `requirements-focal.txt` but I assume it's required since you added it here.

cmatsuoka (approved):
Looks good.

.github/workflows/tests.yaml
sergiusens: it is a dependency of python apt unfortunately and we want the right one

PR #172, opened 2022-02-14
executor: use source type specified in part (CRAFT-831)

### LINKED ISSUE #171: Source type not correctly recognized
Explicit `source-type` declaration not correctly recognized in some cases. E.g. even with `source-type: git` a git repository URL is not recognized unless ending in `.git`.

cmatsuoka:
When obtaining the consolidated list of build packages, honor the
source type specified by the user before attempting to infer it
from the source URI. Fixes #171.

-----


sergiusens (approved):
I see the build packages thing there and notice gcc and have doubts if we should add it or not as a default

PR #184, opened 2022-03-16
plugins: cmake

mr-cal:

-----

Port `cmake` plugin to `craft-parts`.

(CRAFT-906)

### DIFF
--- a/craft_parts/plugins/cmake_plugin.py
+++ b/craft_parts/plugins/cmake_plugin.py
@@ -1,0 +1,118 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2020-2022 Canonical Ltd.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License version 3 as published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""The cmake plugin."""
+
+from typing import Any, Dict, List, Set, cast
+
+from .base import Plugin, PluginModel, extract_plugin_properties
+from .properties import PluginProperties
+
+
+class CMakePluginProperties(PluginProperties, PluginModel):
+    """The part properties used by the cmake plugin."""
+
+    cmake_parameters: List[str] = []
+    cmake_generator: str = "Unix Makefiles"
+
+    # part properties required by the plugin
+    source: str
+
+    @classmethod
+    def unmarshal(cls, data: Dict[str, Any]):
+        """Populate class attributes from the part specification.
+
+        :param data: A dictionary containing part properties.
+
+        :return: The populated plugin properties data object.
+
+        :raise pydantic.ValidationError: If validation fails.
+        """
+        plugin_data = extract_plugin_properties(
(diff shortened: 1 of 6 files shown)


tests/integration/plugins/test_cmake.py
```diff
@@ -0,0 +1,91 @@
...
+import yaml
+
+from craft_parts import LifecycleManager, Step
+
+
+@pytest.fixture
+def cmake():
+    subprocess.run(["pip", "install", "cmake"], check=True)
+    yield
+    subprocess.run(["pip", "uninstall", "cmake", "--yes"], check=True)
```
cmatsuoka, line 31: This is an interesting way to install cmake, but I think it's better to just install it from the archive when preparing the CI environment in the [github workflow](https://github.com/canonical/craft-parts/blob/7aa01bd45fb8652a09484f46ee77278f03a50f50/.github/workflows/tests.yaml#L82).

tests/integration/plugins/test_cmake.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+@pytest.mark.usefixtures("mocker")
```
cmatsuoka, line 34: Is this necessary? I don't think we can use the mocker fixture this way.

tests/integration/plugins/test_cmake.py
```diff
@@ -0,0 +1,91 @@
...
+            """\
+            cmake_minimum_required(VERSION 2.6)
+            project(cmake-hello C)
+            add_executable(cmake-hello hello.c)
+            install(TARGETS cmake-hello RUNTIME DESTINATION bin)
+            """
+        )
+    )
+
+    # ninja is installed in the ci test setup
```
cmatsuoka, line 81: This should be cmake, if we install it in the ci setup.

craft_parts/plugins/cmake_plugin.py
```diff
@@ -0,0 +1,125 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2020 Canonical Ltd.
```
cmatsuoka, line 3: Since we're porting this to craft-parts we should use 2020-2022 in the copyright notice.

tests/integration/plugins/test_cmake.py
```diff
@@ -0,0 +1,91 @@
...
+
+@pytest.fixture
+def cmake():
+    subprocess.run(["pip", "install", "cmake"], check=True)
+    yield
+    subprocess.run(["pip", "uninstall", "cmake", "--yes"], check=True)
+
+
+@pytest.mark.usefixtures("mocker")
+@pytest.mark.usefixtures("cmake")
```
cmatsuoka, line 35: This should also be unnecessary if we install cmake in the ci setup.

cmatsuoka (changes requested):

mr-cal:
Thanks for the thorough review, there was more cruft leftover than I realized.

craft_parts/plugins/cmake_plugin.py
```diff
@@ -0,0 +1,125 @@
...
+          parameters to pass to the build using the common cmake semantics.
+
+        - cmake-generator
+          (string; default: "Unix Makefiles")
+          Determine what native build system is to be used.
+          Can be either `Ninja` or `Unix Makefiles` (default).
+
+    This plugin also interprets these specific build-environment entries:
+
+        - SNAPCRAFT_CMAKE_ARGS
```
sergiusens, line 73: I would remove this fow now, we used this from the extension for kde but it shouldbe solved some other way

sergiusens: we can re-add later if creativity does not allow for us to get there :-)

sergiusens (approved):
we don't seem to be using environment validation here, for perhaps to account for someone wanting to use a newer cmake but this is a good start.

cmatsuoka (approved):
Thanks Callahan!

PR #185, opened 2022-03-18
craftctl: update to comply with specification

cmatsuoka:
The specification of craftctl lists the expected behavior of the
`default`, `set`, and `get` commands, including syntax consistent
with snapctl (when applicable), name validation, protection against
set after consumption, and restricting variable setting to a single
application-specified part. Update the existing implementation
to follow the specification.

-----
CRAFT-935
CRAFT-75


craft_parts/ctl.py
```diff
@@ -20,41 +20,65 @@
...
-        client(cmd, param)
-    except RuntimeError as err:
-        logger.error("ctl error: %s", err)
-        sys.exit(1)
+    Craftctl is used to execute built-in step handlers and to get and set
+    variables in the running parts processor context.
+    """
 
+    @classmethod
+    def handle_command(
```
sergiusens, line 36: this is personal preference, but can we go with the model of registering command handlers instead of using inheritance for this? Any pros and cons you analyzed here?

craft_parts/ctl.py
```diff
@@ -70,8 +94,33 @@ def client(cmd: str, args: List[str]):
...
 
     with open(feedback_fifo, "r") as fifo:
-        feedback = fifo.readline().strip()
+        feedback = fifo.readline().split(" ", 1)
+
+    status = feedback[0]
+    message = feedback[1].strip() if len(feedback) > 1 else ""
+    retval = None
+
+    if status == "OK":
```
sergiusens, line 103: can you add a comment about what sends the OK and ERR (I can tell in this PR, but in the future I might certainly  forget where this comes from :-))

craft_parts/ctl.py
sergiusens: agreed to leave it out for now as nothing needs this

craft_parts/infos.py
```diff
@@ -131,18 +143,55 @@ def project_options(self) -> Dict[str, Any]:
...
+
+        if name in self._consumed_project_vars:
+            raise ValueError(
+                f"cannot set variable {name!r}, it has already been consumed"
+            )
+
+        if self._project_vars_part_name in [None, part_name]:
+            self._project_vars[name] = value
+
+    def get_project_var(self, name: str, *, consume: bool = True) -> str:
```
sergiusens, line 166: this leaked

craft_parts/ctl.py
cmatsuoka: Removed from source code.

craft_parts/ctl.py
cmatsuoka: Added a reminder note.

craft_parts/infos.py
cmatsuoka: The method is correct, however the docstring was copy-and-pasted from a different method (it says "set" instead of "get")

craft_parts/executor/step_handler.py
```diff
@@ -288,32 +293,62 @@ def _handle_control_api(self, scriptlet_name: str, function_call: str) -> None:
...
-        elif function_name == "set":
-            if len(function_args) != 2:
+            self._execute_builtin_handler(step)
+        elif cmd_name == "set":
+            if len(cmd_args) < 1:
+                raise invalid_control_api_call(
+                    message=(f"invalid arguments to command {cmd_name!r}"),
+                )
+
+            for arg in cmd_args:
```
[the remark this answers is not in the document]
cmatsuoka: Changed to allow only one variable to be set per set command.

craft_parts/infos.py
cmatsuoka: This method is intended to be used by the application to consume project variables. Setting variables after they're consumed must result in error to prevent unexpected/confusing situations when a variable is set after being consumed and the new value is not actually used.

craft_parts/infos.py
```diff
@@ -227,15 +276,25 @@ def part_state_dir(self) -> Path:
...
         :param name: The custom argument name.
         :param value: The new custom argument value.
 
         :raise ValueError: If there is no custom argument with the given name.
         """
-        self._project_info.set_custom_argument(name, value)
+        self._project_info.set_project_var(name, value, part_name=self._part_name)
+
+    def get_project_var(self, name: str, *, consume=True) -> str:
+        """Set the value of a project variable.
```
sergiusens, line 290:
```suggestion
        """Get the value of a project variable.
```

craft_parts/executor/step_handler.py
```diff
@@ -288,32 +293,63 @@ def _handle_control_api(self, scriptlet_name: str, function_call: str) -> None:
...
-        function_name = function_json["function"]
-        function_args = function_json["args"]
+        cmd_name = function_json["function"]
+        cmd_args = function_json["args"]
+
+        return self._process_api_commands(
+            cmd_name, cmd_args, step=step, scriptlet_name=scriptlet_name
+        )
+
+    def _process_api_commands(
```
sergiusens, line 303: this method does not seem to gain us much as _handle_control_api is mostly a passthrough for this

craft_parts/executor/step_handler.py
```diff
@@ -288,32 +293,63 @@ def _handle_control_api(self, scriptlet_name: str, function_call: str) -> None:
...
 
         invalid_control_api_call = functools.partial(
             errors.InvalidControlAPICall,
             part_name=self._part.name,
             scriptlet_name=scriptlet_name,
         )
 
-        if function_name in ["pull", "build", "stage", "prime"]:
-            if len(function_args) > 0:
+        if cmd_name == "default":
```
sergiusens, line 315: we might be missing explanation documentation attached to this this

sergiusens: or about craftctl in general

tests/integration/lifecycle/test_craftctl.py
```diff
@@ -0,0 +1,428 @@
...
+            "PYTHONPATH": str(TESTS_DIR.parent),
+        },
+    )
+
+    mocker.patch("craft_parts.utils.os_utils.mount")
+    mocker.patch("craft_parts.utils.os_utils.mount_overlayfs")
+    mocker.patch("craft_parts.utils.os_utils.umount")
+
+
+def test_craftctl_default(new_dir, capfd, mocker):
```
sergiusens, line 49: this is a large test, a code comment in each block describing the intention of the assertions would be nice

tests/integration/lifecycle/test_craftctl.py
```diff
@@ -0,0 +1,428 @@
...
+    parts = yaml.safe_load(parts_yaml)
+
+    lf = craft_parts.LifecycleManager(
+        parts,
+        application_name="test_set",
+        cache_dir=new_dir,
+        project_vars={"myvar": ""},
+    )
+    with lf.action_executor() as ctx:
+        ctx.execute(Action("foo", Step.PULL))
```
sergiusens, line 177: we need to test later running Step.BUILD and retrieving the value of a previously stored Step.PULL (which would not run due to the step caching)

sergiusens:
Looks pretty good, just some nitpicks and perhaps some documentation for craftctl

craft_parts/executor/step_handler.py
cmatsuoka: Split into a new method to reduce complexity (linters also complained about that -- could disable the warning and coalesce back into a single method, but having it separated looks better for organization)

craft_parts/executor/step_handler.py
cmatsuoka: Will add documentation in a separate PR, creating a task for this.

tests/integration/lifecycle/test_craftctl.py
cmatsuoka: Ah I see the problem here. We need to store project vars in step state. Will implement and test.

cmatsuoka: Store project variables in step state along with update flag. On a new execution the sequencer reads the state and passes any project var states in skip actions. Executor updates the variables if action is skipped.

craft_parts/actions.py
```diff
@@ -63,9 +66,12 @@ class Action:
     :param action_type: Action to run for this step.
     :param reason: A textual description of why this action should be
         executed.
+    :param project_vars: The values of project variables from a previous
+        execution if the action type is ``ActionType.SKIP``.
```
sergiusens, line 70: this makes it seem like it is a failure to set this if the action type is not SKIP, we either should have a guard for this or edit the text (useful for SKIP, best used with, or no comment at all).

craft_parts/infos.py
```diff
@@ -141,48 +153,65 @@ def project_options(self) -> Dict[str, Any]:
...
-            raise ValueError(
-                f"cannot set variable {name!r}, it has already been consumed"
-            )
+        if self._project_vars[name].updated:
+            raise RuntimeError(f"variable {name!r} can be set only once")
 
         if self._project_vars_part_name in [None, part_name]:
-            self._project_vars[name] = value
+            self._project_vars[name].value = value
+            self._project_vars[name].updated = True
```
sergiusens, line 188: just an idea, you could eventually set .updated by overriding __setattr__  or __setitem__

craft_parts/state_manager/states.py
```diff
@@ -54,6 +55,14 @@ def load_step_state(part: Part, step: Step) -> Optional[StepState]:
     with open(filename) as yaml_file:
         state_data = yaml.safe_load(yaml_file)
 
+    # fix project vars in loaded state data
+    opt = state_data.get("project-options")
```
sergiusens, line 59: variable naming could be improve in this block

craft_parts/state_manager/states.py
```diff
@@ -54,6 +55,14 @@ def load_step_state(part: Part, step: Step) -> Optional[StepState]:
...
+    if opt:
+        pvars = opt.get("project_vars")
+        if pvars:
+            for key, val in pvars.items():
+                pvars[key] = ProjectVar(**val)
```
sergiusens, line 64: this might be working due to how the internals of dictionaries work, but I would prefer to see some for of reassignment here to state_data or a code comment.

craft_parts/actions.py
cmatsuoka: Changed to "... from a previous execution, used if action type is..."

craft_parts/infos.py
cmatsuoka: I considered that, but I tried to avoid using tricks that could make it harder to follow in the future. But we can improve this later.

craft_parts/state_manager/states.py
cmatsuoka: Expanded variable assignment and added extra comment. Ideally we should devise a way to add proper typing to the `project_options` definition so pydantic could automatically create the object.

tests/integration/lifecycle/test_craftctl.py
cmatsuoka: Added comments to blocks.

sergiusens (approved):
Excellent

PR #186, opened 2022-03-18
executor: standardize environment variable prefix

cmatsuoka:
Instead of using application-specific lifecycle context clients such
as snapcraftctl or charmcraftctl, craft-parts provides craftctl as
a generic client to be used by all applications. Following the same
motivation, use a common `CRAFT_` prefix for environment variables
used inside part scriptlets.

-----
CRAFT-936


sergiusens (commented):
only change I would make is to prefix everything with either CRAFT_PART or CRAFT_PARTS

cmatsuoka:
I considered using `CRAFT_PARTS` as the prefix but names like `CRAFT_PARTS_PART_SRC` or `CRAFT_PARTS_PART_BUILD` looked strange/repetitive. The logic used in the `CRAFT_` prefix is the same transformation used in snapcraftctl to craftctl, but I agree that the final variables look a bit "unbalanced", especially in variables containing first-level directories like `CRAFT_PRIME`. There are a few other options we can discuss, such as adding a `_DIR` suffix to directory variables (`CRAFT_PART_SRC_DIR`, `CRAFT_PART_BUILD_DIR`, `CRAFT_PRIME_DIR`) to avoid repetition and allow the user to be sure that the variable is a path name.

cmatsuoka:
Keeping variable names similar to Snapcraft's only replacing the prefix makes it easier for people willing to migrate.

sergiusens (approved):

cmatsuoka:
Added `CRAFT_PROJECT_DIR` and rebased on main.

PR #190, opened 2022-03-29
craftctl: set variables from parts adopting metadata

cmatsuoka:
Restrict variable setting to the part specified when creating the
parts lifecycle. Applications that don't declare project variables
don't need to specify an adopting part.

-----
CRAFT-963


sergiusens (approved):
Looks good. I am not entirely confident we want this for charmcraft, but this can certainly be relaxed later.

cmatsuoka:
We'll probably want to refactor project variable handling later to add validation, we can also adjust setting scope and adoption to match requirements.

PR #194, opened 2022-04-01
Release 1.4.2

cmatsuoka:
## Changelog

* Fix craftctl error handling
* Fix long recursions in dirty step verification

-----

### DIFF
--- a/craft_parts/executor/step_handler.py
+++ b/craft_parts/executor/step_handler.py
@@ -247,22 +247,22 @@ class StepHandler:
             try:
                 while status is None:
                     function_call = call_fifo.read()
-                    if function_call:
-                        # Handle the function and send feedback to caller.
+                    if not function_call:
+                        status = process.poll()
+                        time.sleep(0.1)  # Don't loop TOO busily
+                        continue
+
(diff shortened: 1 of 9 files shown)


craft_parts/executor/step_handler.py
```diff
@@ -247,22 +247,22 @@ def run_scriptlet(
             try:
                 while status is None:
                     function_call = call_fifo.read()
-                    if function_call:
-                        # Handle the function and send feedback to caller.
+                    if not function_call:
+                        status = process.poll()
+                        time.sleep(0.1)  # Don't loop TOO busily
+                        continue
+
```
facundobatista, line 254: Is this safe? if status is None, and function_call is False, you're potentially in a "forever loop" situation...

facundobatista: Note that the previous code *always* changed status, no matter the value of `function_call`...

cmatsuoka: The script execution will eventually end, and status won't be `None` in this case. FIFO reading will loop until something is read, or the script ends (this FIFO is non-blocking).

sergiusens (approved):

PR #198, opened 2022-04-13
plugins: rust

mr-cal:

-----

Port the `rust` plugin from [`snapcraft`](https://github.com/snapcore/snapcraft/blob/main/snapcraft/plugins/v2/rust.py) to `craft-parts`.

Will land this PR after https://github.com/canonical/craft-parts/pull/199.

CRAFT-955

### DIFF
--- a/craft_parts/plugins/rust_plugin.py
+++ b/craft_parts/plugins/rust_plugin.py
@@ -1,0 +1,131 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2022 Canonical Ltd.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License version 3 as published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""The rust plugin."""
+
+import logging
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, cast
+
+from pydantic import conlist
+
+from . import validator
+from .base import Plugin, PluginModel, extract_plugin_properties
+from .properties import PluginProperties
+
+logger = logging.getLogger(__name__)
+
+# A workaround for mypy false positives
+# see https://github.com/samuelcolvin/pydantic/issues/975#issuecomment-551147305
+if TYPE_CHECKING:
+    UniqueStrList = List[str]
+else:
+    UniqueStrList = conlist(str, unique_items=True)
+
+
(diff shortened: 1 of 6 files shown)


craft_parts/plugins/rust_plugin.py
```diff
@@ -0,0 +1,143 @@
...
+    :param part_name: The part whose build environment is being validated.
+    :param env: A string containing the build step environment setup.
+    """
+
+    def validate_environment(self, *, part_dependencies: Optional[List[str]] = None):
+        """Ensure the environment has the dependencies to build Rust applications.
+
+        :param part_dependencies: A list of the parts this part depends on.
+        """
+        for dependency in ["curl", "gcc", "git"]:
```
mr-cal, line 75: @sergiusens - I moved the dependencies from `get_build_packages()` to `validate_environment()`.  
Is this what you intended for me to do?

sergiusens: Please leave gcc in get_build_packages and git. Completely remove curl and you should be verifying that the rustc and cargo commands are available

craft_parts/plugins/rust_plugin.py
```diff
@@ -0,0 +1,143 @@
...
+
+        rustup_command = dedent(
+            """\
+        if ! command -v rustup 2>/dev/null; then
+            curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs |
+            sh -s -- -y --no-modify-path --profile=minimal
+            export PATH="${HOME}/.cargo/bin:${PATH}"
+        fi
+        """
+        )
```
sergiusens, line 124: This is what the environment should validate.

sergiusens (changes requested):

craft_parts/plugins/rust_plugin.py
mr-cal: That makes much more sense!

craft_parts/plugins/rust_plugin.py
```diff
@@ -0,0 +1,131 @@
...
+        """Return a set of required snaps to install in the build environment."""
+        return set()
+
+    def get_build_packages(self) -> Set[str]:
+        """Return a set of required packages to install in the build environment."""
+        return {"gcc", "git"}
+
+    def get_build_environment(self) -> Dict[str, str]:
+        """Return a dictionary with the environment to use in the build step."""
+        return {"PATH": "${HOME}/.cargo/bin:${PATH}"}
```
sergiusens, line 109: this should be empty too, we don't know where folks would install their rust stuff, I guess we can leave it for now

tests/integration/plugins/test_rust_plugin.py
```diff
@@ -0,0 +1,215 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
```
sergiusens, line 1: can you rename this file to test_rust.py so it follows the file naming pattern in that test dir?

craft_parts/plugins/rust_plugin.py
mr-cal: Agreed on leaving it for now.
If someone creates an issue regarding a custom path for their rust binaries inside a build environment, then we can re-address.

sergiusens (approved):

PR #199, opened 2022-04-13
tests: refactor plugin dependency tests

mr-cal:

-----

Replace redundant fixtures with a fixture factory.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -18,6 +18,8 @@ import http.server
 import os
 import tempfile
 import threading
+from pathlib import Path
+from typing import Optional
 from unittest import mock
 
 import pytest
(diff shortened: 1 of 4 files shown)


sergiusens (commented):
Looks good, but should this fixture be only one and in conftest.py?

mr-cal:
Ah, `conftest.py` is exactly what I was looking for.  Thanks.

tests/conftest.py
```diff
@@ -116,3 +118,38 @@ def fake_snapd():
 def fake_snap_command(mocker):
     """Mock the snap command."""
     return FakeSnapCommand(mocker)
+
+
+@pytest.fixture()
+def dependency_fixture(new_dir):
+    """Fixture factory for dependencies."""
+
+    def _dependency_fixture(
```
sergiusens, line 127: since the scope of this is dependency_fixture, this does not need to be private

sergiusens (commented):
Looks good, just a minor nit

sergiusens (approved):

PR #200, opened 2022-04-14
plugins: npm

mr-cal:

-----

Port the `npm` plugin from [`snapcraft`](https://github.com/snapcore/snapcraft/blob/main/snapcraft/plugins/v2/npm.py) to `craft-parts`.

CRAFT-957

### DIFF
--- a/craft_parts/plugins/npm_plugin.py
+++ b/craft_parts/plugins/npm_plugin.py
@@ -1,0 +1,187 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2022 Canonical Ltd.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License version 3 as published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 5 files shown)


sergiusens (changes requested):
I have been thinking about this one some more, and I think we might just be better off bringing in the plugin verbatim as this will produce a resulting part that cannot run as node is never really copied into CRAFT_PART_INSTALL. If you have a better idea I am all ears :-)

mr-cal:
As discussed, I added a new flag called `npm-include-node`, which defaults to False.
Setting it to true will download and include `node`.

craft_parts/plugins/npm_plugin.py
```diff
@@ -0,0 +1,182 @@
...
+        """Return a set of required snaps to install in the build environment."""
+        return set()
+
+    def get_build_packages(self) -> Set[str]:
+        """Return a set of required packages to install in the build environment."""
+        return {"curl", "gcc"}
+
+    def get_build_environment(self) -> Dict[str, str]:
+        """Return a dictionary with the environment to use in the build step."""
+        return dict(PATH="${CRAFT_PART_INSTALL}/bin:${PATH}")
```
sergiusens, line 151: this should only be set if options.npm_include_node is True, right?

mr-cal: Right. And the same is true for `curl` for `get_build_packages()`.  
I added logic and unit tests for both.

sergiusens (approved):

PR #201, opened 2022-04-14
plugins: conda

mr-cal:

-----
Port the `conda` plugin from [`snapcraft`](https://github.com/snapcore/snapcraft/blob/main/snapcraft/plugins/v2/conda.py) to `craft-parts`.

(CRAFT-1030)

### DIFF
--- a/craft_parts/plugins/conda_plugin.py
+++ b/craft_parts/plugins/conda_plugin.py
@@ -1,0 +1,123 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2022 Canonical Ltd.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License version 3 as published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""The conda plugin."""
+
+import logging
+from typing import Any, Dict, List, Optional, Set, cast
+
+from . import validator
+from .base import Plugin, PluginModel, extract_plugin_properties
+from .properties import PluginProperties
+
+logger = logging.getLogger(__name__)
+
+
+class CondaPluginProperties(PluginProperties, PluginModel):
+    """The part properties used by the conda plugin."""
+
+    # part properties required by the plugin
+    conda_packages: Optional[List[str]] = None
+    conda_python_version: Optional[str] = None
+    conda_install_prefix: str = "/snap/${CRAFT_PROJECT_NAME}/current"
+
+    @classmethod
+    def unmarshal(cls, data: Dict[str, Any]) -> "CondaPluginProperties":
+        """Populate class attributes from the part specification.
+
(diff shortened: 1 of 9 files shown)


craft_parts/plugins/conda_plugin.py
```diff
@@ -0,0 +1,119 @@
...
+    def get_build_environment(self) -> Dict[str, str]:
+        """Return a dictionary with the environment to use in the build step."""
+        return {}
+
+    def get_build_commands(self) -> List[str]:
+        """Return a list of commands to run during the build step."""
+        options = cast(CondaPluginProperties, self._options)
+
+        deploy_command = [
+            "CONDA_TARGET_PREFIX_OVERRIDE=/snap/${CRAFT_PROJECT_NAME}/current",
```
mr-cal, line 106: I took this line from [snapcraft](https://github.com/snapcore/snapcraft/blob/0224b48ea0e1cc7378e1ad392b41466f005ae3e6/snapcraft/plugins/v2/conda.py#L138).  It's limits the flexibility of the plugin, but I've been unable to find any better solution.

mr-cal:
Tested and confirmed conda projects run with snapcraft 7.x, although I will need to submit the following changes for snapcraft:
1. Pass `project_name` to `craft_parts.lifecycle_manager`.
2. Convert underscores to hyphens when writing `meta/snap.yaml`

craft_parts/plugins/conda_plugin.py
```diff
@@ -0,0 +1,123 @@
...
+logger = logging.getLogger(__name__)
+
+
+class CondaPluginProperties(PluginProperties, PluginModel):
+    """The part properties used by the conda plugin."""
+
+    # part properties required by the plugin
+    conda_packages: Optional[List[str]] = None
+    conda_python_version: Optional[str] = None
+    conda_install_prefix: str = "/snap/${CRAFT_PROJECT_NAME}/current"
```
sergiusens, line 35: is the project name not available through options?

craft_parts/plugins/conda_plugin.py
```diff
@@ -0,0 +1,123 @@
...
+    def get_build_commands(self) -> List[str]:
+        """Return a list of commands to run during the build step."""
+        options = cast(CondaPluginProperties, self._options)
+
+        deploy_command = [
+            f"CONDA_TARGET_PREFIX_OVERRIDE={options.conda_install_prefix}",
+            "conda",
+            "create",
+            "--prefix",
+            "$CRAFT_PART_INSTALL",
```
sergiusens, line 114: is the CRAFT_PART_INSTALL not available through options?

craft_parts/plugins/conda_plugin.py
sergiusens: this is probably a snapcraft plugin if this is the default

cmatsuoka: As discussed with @sergiusens, craft-parts plugins can't assume the resulting artifact will be a snap. If this is essential for conda to work, this should be a snapcraft plugin instead.

mr-cal: How about this logic in `get_build_commands`?
```
        if options.conda_install_prefix:
            install_prefix = options.conda_install_prefix
        elif self._part_info.application_name == "snapcraft":
            install_prefix = f"/snap/{self._part_info.project_name}/current"
        else:
            install_prefix = str(self._part_info.part_install_dir)

        deploy_command = [
            f"CONDA_TARGET_PREFIX_OVERRIDE={install_prefix}",
```

sergiusens:
I would just remove the install prefix default and think of a way to add defaults to the model from the application side

sergiusens:
Closing in favor of https://github.com/snapcore/snapcraft/pull/3703

PR #220, opened 2022-05-17
executor: enable application-defined environment variables

cmatsuoka:
Allow application-defined variables to be used during step scriptlet
execution. Applications can use the pre-step callback to set step
environment variables or the prologue callback to set global environment
variables.

-----
CRAFT-1070

### DIFF
--- a/craft_parts/executor/environment.py
+++ b/craft_parts/executor/environment.py
@@ -20,7 +20,7 @@ import io
 import logging
 from typing import Dict, Iterable
 
-from craft_parts.infos import StepInfo
+from craft_parts.infos import ProjectInfo, StepInfo
 from craft_parts.parts import Part
 from craft_parts.plugins import Plugin
 from craft_parts.steps import Step
@@ -41,20 +41,16 @@ def generate_step_environment(
     :return: The environment to use when executing the step.
     """
     # Craft parts' say.
-    our_build_environment = _basic_environment_for_part(part=part, step_info=step_info)
-
-    step = step_info.step
+    parts_environment = _basic_environment_for_part(part=part, step_info=step_info)
 
     # Plugin's say.
-    if step == Step.BUILD:
+    if step_info.step == Step.BUILD:
         plugin_environment = plugin.get_build_environment()
     else:
         plugin_environment = {}
 
     # Part's (user) say.
-    user_build_environment = part.spec.build_environment
-    if not user_build_environment:
-        user_build_environment = []
+    user_environment = part.spec.build_environment or []
 
     # Create the script.
     with io.StringIO() as run_environment:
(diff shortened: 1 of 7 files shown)


cmatsuoka:
> As a user of craft-parts, I'd appreciate some documentation with new features. Thanks for considering!

The callbacks API is still unstable, but we'll document it after it's considered stable for production use (we're exercising the experimental API with application-defined variables in Snapcraft and some changes may still be required).

tests/unit/executor/test_environment.py
```diff
@@ -165,9 +170,11 @@ def test_generate_step_environment_no_project_name(new_dir):
         export CXXFLAGS="-isystem {new_dir}/parts/p1/install/usr/include"
         export LDFLAGS="-L{new_dir}/stage/lib -L{new_dir}/stage/usr/lib -L{new_dir}/stage/lib/aarch64-linux-gnu"
         export PKG_CONFIG_PATH="{new_dir}/stage/usr/share/pkgconfig"
-        ## Plugin Environment
+        ## Plugin environment
         export PLUGIN_ENVVAR="from_plugin"
-        ## User Environment
+        ## Application environment
+        export APP_ENVVAR="from_app"
```
sergiusens, line 176: the test name does not match what is being tested

sergiusens (commented):
Looks good, but if we are not doing docs, an integration test would be nice. Docs should be fine if we document the api as unstable too

cmatsuoka:
Reworked the PR a bit to separate environment variables into global and step environment variables. This will allow global environment variable substitution later. Also added an integration that uses pre-step callbacks to set application-defined step environment variables and the prologue callback to set application-defined global environment variables.

tests/unit/executor/test_environment.py
cmatsuoka: Removed application variable setting from this test.

craft_parts/executor/environment.py
```diff
@@ -63,16 +59,22 @@ def generate_step_environment(
...
             print(f'export {key}="{val}"', file=run_environment)
 
-        print("## Plugin Environment", file=run_environment)
+        print("## Plugin environment", file=run_environment)
         for key, val in plugin_environment.items():
             print(f'export {key}="{val}"', file=run_environment)
 
-        print("## User Environment", file=run_environment)
-        for env in user_build_environment:
+        print("## Application environment", file=run_environment)
```
sergiusens, line 70: application environment should come before part and plugin as these could reference the application provided variables

craft_parts/executor/environment.py
```diff
@@ -88,7 +90,8 @@ def _basic_environment_for_part(part: Part, *, step_info: StepInfo) -> Dict[str,
 
     :return: A dictionary containing the built-in environment.
     """
-    part_environment: Dict[str, str] = _get_step_environment(step_info)
+    part_environment: Dict[str, str] = _get_global_environment(step_info.project_info)
+    part_environment.update(_get_step_environment(step_info))
```
sergiusens, line 94: maybe update the part_environment dict with information from the application here as from the point of view of the user, it is all part of the same thing

cmatsuoka: Updated to include global variables when getting the step environment. The separate call for global will be used later when expanding global variables.

cmatsuoka:
Addressed issues and rebased on current main.

sergiusens (approved):

PR #222, opened 2022-05-19
packages: update the cache before creating list

sergiusens:

-----
CRAFT-1064


cmatsuoka (approved):
Looks good with the overlay fixes and the master switch to turn off updates is very useful when running the cli as a regular user.

PR #224, opened 2022-05-19
plugin: plugin validation with only one part dep

sergiusens:
Setup plugin validation so it only mentions a dependency on one
part. This is still not the best solution to the problem.

Also fixup the node verification for it to check if node is to be
included and if so, skip verification.

-----
I am not entirely happy with this solution. Proposing for discussion.


cmatsuoka (approved):
I think this is good enough for now, we can refactor it later if we find a better solution.

PR #241, opened 2022-07-04
Add groups dependency support

sergio-costas:
This patch allows to add a 'groups' statement to each part,
define one or more groups for each one, and use the groups for
dependency management.

This way, it is possible to add all the binaries and libraries
parts to a "binaries" group, and make another part deppend on
that group (just using the 'after' statement like with a part
name).

This is mainly useful in big snapcraft.yaml files with a lot
of parts, like the gnome-XX-sdk ones.

This MR replaces https://github.com/snapcore/snapcraft/pull/3823

-----

### DIFF
--- a/craft_parts/parts.py
+++ b/craft_parts/parts.py
@@ -280,6 +280,20 @@ class Part:
         )
 
 
+def get_dependencies_full(part: Part, part_list: List[Part]) -> List[str]:
+    """Return the dependencies, taking into account the groups."""
+    dep_list = []
+    part_names = [p.name for p in part_list]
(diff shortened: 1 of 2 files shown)


sergio-costas:
@sergiusens Ok, moved into craft-parts. Now we need to think a good name for the parameter, instead of "groups".

sergio-costas:
@sergiusens This seems to work fine.

mr-cal:
This is a reasonable idea but I'm closing this PR as stale. If we decide to implement this, we'll start with a new spec.

PR #242, opened 2022-07-04
executor: handle empty scriptlets

cmatsuoka:
Empty scriptlets should be handled as such and not as undefined
scriptlets. In the former case, the built-in handler for the step
is not executed, the empty scriptlet is executed instead.

-----
CRAFT-1182

### DIFF
--- a/craft_parts/executor/part_handler.py
+++ b/craft_parts/executor/part_handler.py
@@ -461,7 +461,7 @@ class PartHandler:
         )
 
         scriptlet = self._part.spec.get_scriptlet(step_info.step)
-        if scriptlet:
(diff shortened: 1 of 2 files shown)


facundobatista (approved):
+1

craft_parts/executor/part_handler.py
```diff
@@ -461,7 +461,7 @@ def _run_step(
         )
 
         scriptlet = self._part.spec.get_scriptlet(step_info.step)
-        if scriptlet:
+        if scriptlet is not None:
```
sergiusens, line 464: classic :-)

sergiusens (approved):

PR #243, opened 2022-07-05
executor: only read xattrs if deb stage packages defined

cmatsuoka:
When deb stage packages are installed xattrs are used to mark each file
so they can be traced back to their original package names. For symmetry,
we read attributes only when using deb packages and stage packages are
defined.

-----


facundobatista (approved):
Looks good!

PR #245, opened 2022-07-07
packages: install unpinned

sergiusens:
Collect installed package versions after the installation.

This allows us to collect the real version whilst avoiding a bug
in apt where it would downgrade or remove a package in cases where
 the pinned version cannot be satisfied.

-----
This would no longer auto mark packages (not useful anymore and --destructive-mode is destructive).

Only the requested packages are passed on to apt install with a best effort of collecting what dependencies would be installed or how virtual packages would be resolved to construct the most backwards compatible result at the end of the process.

CRAFT-1187

### DIFF
--- a/craft_parts/packages/deb.py
+++ b/craft_parts/packages/deb.py
@@ -26,7 +26,7 @@ import subprocess
 import sys
 import tempfile
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


cmatsuoka (approved):
This implementation matches the solution previously discussed.

PR #248, opened 2022-07-18
fix: parts: check for out of source builds

mr-cal:

-----
Plugins that supported out-of-source building were not working with the `source-subdir` parameter.

There were two minor issues:

- A missing check for out-of-source plugins when determining the build directory
- cmake and meson plugins not choosing the right source directory

Unfortunately, accessing the `out_of_source_build` property prior to instantiating the plugin was a challenge.  I've made some comments below.

(CRAFT-1185)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,7 +93,7 @@ test-pydocstyle:
 .PHONY: test-pylint
 test-pylint:
 	pylint craft_parts
-	pylint tests --disable=invalid-name,missing-module-docstring,missing-function-docstring,no-self-use,duplicate-code,protected-access,consider-using-with
+	pylint tests --disable=invalid-name,missing-module-docstring,missing-function-docstring,duplicate-code,protected-access,consider-using-with
 
 .PHONY: test-pyright
 test-pyright:
--- a/craft_parts/parts.py
+++ b/craft_parts/parts.py
@@ -23,6 +23,7 @@ from pydantic import BaseModel, Field, ValidationError, validator
 
 from craft_parts import errors
 from craft_parts.dirs import ProjectDirs
+from craft_parts.plugins import get_plugin_class
 from craft_parts.plugins.properties import PluginProperties
 from craft_parts.steps import Step
 
@@ -165,7 +166,7 @@ class Part:
         plugin_name: str = data.get("plugin", "")
 
         self.name = name
-        self.plugin = plugin_name
+        self.plugin_name = plugin_name
         self.plugin_properties = plugin_properties
         self._dirs = project_dirs
         self._part_dir = project_dirs.parts_dir / name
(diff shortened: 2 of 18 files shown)


Makefile
```diff
@@ -93,7 +93,7 @@ test-pydocstyle:
 .PHONY: test-pylint
 test-pylint:
 	pylint craft_parts
-	pylint tests --disable=invalid-name,missing-module-docstring,missing-function-docstring,no-self-use,duplicate-code,protected-access,consider-using-with
+	pylint tests --disable=invalid-name,missing-module-docstring,missing-function-docstring,duplicate-code,protected-access,consider-using-with
```
mr-cal, line 96: https://pylint.pycqa.org/en/latest/whatsnew/2/2.14/summary.html#removed-checkers

craft_parts/parts.py
```diff
@@ -205,8 +206,16 @@ def part_build_dir(self) -> Path:
...
     @property
     def part_build_subdir(self) -> Path:
-        """Return the subdirectory in build containing the source subtree (if any)."""
-        if self.spec.source_subdir:
+        """Return the subdirectory in build containing the source subtree (if any).
+
+        Parts that have a source subdirectory and do not support out-of-source builds
+        will have a build subdirectory.
+        """
+        if (
```
mr-cal, line 214: This logic hadn't been ported from snapcraft_legacy:
https://github.com/snapcore/snapcraft/blob/e03a971b3c6bff8c1caa4f33502cfb6a8ef401eb/snapcraft_legacy/internal/pluginhandler/__init__.py#L106

craft_parts/parts.py
```diff
@@ -205,8 +206,16 @@ def part_build_dir(self) -> Path:
...
+            self.plugin_name != ""
+            and self.spec.source_subdir
+            and not get_plugin_class(self.plugin_name).get_out_of_source_build()
```
mr-cal, line 217: This is using a new `@classmethod` called `get_out_of_source_build()`.

I couldn't use the existing `out_of_source_build` property, because I would have to create an instance of the plugin.  That gets tricky and circular, as the plugin needs part info, but this code is in the part constructor.

craft_parts/plugins/base.py
```diff
@@ -62,6 +62,11 @@ def get_build_environment(self) -> Dict[str, str]:
 
     @property
     def out_of_source_build(self) -> bool:
```
mr-cal, line 64: Should I remove the `out_of_source_build` property?

The only place it is used across all craft libraries and *craft applications is the `ros` plugin in snapcraft.

craft_parts/parts.py
```diff
@@ -165,7 +166,7 @@ def __init__(
         plugin_name: str = data.get("plugin", "")
 
         self.name = name
-        self.plugin = plugin_name
+        self.plugin_name = plugin_name
```
mr-cal, line 169: Renamed to `plugin_name`, because this variable is just a string, not a plugin!

PR #251, opened 2022-07-26
repo: fix_pkg_config removes prefixes from staged snap's .pc files.

mr-cal:

-----
When normalizing pkg-config (.pc) files from staged snaps, directories previously prepended to the prefix parameter will be removed:

- `/build/<snap-name>/stage` from snaps built via launchpad
-  `/root/stage` from snaps built via a provider

An almost duplicate of https://github.com/snapcore/snapcraft/pull/3848.

LP: #[1916281](https://bugs.launchpad.net/snapcraft/+bug/1916281)
(CRAFT-1215)

### DIFF
--- a/craft_parts/packages/normalize.py
+++ b/craft_parts/packages/normalize.py
@@ -157,34 +157,58 @@ def _try_copy_local(path: Path, target: Path) -> bool:
 
 
 def fix_pkg_config(
-    root: Path,
-    pkg_config_file: Path,
-    prefix_trim: Optional[Path] = None,
+    prefix_prepend: Path, pkg_config_file: Path, prefix_trim: Optional[Path] = None
 ) -> None:
-    """Rewrite the prefix entry in a pkg-config file.
+    """Fix the prefix parameter in pkg-config files.
 
-    :param root: The root to add to the configuration prefix.
-    :param pkg_config_file: The pkg-config file to process.
-    :param prefix_trim: The initial path to remove from the configuration prefix.
+    This function does 3 things:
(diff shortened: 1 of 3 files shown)


craft_parts/packages/normalize.py
```diff
@@ -157,35 +157,59 @@ def _try_copy_local(path: Path, target: Path) -> bool:
...
+    - Built during the build stage: the install directory may be prepended
+
+    :param pkg_config_file: pkg-config (.pc) file to modify
+    :param prefix_prepend: directory to prepend to the prefix
+    :param prefix_trim: directory to remove from prefix
     """
-    # FIXME: see https://bugs.launchpad.net/snapcraft/+bug/1916281
-    pattern_trim = None
+    # build patterns
+    prefixes_to_trim = [r"/build/[\w\-. ]+/stage", "/root/stage"]
```
cmatsuoka, line 180: Can we avoid hard-coding patterns based on specific usage cases, and do something based on our pathname variables instead? E.g. providers use `/root/stage` but this value comes from `info.stage_dir`.

mr-cal: After discussing, these were our conclusions:
1. `/root/stage` and `/build/<snap-name>/stage` are hard-coded into pkg-config files in snaps already in the snap store.
2. Because of this, there is not an ideal way to trim out these prefixes without hard-coding these patterns.
3. A future snapcraft improvement would be to trim out these prefixes during the final packing step (like in the `setup_assets()` function).

PR #257, opened 2022-07-29
sources: format local git paths as file://<path>

mr-cal:

-----

Local git sources are reformatted as `file:///absolute/path/to/local/source`
This is done because `git clone --depth=1 path/to/local/source` is invalid.

LP: #[1979529](https://bugs.launchpad.net/snapcraft/+bug/1979529)
(CRAFT-1153)

### DIFF
--- a/craft_parts/sources/git_source.py
+++ b/craft_parts/sources/git_source.py
@@ -221,7 +221,7 @@ class GitSource(SourceHandler):
         ]
 
         if self.source_submodules is None or len(self.source_submodules) > 0:
-            command.extend(["--recurse-submodules=yes"])
+            command.append("--recurse-submodules=yes")
         self._run(command)
 
         command = [self.command, "-C", self.part_src_dir, "reset", "--hard", refspec]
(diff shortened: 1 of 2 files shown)


craft_parts/sources/git_source.py
```diff
@@ -263,12 +278,21 @@ def _clone_new(self):
...
     def is_local(self) -> bool:
         """Verify whether the git repository is on the local filesystem."""
         return Path(self.part_src_dir, ".git").exists()
 
+    def _is_source_local(self) -> bool:
+        """Verify if source is a filepath.
+
+        :return: True if source is a valid filepath,
+        """
+        pattern = re.compile(r"^[\w\-. \/]+$")
```
cmatsuoka, line 293: Maybe check for absence of a scheme and see if the path is a local directory?

craft_parts/sources/git_source.py
```diff
@@ -252,7 +259,15 @@ def _clone_new(self):
             command.extend(["--branch", self.source_tag or self.source_branch])
         if self.source_depth:
             command.extend(["--depth", str(self.source_depth)])
-        self._run(command + [self.source, self.part_src_dir])
+
+        # reformat local sources
+        if self._is_source_local():
+            command.extend([f"file://{Path(self.source).resolve()}"])
```
cmatsuoka, line 265: We could just append to the list since it's a single element.

craft_parts/sources/git_source.py
mr-cal: Sure, I added the checking for the absence of a protocol and simplified the code a bit more.

I'm on the fence about checking if the local path exists.  First, Git's error handling for non-existent paths is good.  Second, if it's a file path to a networked location, it's going to add a delay.

craft_parts/sources/git_source.py
mr-cal: Changed to `append` here and in a few other spots in the file.

craft_parts/sources/git_source.py
cmatsuoka: Just checking the protocol should be good.

PR #258, opened 2022-08-02
packages: fetch stage-packages from host arch's repo

mr-cal:

-----
`stage-packages` are now fetched from the host architecture's repositories instead of the target architecture's repository.
This was causing an error during cross-compiling because the target architecture's repository was not registered.

LP: #[1983009](https://bugs.launchpad.net/snapcraft/+bug/1983009)
(CRAFT-1216)


facundobatista (approved):
Good, thanks!

cmatsuoka (approved):
This architecture setup system is a bit confusing and is based on the legacy snapcraft implementation. The fix solves the immediate host/target problem but for the mid-term we should consider deeper refactoring and removal of unused fields (now that we have a better idea of the boundaries between craft-parts and applications).

PR #259, opened 2022-08-03
Release 1.10.2

cmatsuoka:
## Changelog

* Fix git source format error when cloning using depth
* Use host architecture when installing stage packages


mr-cal (approved):
Thanks for releasing!

PR #263, opened 2022-08-16
plugins: fix cmake plugin prefix

cmatsuoka:
Properly format the cmake prefix path string, and set the installation
prefix to match v1 plugins.

-----
CRAFT-1154

### DIFF
--- a/craft_parts/plugins/cmake_plugin.py
+++ b/craft_parts/plugins/cmake_plugin.py
@@ -89,7 +89,8 @@ class CMakePlugin(Plugin):
     def get_build_environment(self) -> Dict[str, str]:
         """Return a dictionary with the environment to use in the build step."""
         return {
-            "CMAKE_PREFIX_PATH": "{self._part_info.stage_dir}",
+            # Also look for staged headers and libraries.
+            "CMAKE_PREFIX_PATH": str(self._part_info.stage_dir)
         }
 
     def get_build_commands(self) -> List[str]:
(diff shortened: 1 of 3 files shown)


craft_parts/plugins/cmake_plugin.py
```diff
@@ -88,9 +88,7 @@ def get_build_packages(self) -> Set[str]:
 
     def get_build_environment(self) -> Dict[str, str]:
         """Return a dictionary with the environment to use in the build step."""
-        return {
-            "CMAKE_PREFIX_PATH": "{self._part_info.stage_dir}",
-        }
+        return {"CMAKE_PREFIX_PATH": f"{self._part_info.stage_dir}"}
```
sergiusens, line 91: not related to the PR, but why is this set to stage_dir?

sergiusens, line 91: Side question, is this a better alternative?
```suggestion
        return {"CMAKE_PREFIX_PATH": str(self._part_info.stage_dir)}
```

craft_parts/plugins/cmake_plugin.py
```diff
@@ -101,6 +99,7 @@ def get_build_commands(self) -> List[str]:
             f'"{self._part_info.part_src_subdir}"',
             "-G",
             f'"{options.cmake_generator}"',
+            "-DCMAKE_INSTALL_PREFIX=",
```
sergiusens, line 102: please add a code comment on the effects or why this is here.

craft_parts/plugins/cmake_plugin.py
cmatsuoka: This allows cmake to find staged headers and libs. The regular search paths for system headers and libraries are still used, this only adds staged locations to the list.

cmatsuoka: It is, using that instead.

craft_parts/plugins/cmake_plugin.py
cmatsuoka: Comment added for this variable and for `CMAKE_PREFIX_PATH` above.

cmatsuoka:
Amended and rebased on main.

mr-cal (approved):
Looks good.  I tested with a project that previously had 
```
cmake-parameters:
  - -DCMAKE_INSTALL_PREFIX=/usr/
  - -DCMAKE_PREFIX_PATH="${SNAPCRAFT_STAGE}/usr"
```
in a few of the parts.  Using this PR, that snap builds with or without those extra parameters.

sergiusens (approved):

PR #267, opened 2022-08-18
plugins: drop cmake install prefix argument

cmatsuoka:
Setting `CMAKE_INSTALL_PREFIX` by default makes the cmake plugin
compatible with the Snapcraft cmake plugin for core18 and solves
usability issues, but it also breaks compatibility with existing
core22 snaps that expect binaries to be installed in `usr/local/bin`
by default.

To keep compatibility with existing core22 snaps, the prefix path
will not be defined by default, and snap authors are encouraged
to define `-DCMAKE_INSTALL_PREFIX=` in cmake parameters to be
compatible with the core18 cmake plugin, so that libraries
built with cmake and staged by a different part can be correctly
detected without the need to define additional parameters such as
`CMAKE_INCLUDE_PATH` and `CMAKE_LIBRARY_PATH`.

-----

### DIFF
--- a/craft_parts/plugins/cmake_plugin.py
+++ b/craft_parts/plugins/cmake_plugin.py
@@ -57,7 +57,16 @@ class CMakePlugin(Plugin):
     For more information check the 'plugins' topic for the former and the
     'sources' topic for the latter.
 
-    Additionally, this plugin uses the following plugin-specific keywords:
+    This implementation follows the syntax and behavior used in the
+    Snapcraft cmake plugin for core20. Unlike the cmake plugin used for
+    core18, ``CMAKE_INSTALL_PREFIX`` is not automatically set. To retain
+    compatibility with the Snapcraft core18 plugin, define the cmake
+    parameter ``-DCMAKE_INSTALL_PREFIX=`` in your project.  This also
+    allows libraries built using the cmake plugin and staged by a different
+    part to be automatically recognized without defining additional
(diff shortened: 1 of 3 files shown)


cmatsuoka:
This should address issues discussed in snapcore/snapcraft#3878.

craft_parts/plugins/cmake_plugin.py
```diff
@@ -102,11 +111,6 @@ def get_build_commands(self) -> List[str]:
             f'"{self._part_info.part_src_subdir}"',
             "-G",
             f'"{options.cmake_generator}"',
-            # Install on a location we search when building using staged files
-            # (e.g. CMAKE_PREFIX_PATH/lib, CMAKE_PREFIX_PATH/lib/<arch> for libs,
-            # see https://cmake.org/cmake/help/latest/command/find_library.html).
-            # This can be overridden by the user using cmake-parameters.
-            "-DCMAKE_INSTALL_PREFIX=",
         ] + options.cmake_parameters
```
sergiusens, line 114: we could detect that -DCMAKE_INSTALL_PREFIX is not defined an emit a warning here

cmatsuoka: This parameter is useful when building libraries that will be staged and used by another part that also uses the cmake plugin. This is a very specific case and the majority of use cases of the cmake plugin won't need it, but the warning would be emited anyway leading to some confusion as to when the parameter is really needed. Using external documentation in this case would allow us to explain the scenario in more detail.

sergiusens (approved):

PR #269, opened 2022-08-24
sources: support "deb" sources

tigarmo:

Unpack deb packages in the same way that the stage-packages DebPackage
does (via dpkg-deb). This motivates a small refactoring to move
"extract_deb()" to a place that both handlers can use.

A lot of the code on this PR comes from a work-in-progress branch by @cmatsuoka . Thanks!

-----
CRAFT-117

### DIFF
--- a/craft_parts/errors.py
+++ b/craft_parts/errors.py
@@ -538,3 +538,16 @@ class OverlayPermissionError(PartsError):
         brief = "Using the overlay step requires superuser privileges."
 
         super().__init__(brief=brief)
+
+
+class DebError(PartsError):
+    """A "deb"-related command failed."""
+
+    def __init__(self, deb_path, command, exit_code):
+        brief = (
+            f"Failed when handling {deb_path}: "
+            f"command {command!r} exited with code {exit_code}."
+        )
+        resolution = "Make sure the deb file is correctly specified."
+
+        super().__init__(brief=brief, resolution=resolution)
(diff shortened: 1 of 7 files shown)


craft_parts/errors.py
```diff
@@ -538,3 +538,14 @@ def __init__(self) -> None:
...
 
         super().__init__(brief=brief)
+
+
+class DebError(PartsError):
+    """A "deb"-related command failed."""
+
+    def __init__(self, deb_path, command, exit_code):
+        brief = f"Failed when handling {deb_path}: "
+        brief += f"command {command!r} exited with code {exit_code}."
```
cmatsuoka, line 548: Style note: we usually format long strings as
```
foo = (
   "something very long ..."
   "... continued here"
)
```
Instead of using `+=`

craft_parts/utils/deb_utils.py
```diff
@@ -0,0 +1,38 @@
...
+
+import subprocess
+from pathlib import Path
+from typing import Callable
+
+from craft_parts import errors
+from craft_parts.utils import os_utils
+
+
+def extract_deb(
```
cmatsuoka, line 27: I wouldn´t mind this to be part to the deb source module, I don't see components other than the deb source handler and the deb package handler using deb extraction, and package handling may be spun off as a separate project in the future.

tigarmo: That would mean that `packages.deb` would import `sources.deb_source`. Is that ok?

sergiusens: I think that having deb_utils is fine

cmatsuoka (approved):
Thanks!

PR #271, opened 2022-08-29
plugins: support "go generate" in go parts

tigarmo:
plugins: support "go generate" in go parts
    
GoPlugin's build step will now call "go generate" before the actual
build/install to support code generation. This shouldn't affect existing
uses of the plugin because the call is harmless if there's no code to
generate.

-----

CRAFT-1316

### DIFF
--- a/tests/integration/plugins/test_go.py
+++ b/tests/integration/plugins/test_go.py
@@ -72,3 +72,37 @@ def test_go_plugin(new_dir, mocker):
 
     output = subprocess.check_output([str(binary)], text=True)
     assert output == "I can eat glass and it doesn't hurt me."
+
+
+def test_go_generate(new_dir):
+    """Test code generation via "go generate" in parts using the go plugin
+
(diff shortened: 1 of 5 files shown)


tests/integration/plugins/test_go.py
```diff
@@ -72,3 +72,37 @@ def test_go_plugin(new_dir, mocker):
...
     assert output == "I can eat glass and it doesn't hurt me."
+
+
+def test_go_generate(new_dir):
+    """Test code generation via "go generate" in parts using the go plugin
+
+    The go code in the "test_go" dir uses "gen/generator.go" to create, at build time,
+    the "main.go" file that produces the final binary.
+    """
+    source_location = Path(__file__).parent / "test_go"
```
tigarmo, line 83: _compromise_

tigarmo:
@cmatsuoka @sergiusens I removed `datadir` so this should be ready for review. thanksss

sergiusens (approved):

PR #272, opened 2022-08-30
tests: make integration tests use datadir

tigarmo:
The `pytest-datadir` plugin enable per-test-module data, that is exposed
to tests via the `datadir` fixture. This proof-of-concept commit
refactors the existing integration tests for the rust plugin to use this
scheme.

The main change is that `test_rust.py` is simplified because all the
runtime rust code generation is moved to data diles in the `test_rust/`
dir; this is useful because these rust files are static, and their
generation is not what's being tested.

-----

### DIFF
--- a/tests/integration/plugins/test_rust.py
+++ b/tests/integration/plugins/test_rust.py
@@ -23,38 +23,16 @@ import yaml
 from craft_parts import LifecycleManager, Step
 
 
-def test_rust_plugin(new_dir):
+def test_rust_plugin(new_dir, datadir):
     parts_yaml = textwrap.dedent(
         """\
         parts:
           foo:
             plugin: rust
-            source: .
+            source: test_rust/simple
         """
     )
     parts = yaml.safe_load(parts_yaml)
-
-    Path("Cargo.toml").write_text(
-        textwrap.dedent(
-            """\
-            [package]
-            name = "rust-hello"
-            version = "1.0.0"
-            edition = "2021"
-            """
-        )
-    )
-
-    Path("src").mkdir()
-    Path("src/main.rs").write_text(
-        textwrap.dedent(
-            """\
-            fn main() {
-                println!("hello world");
-            }
-            """
-        )
-    )
 
     lifecycle = LifecycleManager(
         parts, application_name="test_rust_plugin", cache_dir=new_dir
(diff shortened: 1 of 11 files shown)


tests/integration/plugins/test_rust.py
```diff
@@ -29,33 +29,11 @@ def test_rust_plugin(new_dir):
         parts:
           foo:
             plugin: rust
-            source: .
+            source: simple
         """
     )
     parts = yaml.safe_load(parts_yaml)
 
-    Path("Cargo.toml").write_text(
```
tigarmo, line 37: all the files in the `test_rust/` dir have the contents that were removed from these tests, verbatim; to generate them I ran the original tests and then just copied the contents from the temp dir.

tests/integration/plugins/test_rust.py
```diff
@@ -127,80 +74,18 @@ def test_rust_plugin_features(new_dir):
...
 
-def test_rust_plugin_workspace(new_dir):
+def test_rust_plugin_workspace(new_dir, datadir):
     parts_yaml = textwrap.dedent(
         """\
         parts:
           foo:
             plugin: rust
-            source: .
+            source: test_rust/workspace
```
facundobatista, line 83: I like for this to be simpler than before... but what's the relation between `datadir` and this directory?

Does `datadir` change the current working directory? in the documentation the examples do use `datadir` as the path base, why not here?

tests/integration/plugins/test_rust.py
```diff
@@ -127,80 +74,18 @@ def test_rust_plugin_features(new_dir):
...
             rust-path: ["hello"]
         """
     )
     parts = yaml.safe_load(parts_yaml)
 
-    Path("Cargo.toml").write_text(
```
facundobatista, line 142: Did these tests create stuff in the user's working directory? Ugly.

tests/integration/plugins/test_rust.py
tigarmo: (answering both comments on this one)
`datadir` doesn't change the cwd, it creates a dir inside the test's temp dir. 
The tests use a `new_dir` fixture that changes the cwd to `tmpdir`:

```python
@pytest.fixture
def new_dir(tmpdir):
    """Change to a new temporary directory."""

    cwd = os.getcwd()
    os.chdir(tmpdir)

    yield tmpdir

    os.chdir(cwd)
```

... so the tests that do `Path("bla").write_text(...)` are actually writing into the temp dir.
I'm not 100% sure why, but I think the code in `craft_parts` is set up to work off of the cwd (possibly to mimic the behavior of the command line, like running `snapcraft` on a directory that has a `snapcraft.yaml` file.

tigarmo: For comparison, here's how the test might look like with only `datadir`:

```python
def test_rust_plugin(datadir):
    parts_yaml = textwrap.dedent(
        f"""\
        parts:
          foo:
            plugin: rust
            source: {datadir}/simple
        """
    )

    lifecycle = LifecycleManager(
        parts, application_name="test_rust_plugin", cache_dir=datadir, work_dir=datadir
    )
    actions = lifecycle.plan(Step.PRIME)

    with lifecycle.action_executor() as ctx:
        ctx.execute(actions)

    binary = Path(lifecycle.project_info.prime_dir, "bin", "rust-hello")

    output = subprocess.check_output([str(binary)], text=True)
    assert output == "hello world\n"
```

there are more explicit uses of `datadir` in the yaml and in the `LifecycleManager` constructor. While I personally prefer this approach I think it's mostly subjective, and it would be considerably different than what's currently used (it wasn't my intention to change the tests *too* much).

tests/integration/plugins/test_rust.py
sergiusens: no, new_dir changes cwd to a tmp directory

tests/integration/plugins/test_rust.py
sergiusens: why don't we flatten the tree? can the data_dir fixture operate at the method level instead of the module level?

tigarmo: > can the data_dir fixture operate at the method level instead of the module level?

No, it works at the module level. So if the module has many tests that want to write the same file (like `Cargo.toml` in this case) you end up having to nest directories inside the datadir

sergiusens: the amount of data written here is not much, but it could lead to a burden if the dataset to copy gets large for one test.

tigarmo:
Thanks everyone for the review. Closing this PR and opening issue #275 to have a proper discussion on the topic.

PR #279, opened 2022-09-13
spec: add "stage-slices" for chisel slices

tigarmo:
The slices are defined in the regular "stage-packages" by virtue of the fact that .deb package names *cannot* have underscores while chisel slices *must always* have an underscore to distinguish between package name and the specific slice.

-----

(CRAFT-1329)

### DIFF
--- a/craft_parts/parts.py
+++ b/craft_parts/parts.py
@@ -19,10 +19,11 @@
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Sequence, Set
 
-from pydantic import BaseModel, Field, ValidationError, validator
+from pydantic import BaseModel, Field, ValidationError, root_validator, validator
 
 from craft_parts import errors, plugins
 from craft_parts.dirs import ProjectDirs
+from craft_parts.packages import platform
 from craft_parts.plugins.properties import PluginProperties
 from craft_parts.steps import Step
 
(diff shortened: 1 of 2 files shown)


craft_parts/parts.py
```diff
@@ -77,6 +78,31 @@ def validate_relative_path_list(cls, item):
         ), f"{item!r} must be a relative path (cannot start with '/')"
         return item
 
+    @root_validator(pre=True)
+    def validate_root(cls, values):
+        """Check if the part spec has a valid configuration of packages and slices."""
+        assert "stage-slices" not in values, (
+            "'stage-slices' must not be explicitly passed in the part data; "
+            "it is derived from 'stage-packages'"
+        )
```
cmatsuoka, line 87: It's better to keep the `PartSpec` model reflecting what the user can actually define instead of creating a phantom field as this could cause unexpected behavior when marshaling the part data again. We could use a simple validator for `stage_packages` to make sure it contains only one type of package or other and install slices from `stage_packages` later instead of `stage_slices` if we detect it's using slices.

cmatsuoka (changes requested):
Slices are still packages so we can decide later what to do based on the kind of packages they are, instead of doing it at input validation time.

craft_parts/parts.py
tigarmo: I see. I actually agree with you with maintaining the "purity" of the `PartSpec`, and I considered keeping everything in `stage_packages`. The reason that I chose this path is that if we keep everything in `stage_packages` then every bit of code that references `stage_packages` today will have to be updated because that list might contain slices instead of .deb packages.

So I guess my question is, is that acceptable? Is there code in other repos (like the *craft applications) that manipulate `stage_packages`?

cmatsuoka: The idea is that `stage_packages` could be used regardless of actual content throughout the code except for type-specific code in the `packages` package. Since the package type is set for the entire project and slices can be used per-part (or so it seems), we could handle that in `packages.deb` and (hopefully) leave the rest of the code unchanged.

cmatsuoka (approved):
Thanks!

PR #281, opened 2022-09-21
unpack chisel slices

tigarmo:
This is done by using the `PullState` to "transmit" the list of slices from the pull step (that currently doesn't do anything for slices) to the build step (that does the actual cutting).

The integration test needs the "chisel" binary, so it's set to run on CI only. This required updating the workflow to use a newer version of "go" to build chisel. I then needed to revert back to the default Ubuntu 20.04 version because other tests need an older version (might be worth looking into).

In addition to the tests I also built a local `rockcraft` snap using this code, and used it to build an image containing "sliced" files.

### DIFF
--- a/tests/integration/lifecycle/test_chisel_lifecycle.py
+++ b/tests/integration/lifecycle/test_chisel_lifecycle.py
@@ -1,0 +1,56 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2022 Canonical Ltd.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License version 3 as published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+import os
+import textwrap
+from pathlib import Path
+
+import pytest
+import yaml
+
+import craft_parts
+from craft_parts import Step
+
+IS_CI: bool = os.getenv("CI") == "true"
+
+
+@pytest.mark.skipif(not IS_CI, reason="This test needs 'chisel' and only runs on CI.")
+def test_chisel_lifecycle(new_dir):
+    """Integrated test for Chisel support.
+
+    Note that since this test needs the "chisel" binary, it currently only runs on CI.
+    """
+    _parts_yaml = textwrap.dedent(
+        """\
+        parts:
+          foo:
+            plugin: nil
+            stage-packages: [ca-certificates_data]
+        """
+    )
+
+    parts = yaml.safe_load(_parts_yaml)
+
+    lf = craft_parts.LifecycleManager(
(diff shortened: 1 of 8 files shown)


tests/integration/lifecycle/test_chisel_lifecycle.py
```diff
@@ -0,0 +1,49 @@
...
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
```
cmatsuoka, line 14: Craft-parts files should be LGPL instead of GPL.

tests/unit/packages/conftest.py
```diff
@@ -0,0 +1,44 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2022 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
```
cmatsuoka, line 6: Craft-parts files should be LGPL instead of GPL.

tests/unit/packages/test_chisel.py
```diff
@@ -0,0 +1,116 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2022 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
```
cmatsuoka, line 6: Craft-parts files should be LGPL instead of GPL.

tests/integration/lifecycle/test_chisel_lifecycle.py
```diff
@@ -0,0 +1,49 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2022 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
```
cmatsuoka, line 6: Craft-parts files should be LGPL instead of GPL.

tests/unit/packages/conftest.py
tigarmo: thanks that's a good catch. I think I copied this over from some other craft-parts file, so I'll look into fixing that in another PR (will fix these new headers here though)

sergiusens (approved):

tests/integration/lifecycle/test_chisel_lifecycle.py
```diff
@@ -0,0 +1,56 @@
...
+import textwrap
+from pathlib import Path
+
+import pytest
+import yaml
+
+import craft_parts
+from craft_parts import Step
+
+IS_CI: bool = os.getenv("CI") == "true"
```
cmatsuoka, line 26: TIL about the `CI` variable.

cmatsuoka (approved):

tests/integration/lifecycle/test_chisel_lifecycle.py
tigarmo: yeah I got it from here: https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables

PR #287, opened 2022-09-23
plugins: add kernel plugin implementation

kubiko:
Adding kernel plugin to build kernel snaps from linux kernel source.

Kernel snap is a specific snap for Ubuntu Core which contains kernel binary, firmware, kernel modules, dtbs, initrd
There are various options usually involved, initrd has to be packaged in a specific way and have certain content.
At the same time, many of the steps are common to all the kernel snaps builds. This plugin is trying to streamline this as make kernel snap build easier.

### DIFF
--- a/craft_parts/plugins/kernel_plugin.py
+++ b/craft_parts/plugins/kernel_plugin.py
@@ -1,0 +1,1653 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+# pylint: disable=line-too-long,missing-function-docstring,import-error,too-many-lines,missing-class-docstring,attribute-defined-outside-init,too-many-instance-attributes
+#
+# Copyright 2020 Canonical Ltd.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License version 3 as published by the Free Software Foundation.
+#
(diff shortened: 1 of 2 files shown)


cmatsuoka:
The kernel plugin seems to be very snap-specific, so we can implement it in Snapcraft as an application plugin instead of adding it directly to craft-parts.

kubiko:
> The kernel plugin seems to be very snap-specific, so we can implement it in Snapcraft as an application plugin instead of adding it directly to craft-parts.

It is indeed very snap-specific. It's a bit confusing for me where snapcraft plugins go at this stage, but since I found most plugins in craft-parts, it was my first shot where to add it :)

cmatsuoka:
> It is indeed very snap-specific. It's a bit confusing for me where snapcraft plugins go at this stage, but since I found most plugins in craft-parts, it was my first shot where to add it :)

See how the colcon and conda plugins are defined in `snapcraft/parts/plugins`, we can define this one there too.

facundobatista:
I concur, this looks more like a Snapcraft plugin that a generic Craft Parts one.

kubiko:
Closing this PR and adding kernel plugin as snapcraft plugin to the existing PR.
Please review it there.
https://github.com/snapcore/snapcraft/pull/3922

PR #295, opened 2022-10-12
chroot: fix device nodes in overlay base image

cmatsuoka:
Device nodes won't appear correctly in a regular bind mount inside a container, so use rbind instead.

-----
CRAFT-1416

### DIFF
--- a/craft_parts/overlays/chroot.py
+++ b/craft_parts/overlays/chroot.py
@@ -47,12 +47,14 @@ def chroot(path: Path, target: Callable, *args, **kwargs) -> Any:
     child = multiprocessing.Process(
         target=_runner, args=(Path(path), child_conn, target, args, kwargs)
     )
+    logger.debug("[pid=%d] set up chroot", os.getpid())
     _setup_chroot(path)
     try:
         child.start()
         res, err = parent_conn.recv()
         child.join()
     finally:
+        logger.debug("[pid=%d] clean up chroot", os.getpid())
         _cleanup_chroot(path)
 
     if isinstance(err, str):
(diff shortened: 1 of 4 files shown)


craft_parts/overlays/chroot.py
```diff
@@ -150,4 +153,7 @@ def _cleanup_chroot_linux(path: Path) -> None:
 
         if mountpoint.exists():
             logger.debug("[pid=%d] umount: %r", pid, str(mountpoint))
-            os_utils.umount(str(mountpoint))
+            if entry.option == "--rbind":
+                os_utils.umount(str(mountpoint), "--recursive", "--lazy")
```
sergiusens, line 157: why is lazy required?

cmatsuoka: It covers the case where a filesystem under /dev is busy (e.g. /dev/shm being used), in which case /dev becomes unmountable. This shouldn't be the case under normal conditions when using a build instance, however it may happen in destructive mode in certain environments such as the github test machines. In practice it sets the `MNT_DETACH` flag in umount(2) to defer actual unmounting, and should be safe to use in our scenario.

cmatsuoka: Adding a comment to the source code.

craft_parts/overlays/chroot.py
```diff
@@ -104,8 +106,8 @@ def _cleanup_chroot(path: Path) -> None:
     _Mount(None, "/etc/resolv.conf", "/etc/resolv.conf", "--bind"),
     _Mount("proc", "proc", "/proc", None),
     _Mount("sysfs", "sysfs", "/sys", None),
-    _Mount(None, "/dev", "/dev", "--bind"),
-    _Mount("tmpfs", "tmpfs", "/dev/shm", None),
```
tigarmo, line 108: how come the `tmpfs` mount was removed?

tigarmo: oh is it because its mounted under `/dev`?

tigarmo (approved):
nice job debugging this!

craft_parts/overlays/chroot.py
cmatsuoka: Yes, it's mounted under /dev by `--rbind`, and removed with `--recursive`.

PR #296, opened 2022-10-13
parts: add Permissions model

tigarmo:
Add the base `Permissions` model with basic API. No actual integration into the lifecycle yet (that comes next).

-----

### DIFF
--- a/craft_parts/permissions.py
+++ b/craft_parts/permissions.py
@@ -1,0 +1,102 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2022 Canonical Ltd.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License version 3 as published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Specify and apply permissions and ownership to part-owned files."""
+
+import os
+from fnmatch import fnmatch
+from pathlib import Path
+from typing import List, Optional, Union
+
+from pydantic import BaseModel, root_validator
+
+
+class Permissions(BaseModel):
+    """Description of the ownership and permission settings for a set of files.
+
+    A `Permissions` object specifies that a given pattern-like `path` should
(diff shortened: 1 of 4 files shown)


craft_parts/permissions.py
```diff
@@ -0,0 +1,102 @@
...
+      - `path` is optional and defaults to "everything";
+      - `owner` and `group` are optional if both are omitted - that is, if
+        one of the pair is specified then both must be;
+      - `mode` is a string containing an integer in base 8. For example, "755",
+      "0755" and "0o755" are all accepted and are the equivalent of calling
+      `chmod 755 ...`.
+
+    """
+
+    path: str = "*"
```
cmatsuoka, line 45: Should the default path be `*`, or `**/*`? The latter case would match every file regardless of directory.

craft_parts/permissions.py
```diff
@@ -0,0 +1,102 @@
...
+
+def filter_permissions(
+    target: Union[Path, str], permissions: List[Permissions]
+) -> List[Permissions]:
+    """Get the subset of `permissions` whose path patterns apply to `target`."""
+    return [p for p in permissions if p.applies_to(target)]
+
+
+def apply_permissions(target: Union[Path, str], permissions: List[Permissions]) -> None:
+    """Apply all permissions configurations in `permissions` to `target`."""
```
cmatsuoka, line 100: To be properly formatted by Sphinx I believe we should use double backticks.

cmatsuoka (changes requested):
We should also take permission and ownership into account when checking for stage file collision, if a file is present in more than one part, and more than one part sets permissions for that file. If just one part sets permissions we can assume that the default permission set by other parts are to be overriden. See `craft_parts.executor.check_for_stage_collisions()`.

tigarmo:
That's a great point, actually. This first PR doesn't deal with the actual integration with the lifecycle but I added your comment on that other issue (that will be implemented next)

craft_parts/permissions.py
tigarmo: So the reason that I used `"*"` is because I'm using `fnmatch()` instead of, for example, `glob()`. `fnmatch()` works with strings as a "simplified" regex, so `"*"` matches *everything*.

Now, the reason that I'm using `fnmatch()` is because I'm thinking that this will be integrated into `migrate_files()`, which already receives the full set of files/paths to copy. So, while `glob()` takes a pattern and gives you the filenames that would match it, I kind of need the opposite: given a filepath, does this pattern match it? That's why I used `fnmatch()`.

Do you have a good idea for a better way to do this? How do we do this elsewhere?

cmatsuoka: Ah, I was unaware that fnmatch would match it that way. Since just `*` sets the default behaviour we want it should be fine.

cmatsuoka (approved):
We should check at some point if there's a better alternative to `fnmatch()` that takes the difference between toplevel files and all files including subdirs into account and use that to avoid surprising/unexpected behavior.

PR #297, opened 2022-10-13
overlay: only refresh overlay packages if necessary

tigarmo:
Only mount & refresh the overlay package cache if the part has overlay-packages; otherwise we're doing extra expensive work needlessly.

-----

### DIFF
--- a/craft_parts/executor/executor.py
+++ b/craft_parts/executor/executor.py
@@ -91,7 +91,7 @@ class Executor:
 
         # update the overlay environment package list to allow installation of
         # overlay packages.
-        if any(p.has_overlay for p in self._part_list):
+        if any(p.spec.overlay_packages for p in self._part_list):
             with overlays.PackageCacheMount(self._overlay_manager) as ctx:
                 ctx.refresh_packages_list()
 
--- a/tests/integration/lifecycle/test_overlay.py
+++ b/tests/integration/lifecycle/test_overlay.py
@@ -35,7 +35,13 @@ def fake_call(mocker):
 def mock_overlay_support_prerequisites(mocker):
     mocker.patch.object(sys, "platform", "linux")
     mocker.patch("os.geteuid", return_value=0)
-    mocker.patch("craft_parts.overlays.OverlayManager.refresh_packages_list")
+    mock_refresh = mocker.patch(
+        "craft_parts.overlays.OverlayManager.refresh_packages_list"
+    )
+    yield
+    # Make sure that refresh_packages_list() was *not* called, as it's an expensive call that
+    # overlays without packages do not need.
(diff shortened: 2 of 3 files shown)


craft_parts/executor/executor.py
```diff
@@ -91,7 +91,7 @@ def prologue(self) -> None:
 
         # update the overlay environment package list to allow installation of
         # overlay packages.
-        if any(p.has_overlay for p in self._part_list):
+        if any(p.spec.overlay_packages for p in self._part_list):
```
tigarmo, line 94: not sure if this is an 'overreach' from the executor; it might be preferrable to add a `has_overlay_packages` method/property in the Part (like `has_overlay`)

tests/integration/lifecycle/test_overlay.py
```diff
@@ -35,7 +35,13 @@ def fake_call(mocker):
...
     mocker.patch.object(sys, "platform", "linux")
     mocker.patch("os.geteuid", return_value=0)
-    mocker.patch("craft_parts.overlays.OverlayManager.refresh_packages_list")
+    mock_refresh = mocker.patch(
+        "craft_parts.overlays.OverlayManager.refresh_packages_list"
+    )
+    yield
+    # Make sure that refresh_packages_list() was *not* called, as it's an expensive call that
+    # overlays without packages do not need.
+    assert not mock_refresh.called
```
tigarmo, line 44: this one is just to be double-extra-sure, heh

craft_parts/executor/executor.py
cmatsuoka: That should be fine for now.

tests/integration/lifecycle/test_overlay.py
sergiusens: heh, negative assertions can become eternal

tigarmo: I think I know what you mean, but can you elaborate? Should I remove this assertions?

sergiusens: sorry, it was a side comment :-)

sergiusens (approved):

craft_parts/executor/executor.py
```diff
@@ -91,7 +91,7 @@ def prologue(self) -> None:
 
         # update the overlay environment package list to allow installation of
         # overlay packages.
-        if any(p.has_overlay for p in self._part_list):
```
sergiusens, line 94: if p.has_overlay is unused, it might be worth removing

tigarmo: it does have other uses (unrelated to packages)

PR #303, opened 2022-10-18
executor: separate environment setup from build script

mr-cal:

-----
Separate environment setup from the build script.  Used by https://github.com/snapcore/snapcraft/pull/3951.

~TODO - add unit tests~

(CRAFT-1411)

### DIFF
--- a/craft_parts/executor/step_handler.py
+++ b/craft_parts/executor/step_handler.py
@@ -119,16 +119,26 @@ class StepHandler:
         # Plugin commands.
         plugin_build_commands = self._plugin.get_build_commands()
 
-        # Save script to execute.
+        # save script to set the build environment
+        build_environment_script_path = (
+            self._part.part_run_dir.absolute() / "environment.sh"
+        )
+        build_environment_script_path.write_text(self._env)
+        build_environment_script_path.chmod(0o644)
+
+        # save script to execute the build commands
         build_script_path = self._part.part_run_dir.absolute() / "build.sh"
         with build_script_path.open("w") as run_file:
-            print(self._env, file=run_file)
+            print("#!/bin/bash", file=run_file)
+            print("set -euo pipefail", file=run_file)
+            print(f"source {build_environment_script_path}", file=run_file)
             print("set -x", file=run_file)
 
             for build_command in plugin_build_commands:
                 print(build_command, file=run_file)
 
         build_script_path.chmod(0o755)
+        logger.debug("Executing %r", build_script_path)
 
         try:
             subprocess.run(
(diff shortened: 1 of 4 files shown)


craft_parts/executor/step_handler.py
```diff
@@ -119,16 +119,28 @@ def _builtin_build(self) -> StepContents:
         # Plugin commands.
         plugin_build_commands = self._plugin.get_build_commands()
 
-        # Save script to execute.
+        # save script to set the build environment
+        build_environment_script_path = (
+            self._part.part_run_dir.absolute() / "environment.sh"
+        )
+        with build_environment_script_path.open("w") as run_file:
+            print(self._env, file=run_file)
```
cmatsuoka, line 127: We could simplify this as `build_environment_script_path.write_text(self._env)`

tests/unit/executor/test_step_handler.py
```diff
@@ -95,9 +102,51 @@ def test_run_builtin_build(self, new_dir, mocker):
...
+        build_script_path = Path(new_dir / "parts/p1/run/build.sh")
+        environment_script_path = Path(new_dir / "parts/p1/run/environment.sh")
+
+        assert get_mode(environment_script_path) == 0o644
+        with open(environment_script_path, "r") as file:
+            assert file.read() == dedent(
+                f"""\
+                #!/bin/bash
+                set -euo pipefail
+                # Environment
```
cmatsuoka, line 114: Now that this resides in a file of its own that's sourced from a different script we could remove these lines and leave only variable exporting, WDYT?

tests/unit/executor/test_step_handler.py
```diff
@@ -95,9 +102,51 @@ def test_run_builtin_build(self, new_dir, mocker):
...
+                ## User environment
+                """
+            )
+
+        assert get_mode(build_script_path) == 0o755
+        with open(build_script_path, "r") as file:
+            assert file.read() == dedent(
+                f"""\
+                #!/bin/bash
+                set -euo pipefail
```
cmatsuoka, line 141: The lines are already in the main script, and the script that only set variables doesn't need an interpreter because it would be sourced anyway.

tests/unit/executor/test_step_handler.py
mr-cal: Oops, I didn't post my comment from yesterday.
Yes, I agreer.  I don't think there's any need for the first 2 lines anymore.

This appears to work for both the builtin `build` and for build scriplets.

cmatsuoka (approved):
Thanks!

PR #304, opened 2022-10-18
migration: support Permissions when migrating files

tigarmo:
The migration of part files from 'stage' to 'prime' now takes into account the part's 'permissions' definition. This means forwarding the permissions lists all the way down to 'link_or_copy()'.

-----

(CRAFT-1442)


cmatsuoka (approved):
Great work, this covers what's needed to migrate permissions along with files.

PR #305, opened 2022-10-19
executor: take permissions into account for collisions

tigarmo:
Two paths are considered to be colliding if they both have permissions definitions. The definitions are not "evaluated" to see if they possibly match (that is, if both sets of definitions would generate the same result) for simplicity, but this can be improved in the future.

-----

### DIFF
--- a/craft_parts/executor/collisions.py
+++ b/craft_parts/executor/collisions.py
@@ -24,7 +24,7 @@ from craft_parts import errors, permissions
 from craft_parts.executor import filesets
 from craft_parts.executor.filesets import Fileset
 from craft_parts.parts import Part
-from craft_parts.permissions import Permissions
+from craft_parts.permissions import Permissions, permissions_are_compatible
 
 
 def check_for_stage_collisions(part_list: List[Part]) -> None:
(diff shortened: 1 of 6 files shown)


craft_parts/executor/collisions.py
```diff
@@ -100,8 +123,8 @@ def paths_collide(path1: str, path2: str) -> bool:
     if not (path1_is_dir and path2_is_dir) and _file_collides(path1, path2):
         return True
 
-    # Otherwise, paths do not conflict.
-    return False
+    # Otherwise, paths only conflict if both have permissions defined for them.
+    return all([permissions_path1, permissions_path2])
```
cmatsuoka, line 127: Shouldn't they conflict if the permissions are actually different?

tigarmo: As we discussed, I pushed a commit that does some comparing logic to have paths collide if their permissions are "different". The code is probably unorganized and overcomplicated because my brain is a bit fried, but please take a look. Maybe you have a cleaner approach.
The failing test is, as far as I can tell, unrelated to this PR (a `git` update broke a test)

cmatsuoka (approved):
Thanks!

PR #306, opened 2022-10-19
many: make changed file list available to plugins

cmatsuoka:
Part plugins can make better decisions on how to handle part building if they know what has been changed instead of relying solely on the current state of the filesystem and on static plugin properties defined in the project. For example, retrieving plugin-specific dependency packages could be avoided entirely if no changes were made to the files listing these dependencies, resulting in faster execution times and preventing unnecessary network access.

-----

### DIFF
--- a/craft_parts/plugins/base.py
+++ b/craft_parts/plugins/base.py
@@ -17,9 +17,12 @@
 """Plugin base class and definitions."""
 
 import abc
+from copy import deepcopy
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Type
 
 from pydantic import BaseModel
+
+from craft_parts.actions import ActionProperties
 
 from .properties import PluginProperties
 from .validator import PluginEnvironmentValidator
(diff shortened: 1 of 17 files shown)


craft_parts/plugins/base.py
```diff
@@ -47,6 +50,7 @@ def __init__(
     ) -> None:
         self._options = properties
         self._part_info = part_info
+        self._action_properties: ActionProperties
```
tigarmo, line 53: does this actually work? Is it to satisfy `mypy`?

tests/integration/lifecycle/test_plugin_changesets.py
```diff
@@ -0,0 +1,235 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2021 Canonical Ltd.
```
tigarmo, line 3:
```suggestion
# Copyright 2022 Canonical Ltd.
```

tigarmo (approved):
Very nice! I only had a couple of fairly minor comments

craft_parts/plugins/base.py
cmatsuoka: Yes, it's enforcing typing as a hint to mypy or pyright.

tests/integration/lifecycle/test_plugin_changesets.py
cmatsuoka: Updating as suggested.

PR #313, opened 2022-10-27
Add ubuntu-22.04 to CI runners

tigarmo:
This is needed for a proper Chisel integration test, and it's also healthy to run the tests on the latest LTS. I did uncover an issue with the `GoPlugin` (`go mod download` semantics changed) and discovered that the overlay smoke test doesn't work (at least, not out-of-the-box). It's a TODO for now because it'll take some investigating but we should look into it soon

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -54,6 +54,9 @@ jobs:
       matrix:
         os: [ubuntu-20.04]
         python-version: ["3.7", "3.8", "3.9", "3.10"]
+        include:
+          - os: ubuntu-22.04
+            python-version: "3.10"
 
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yaml
```diff
@@ -97,10 +102,11 @@ jobs:
         run: |
           make test-integrations
       - name: Run overlay smoke test
+        if: ${{ matrix.adjective == 'focal' }} #TODO: fix this smoke test for jammy
```
tigarmo, line 105: failure if enabled: https://github.com/canonical/craft-parts/actions/runs/3339121000/jobs/5527668931

cmatsuoka: This is interesting, because `/usr/bin/apt-key` is ithere and can be executed. I wonder if this is related to the bad character device problem we had in rockcraft.

cmatsuoka (approved):
Thanks!

PR #315, opened 2022-11-02
sources: add the plain file source handler

### LINKED ISSUE #136: Can't download an uncompressed binary
Similar to https://bugs.launchpad.net/snapcraft/+bug/1839842.

My use case is that I'd like to download a binary like https://dl.k8s.io/release/v1.22.1/bin/linux/amd64/kubectl

That doesn't currently work from what I can see, and requires a workaround similar to the snapcraft bug:

```yaml
parts:
  kubectl:
    plugin: dump
    source: .
    override-pull: |
      curl -sSLO https://dl.k8s.io/release/v1.22.1/bin/linux/amd64/kubectl
      chmod +x kubectl
    build-packages:
      - curl
```

cmatsuoka:
Expose source type `file` to use source files that don't require provisioning.

-----
Fixes #136

### DIFF
--- a/craft_parts/sources/base.py
+++ b/craft_parts/sources/base.py
@@ -178,6 +178,7 @@ class FileSourceHandler(SourceHandler):
 
     # pylint: enable=too-many-arguments
 
+    @abc.abstractmethod
     def provision(
         self,
         dst: Path,
--- a/craft_parts/sources/base.py
+++ b/craft_parts/sources/base.py
@@ -178,7 +178,6 @@ class FileSourceHandler(SourceHandler):
 
     # pylint: enable=too-many-arguments
 
-    @abc.abstractmethod
     def provision(
         self,
         dst: Path,
--- a/craft_parts/sources/file_source.py
+++ b/craft_parts/sources/file_source.py
@@ -1,0 +1,73 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2022 Canonical Ltd.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License version 3 as published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Implement the plain file source handler."""
+
+from pathlib import Path
+from typing import List, Optional
+
+from craft_parts.dirs import ProjectDirs
+
+from . import errors
+from .base import FileSourceHandler
+
+
+class FileSource(FileSourceHandler):
+    """The tar source handler."""
+
+    # pylint: disable=too-many-arguments
+    def __init__(
+        self,
+        source: str,
+        part_src_dir: Path,
+        *,
+        cache_dir: Path,
+        source_tag: Optional[str] = None,
+        source_commit: Optional[str] = None,
+        source_branch: Optional[str] = None,
+        source_depth: Optional[int] = None,
+        source_submodules: Optional[List[str]] = None,
+        source_checksum: Optional[str] = None,
+        project_dirs: Optional[ProjectDirs] = None,
+        ignore_patterns: Optional[List[str]] = None,
+    ):
+        super().__init__(
+            source,
+            part_src_dir,
+            cache_dir=cache_dir,
+            source_commit=source_commit,
+            source_tag=source_tag,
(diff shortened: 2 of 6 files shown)


craft_parts/sources/file_source.py
```diff
@@ -0,0 +1,73 @@
...
+from typing import List, Optional
+
+from craft_parts.dirs import ProjectDirs
+
+from . import errors
+from .base import FileSourceHandler
+
+
+class FileSource(FileSourceHandler):
+    """The tar source handler."""
```
tigarmo, line 29: update this docstring

craft_parts/sources/file_source.py
```diff
@@ -0,0 +1,73 @@
...
+
+    # pylint: disable=too-many-arguments
```
tigarmo, line 31: (minor) I think `disable-next` could work here? Not sure if that works with multi-line function signatures

craft_parts/sources/base.py
```diff
@@ -178,7 +178,6 @@ def __init__(
 
     # pylint: enable=too-many-arguments
 
-    @abc.abstractmethod
```
tigarmo, line 181: from my understanding of the PR, it looks like you removed this `abstractmethod` decorator here because `FileSource` doesn't need to do any provisioning, right? If that's the case, I think it might be preferable to keep the `abstractmethod` and have `FileSource` override it with an empty implementation. It's not a strong preference but I believe it's good that future `FileSourceHandler` specializations *have* to implement this, even if as a no-op, because it means they will at least have to think about.

tests/integration/sources/test_file.py
```diff
@@ -0,0 +1,70 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2021 Canonical Ltd.
```
tigarmo, line 3:
```suggestion
# Copyright 2022 Canonical Ltd.
```

tests/integration/sources/test_file.py
```diff
@@ -0,0 +1,70 @@
...
+    )
+
+    with lf.action_executor() as ctx:
+        ctx.execute(Action("foo", Step.PULL))
+
+    foo_src_dir = Path("parts", "foo", "src")
+    assert Path(foo_src_dir, "foobar").read_text() == "content"
+
+
+def test_source_tar_error(new_dir):
```
tigarmo, line 52:
```suggestion
def test_source_file_error(new_dir):
```

tests/unit/sources/test_file_source.py
```diff
@@ -0,0 +1,66 @@
...
+
+import pytest
+import requests
+
+from craft_parts.sources import sources
+
+
+@pytest.mark.http_request_handler("FakeFileHTTPRequestHandler")
+class TestFileSource:
+    """Tests for the tar source handler."""
```
tigarmo, line 27:
```suggestion
    """Tests for the file source handler."""
```

tests/unit/sources/test_file_source.py
```diff
@@ -0,0 +1,66 @@
...
+            "50ad76d1c2f50c4935d11d50211945ca0ecb980c04c98099"
+            "085b0c3"
+        )
+        file_source = sources.FileSource(
+            source, Path(), cache_dir=new_dir, source_checksum=expected_checksum
+        )
+
+        file_source.pull()
+
+        download_spy = mocker.spy(requests, "get")
```
tigarmo, line 64: a suggestion: move this `spy` call to before the first `pull()`, and change the assertion to check that `call_count == 1`. This way we can be sure that `requests.get()` is being called, and in the (unlikely) event that the backend implementation changes this will correctly break and need to be updated.

tigarmo (changes requested):
nice! The only changes that I would consider 'mandatory' are the references to 'tar' and the copyright year, everything else you can consider just suggestions.

craft_parts/sources/file_source.py
cmatsuoka: Oops. Fixed.

craft_parts/sources/file_source.py
cmatsuoka: Good idea, that works.

tests/integration/sources/test_file.py
cmatsuoka: Typo fixed.

craft_parts/sources/base.py
cmatsuoka: The overall structure of the sources should be refactored at some point to properly handle vcs-related sources (that accept `source-tag`, `source-branch`, `source-commit`, etc) but until then I'll do as suggested and keep the abstract method, overriding it in the plain file source handler implementation.

tests/unit/sources/test_file_source.py
cmatsuoka: Changed as suggested.

tests/integration/sources/test_file.py
cmatsuoka: Fixed.

tigarmo (approved):
nice!

PR #317, opened 2022-11-10
info: pass build attributes and state to post-step callback

cmatsuoka:
Add build attributes and step execution state to the part information passed to post-step callbacks. This allows further processing of migrated files according to part properties set by the user.

-----

### DIFF
--- a/tests/integration/executor/test_callbacks.py
+++ b/tests/integration/executor/test_callbacks.py
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright 2021 Canonical Ltd.
+# Copyright 2021-2022 Canonical Ltd.
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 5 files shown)


tests/integration/executor/test_callbacks.py
```diff
@@ -191,6 +198,25 @@ def test_callback_post(tmpdir, capfd, step, action_type):
         assert out == f"step {step!r}\ncallback\n"
 
 
+@pytest.mark.parametrize("step", list(Step))
+def test_callback_post_state_info(tmpdir, step):
+    callbacks.register_post_step(_state_step_callback, step_list=[step])
```
tigarmo, line 203: I think it would be useful to make sure that the callback was actually called, to prevent the unlikely-but-possible scenario of the code being refactored in a way that makes the callback silently stop being called.
Something like this, with `_state_step_callback()` as an inner function:

```python
def test_callback_post_state_info(tmpdir, step):
    callback_called = [False]

    def _state_step_callback(info: StepInfo) -> bool:
        assert info.state is not None
        assert isinstance(info.state, StepState)
        callback_called[0] = True
        return True

    # ... rest of test
    assert callback_called[0]
```

tigarmo (approved):

tests/integration/executor/test_callbacks.py
cmatsuoka: Implemented as suggested.

PR #324, opened 2022-11-22
plugins: add SConsPlugin

tigarmo:
This plugin builds SCons projects and is heavily based on Snapcraft's core18 plugin. Remarks:

- The plugin builds artifacts through two commands: "scons" and "scons install". This assumes that "install" is a valid target in the source's SConstruct file.
- Since there is no "official" way to configure the installation dir for SCons-generated artifacts, the plugin assumes the common convention of setting the DESTDIR environment variable to the part's install dir.

-----

CRAFT-1418

### DIFF
--- a/craft_parts/plugins/scons_plugin.py
+++ b/craft_parts/plugins/scons_plugin.py
@@ -27,7 +27,7 @@ from .properties import PluginProperties
 class SConsPluginProperties(PluginProperties, PluginModel):
     """The part properties used by the SCons plugin."""
 
-    scons_options: List[str] = []
+    scons_parameters: List[str] = []
 
     # part properties required by the plugin
     source: str
@@ -98,7 +98,7 @@ class SConsPlugin(Plugin):
 
     The plugin supports the following keywords:
 
-    - ``scons-options``
+    - ``scons-parameters``
       (list of strings)
       Additional values to pass to the ``scons`` and ``scons install`` command
       lines.
@@ -126,6 +126,6 @@ class SConsPlugin(Plugin):
         options = cast(SConsPluginProperties, self._options)
 
         return [
-            " ".join(["scons"] + options.scons_options),
-            " ".join(["scons", "install"] + options.scons_options),
+            " ".join(["scons"] + options.scons_parameters),
+            " ".join(["scons", "install"] + options.scons_parameters),
         ]
--- a/craft_parts/plugins/scons_plugin.py
+++ b/craft_parts/plugins/scons_plugin.py
@@ -53,8 +53,8 @@ class SConsPlugin(Plugin):
     (C/C++ compiler, Java compiler, etc) must be declared via ``build-packages``.
     Since there is no "official" way of defining the target installation directory
     for SCons-built artifacts, the default build will set the DESTDIR environment
-    variable which the SConstruct file should use to configure its ``Install()``
-    builder target.
+    variable which contains the root which the SConstruct file should use to
+    configure its ``Install()`` builder target.
 
     The plugin supports the following keywords:
 
@@ -77,7 +77,7 @@ class SConsPlugin(Plugin):
     def get_build_environment(self) -> Dict[str, str]:
         """Return a dictionary with the environment to use in the build step."""
         return {
-            "DESTDIR": f"{self._part_info.part_install_dir}/bin",
+            "DESTDIR": f"{self._part_info.part_install_dir}",
         }
 
     def get_build_commands(self) -> List[str]:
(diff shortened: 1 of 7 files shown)


craft_parts/plugins/scons_plugin.py
```diff
@@ -0,0 +1,92 @@
...
+        :raise pydantic.ValidationError: If validation fails.
+        """
+        plugin_data = extract_plugin_properties(
+            data, plugin_name="scons", required=["source"]
+        )
+        return cls(**plugin_data)
+
+
+class SConsPlugin(Plugin):
+    """A plugin for go projects using go.mod.
```
tigarmo, line 50: remember to update this copy/pasted docstring

craft_parts/plugins/scons_plugin.py
```diff
@@ -0,0 +1,90 @@
...
+    The plugin supports the following keywords:
+
+    - ``scons-options``
+      (list of strings)
+      Additional values to pass to the ``scons`` and ``scons install`` command
+      lines.
+    """
+
+    properties_class = SConsPluginProperties
+
```
sergiusens, line 68: I would add a validator_class here too

tigarmo: what should it validate? Should the plugin not have a 'hardcoded' dependency on the `scons` package?

tests/integration/plugins/test_scons.py
```diff
@@ -0,0 +1,39 @@
...
+    lf = LifecycleManager(
+        parts, application_name="test_scons", cache_dir=new_dir, work_dir=new_dir
+    )
+    actions = lf.plan(Step.PRIME)
+
+    with lf.action_executor() as ctx:
+        ctx.execute(actions)
+
+    binary = Path(lf.project_info.prime_dir, "bin", "hello")
+    assert binary.is_file()
```
cmatsuoka, line 35: Since the package installs the built artifact directly at the destination root instead of bindir, it should test for the presence there as well (at `prime/hello`).

craft_parts/plugins/scons_plugin.py
cmatsuoka: If we follow the pattern used by the go and npm plugins, it can allow the user to use existing scons or even build it in a different part.

tigarmo: Updated the code to follow the go plugin's validation (the tests are basically copy/paste)

craft_parts/plugins/scons_plugin.py
```diff
@@ -0,0 +1,90 @@
...
+        return set()
+
+    def get_build_packages(self) -> Set[str]:
+        """Return a set of required packages to install in the build environment."""
+        return {"scons"}
+
+    def get_build_environment(self) -> Dict[str, str]:
+        """Return a dictionary with the environment to use in the build step."""
+        return {
+            "DESTDIR": f"{self._part_info.part_install_dir}/bin",
```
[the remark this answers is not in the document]
tigarmo: Very glad you caught this, thanks. I somehow misread the original implementation, but you right that `DESTDIR` should point to the install dir. Update this var and the test that uses it

tests/integration/plugins/test_scons.py
tigarmo: Updated the `SConstruct` file to build into `DESTDIR/bin`

craft_parts/plugins/scons_plugin.py
```diff
@@ -0,0 +1,131 @@
...
+from .. import errors
+from . import validator
+from .base import Plugin, PluginModel, extract_plugin_properties
+from .properties import PluginProperties
+
+
+class SConsPluginProperties(PluginProperties, PluginModel):
+    """The part properties used by the SCons plugin."""
+
+    scons_options: List[str] = []
```
cmatsuoka, line 30: Other plugins seem to use `<name>_parameters`, we may want to follow this pattern for consistency.

cmatsuoka:
I used the plugin to build a real scons-based snap and it worked well -- the only thing is that my package didn't seem to follow the `DESTDIR` convention so I had to define `destdir=` in the plugin parameters

craft_parts/plugins/scons_plugin.py
sergiusens: yes, I would make things general change if possible

tigarmo: I originally used `scons_options` to mirror the original plugin in snapcraft, but I'm fine with `scons_parameters`. I'll update it

tigarmo: updated

tigarmo:
As far as I can tell that a general SCons issue, every project handles the target directory setting differently

PR #332, opened 2022-11-29
plugins: add ant plugin

tigarmo:
The plugin requires the ``ant`` tool installed on the system. This can be achieved by adding the appropriate declarations to ``build-packages`` or ``build-snaps``, or by having it installed or built in a different part. In this case, the name of the part supplying ``ant`` must be "ant-deps".

Additionally, Java projects need a dev kit (jdk) to build and a runtime environment (jre) to run. There are multiple choices here, but frequently adding ``default-jdk-headless`` to ``build-packages`` and ``default-jre-headless`` to ``stage-packages`` is enough.

Once built, the plugin will create the following structure in the part's install dir (which will later be staged/primed/packaged):

- A ``bin/java`` symlink pointing to the actual ``java`` binary provided by the jre;
- A ``jar/`` directory containing the .jar files generated by the build.

-----

CRAFT-1464


craft_parts/plugins/ant_plugin.py
```diff
@@ -0,0 +1,183 @@
...
+from .properties import PluginProperties
+
+logger = logging.getLogger(__name__)
+
+
+class AntPluginProperties(PluginProperties, PluginModel):
+    """The part properties used by the Ant plugin."""
+
+    ant_build_targets: List[str] = []
+    ant_build_file: Optional[str] = None
```
tigarmo, line 35: in the core18 plugin this as called `ant-buildfile`, because the Ant option is called `-buildfile`, but I thought I'd keep it consistent with `ant-build-targets`.

craft_parts/plugins/ant_plugin.py
```diff
@@ -0,0 +1,183 @@
...
+            command.append(f"-D{prop_name}={prop_value}")
+
+        command.extend(options.ant_build_targets)
+
+        return [
+            " ".join(command),
+        ] + get_java_post_build_commands()
+
+
+def get_java_post_build_commands() -> List[str]:
```
tigarmo, line 157: This is what the core18 ant and maven plugins do; as a function here we can reuse it later for other java-based plugins

tests/integration/plugins/test_ant.py
```diff
@@ -0,0 +1,60 @@
...
+    monkeypatch.setenv("CRAFT_PARTS_PACKAGE_REFRESH", "0")
+    source_location = Path(__file__).parent / "test_ant"
+
+    parts_yaml = textwrap.dedent(
+        f"""
+        parts:
+          foo:
+            plugin: ant
+            source: {source_location}
+            stage-packages: [default-jre-headless]
```
tigarmo, line 36: I was surprised to see that this test passed on CI without me having to add `ant` and the jdk on the part.

craft_parts/plugins/ant_plugin.py
```diff
@@ -0,0 +1,183 @@
...
+        :raises PluginEnvironmentValidationError: If ``ant`` is invalid
+        and there are no parts named ant.
+        """
+        version = self.validate_dependency(
+            dependency="ant",
+            plugin_name="ant",
+            part_dependencies=part_dependencies,
+            argument="-version",
+        )
+        if not version.startswith("Apache Ant") and (
```
cmatsuoka, line 77: It's a good thing that ant doesn't use control characters for terminal formatting to display the version like maven does.

tests/unit/plugins/test_ant_plugin.py
```diff
@@ -0,0 +1,158 @@
...
+            "ant-properties": {"prop1": "1", "prop2": "2"},
+        }
+    )
+    plugin = AntPlugin(properties=properties, part_info=part_info)
+
+    assert (
+        plugin.get_build_commands()
+        == ["ant -f myfile.txt -Dprop1=1 -Dprop2=2 compile jar"]
+        + get_java_post_build_commands()
+    )
```
cmatsuoka, line 158: Just nitpick and really important, but black likes to format this kind of line in two different ways and the other formatting is what we generally use in craft-parts, I think it's easier to read.
```
assert plugin.get_build_commands() == (
    something
)
```

tests/integration/plugins/test_ant.py
cmatsuoka: If jdk is pre-installed in the environment, can something unexpected happen if we use a different version at runtime? If so it could be interesting to make sure the stage package has a consistent version.

craft_parts/plugins/ant_plugin.py
cmatsuoka: Maybe we could have a Java base plugin superclass providing the infrastructure as protected methods to java-related plugins? We can think about that and refactor later if we decide to do something like that.

tigarmo:
@cmatsuoka thanks for the review. I fixed the assertion notation and added CRAFT-1526 for the remaining items as possible improvements

cmatsuoka (approved):
Thanks Tiago!

PR #335, opened 2022-12-14
packages: improve chisel error handling

tigarmo:
Capture the output of 'chisel cut' to add the (short) error message to the generated exception, which is now a more specific ChiselError.

-----

CRAFT-1499


cmatsuoka (approved):
Adding additional error details as such is a sound approach. We can think about having some common infrastructure to handle process execution in a better/more flexible way in a future refactoring.

PR #337, opened 2022-12-16
Linters: migrate from flake8 to ruff

lengau:
This configures ruff and sets it to the default, while still giving
developers the option to run flake8 if they choose using 
`make test-flake8`
CRAFT-1506


pyproject.toml
```diff
@@ -25,3 +25,93 @@ extension-pkg-whitelist = [
...
+    "N",  # Pep8 naming
+    "YTT",  # flake8-2020: Misuse of `sys.version` and `sys.version_info`
+    "ANN",  # Annotations
+    "S",  # Checks for common security issues
+    "BLE",  # Blind exception
+    "B",  # Opinionated bugbear linting
+    "C4",  # better comprehensions
+    "T10",  # Ensure we don't leave code that starts the debugger in
+    "ICN",  # Unconventional import aliases.
+    "Q",  # oh gawd there's gonna be a lot of this isn't there
```
tigarmo, line 52: what's this one?

Makefile
```diff
@@ -72,8 +72,13 @@ test-codespell:
 
 .PHONY: test-flake8
 test-flake8:
+	echo "\033[0;31mWARNING\033[0m: Did you mean to run \`make ruff\`?"
 	flake8 $(SOURCES)
 
+.PHONY: test-ruff
+test-ruff:
+	ruff .
```
tigarmo, line 80: maybe `ruff $(SOURCES)` instead? To avoid looking at virtual env files, etc

tigarmo (commented):
nice!

Makefile
lengau: The default `exclude` excludes .git, venv, build, etc. ([Full default config](https://github.com/charliermarsh/ruff#configuration)), so this ends up scanning the same files. I'll update it to use `$(SOURCES)` for consistency though.

pyproject.toml
lengau: Lol oops I forgot to update the comment. The comment was meant for when I tried the `T20` rules which complain about each usage of `print` (I was curious - it makes sense for a web API but not for what we're using here).

Q is for consistent quotations. Line 101 turns off sensing bad quotes in inline strings just because we have 2 single-quoted strings.

Related: I [made a feature request](https://github.com/charliermarsh/ruff/issues/1256) for a warning level in ruff so we can have things like that enabled as warnings while still allowing them to pass linting in CI.

tigarmo (approved):
thanks!

PR #339, opened 2022-12-22
plugins: add base class for java-related plugins

cmatsuoka:
Add a common base class for plugins that build Java projects. This is currently used by the Ant plugin, but it should be also used by the upcoming Maven plugin.

-----


tigarmo (approved):
nice!

PR #343, opened 2023-01-02
states: make stage package tracking optional

cmatsuoka:
Stage package tracking is used Snapcraft to add a list of primed stage packages to the manifest file when `--enable-manifest` is used. This is implemented using extended file attributes in Linux.

Making stage package tracking optional brings in a number of advantages: it allows other applications such as Rockcraft and Charmcraft to run faster and consume less resources, and also allows work dirs on tmpfs when package tracking is disabled (e.g. when configuring the default lxd profile for application projects to mount `/tmp` and `/root` as tmpfs).

-----


tigarmo (approved):
nice! Thanks

PR #345, opened 2023-01-04
executor: fix directory cleaning

### LINKED ISSUE #329: Executor clean may not be 100% effective
in `executor/executor.py`, `Executor.clean` has this code:

```
            with contextlib.suppress(FileNotFoundError):
                shutil.rmtree(self._project_info.prime_dir)
                if initial_step <= Step.STAGE:
                    shutil.rmtree(self._project_info.stage_dir)
                if initial_step <= Step.PULL:
                    shutil.rmtree(self._project_info.parts_dir)
```

Three `rmtree` surrounded by the `supress` context manager. If the first `rmtree` fails with `FileNotFoundError` the exception will be suppressed, but the other two `rmtree` will NOT be run.

It's worth to check if this pattern is repeated elsewhere.

cmatsuoka:
Fix bad logic in the executor cleaning that could lead to lifecycle directories that should be cleaned unexpectedly remaining after clean is issued. Thanks to Facundo Batista for spotting this bug, fixes #329.

-----

### DIFF
--- a/craft_parts/executor/executor.py
+++ b/craft_parts/executor/executor.py
@@ -16,7 +16,6 @@
 
 """Definitions and helpers for the action executor."""
 
-import contextlib
 import logging
 import shutil
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


facundobatista (approved):
Wonderful, thanks!

PR #349, opened 2023-01-06
linters: fix new ruff issues

tigarmo:
ruff is a newish tool and it's getting a lot of new releases; we don't currently pin it and the latest versions need two fixes: be more permissive with temporary dirs in tests, and mark that hashlib.sha1() is not being used for security.

-----

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -110,6 +110,8 @@ ignore = [
     "ANN", # Ignore type annotations in tests
     "S101",  # Yeah of course we assert in tests
     "B009",  # Allow calling `getattr` in tests since it can be used to make the test clearer.
+    "S103", # Allow `os.chmod` setting a permissive mask `0o555` on file or directory
+    "S108", # Allow Probable insecure usage of temporary file or directory
 ]
 "__init__.py" = ["I001"]  # Imports in __init__ filesare allowed to be out of order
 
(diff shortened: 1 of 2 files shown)


lengau (approved):
I made one addition, freezing the Ruff version so new releases don't come along and break things again. When we add renovate to this repository the Renovate PRs to update ruff will be the ones that fail these checks, preventing this.

https://github.com/tigarmo/craft-parts/pull/1

pyproject.toml
```diff
@@ -110,6 +110,8 @@ ignore = [
     "ANN", # Ignore type annotations in tests
     "S101",  # Yeah of course we assert in tests
     "B009",  # Allow calling `getattr` in tests since it can be used to make the test clearer.
+    "S103", # Allow `os.chmod` setting a permissive mask `0o555` on file or directory
+    "S108", # Allow Probable insecure usage of temporary file or directory
```
cmatsuoka, line 114: It's better to not include these project-wide, and instead add a noqa in the test where the 555 chmod is used, and refactor the plugin tests to not set `part_info._part_install_dir` and test against `new_dir / "parts/foo/install"` instead.

tigarmo: It doesn't show on the default diff but these ignores only affect the `"tests/**.py"` files:
image

tigarmo: do you think we should update the test code regardless?

cmatsuoka: If it's only for tests it should be fine.

cmatsuoka (approved):

PR #350, opened 2023-01-16
linters: set disallow_untyped_defs in mypy

tigarmo:
This forces type annotations for functions in craft_parts/.

-----

### DIFF
--- a/craft_parts/executor/executor.py
+++ b/craft_parts/executor/executor.py
@@ -19,7 +19,7 @@
 import logging
 import shutil
 from pathlib import Path
-from typing import Dict, List, Optional, Union
+from typing import Any, Dict, List, Optional, Union
 
 from craft_parts import callbacks, overlays, packages, parts, plugins
 from craft_parts.actions import Action, ActionType
@@ -280,7 +280,7 @@ class ExecutionContext:
         self._executor.prologue()
         return self
 
-    def __exit__(self, *exc):
+    def __exit__(self, *exc: Any) -> None:
         self._executor.epilogue()
 
     def execute(
--- a/craft_parts/executor/step_handler.py
+++ b/craft_parts/executor/step_handler.py
@@ -158,14 +158,14 @@ class StepHandler:
         srcdir = str(self._part.part_install_dir)
         files, dirs = filesets.migratable_filesets(stage_fileset, srcdir)
 
-        def pkgconfig_fixup(file_path):
+        def pkgconfig_fixup(file_path: str) -> None:
             if os.path.islink(file_path):
                 return
             if not file_path.endswith(".pc"):
                 return
             packages.fix_pkg_config(
                 prefix_prepend=self._part.stage_dir,
-                pkg_config_file=file_path,
+                pkg_config_file=Path(file_path),
                 prefix_trim=self._part.part_install_dir,
             )
 
--- a/craft_parts/overlays/overlay_manager.py
+++ b/craft_parts/overlays/overlay_manager.py
@@ -20,7 +20,9 @@ import logging
 import os
 import sys
 from pathlib import Path
-from typing import List, Optional
+from typing import Any, List, Optional
+
+from typing_extensions import Literal
 
 from craft_parts import packages
 from craft_parts.infos import ProjectInfo
@@ -127,7 +129,8 @@ class OverlayManager:
 
         mount_dir = self._project_info.overlay_mount_dir
         # Ensure we always run refresh_packages_list by resetting the cache
-        packages.Repository.refresh_packages_list.cache_clear()
+        if hasattr(packages.Repository.refresh_packages_list, "cache_clear"):
+            packages.Repository.refresh_packages_list.cache_clear()  # type: ignore
         chroot.chroot(mount_dir, packages.Repository.refresh_packages_list)
 
     def download_packages(self, package_names: List[str]) -> None:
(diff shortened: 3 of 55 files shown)


tigarmo:
@cmatsuoka the second commit has test code that I had to update on account of typing fixes

craft_parts/executor/executor.py
```diff
@@ -280,7 +280,7 @@ def __enter__(self) -> "ExecutionContext":
         self._executor.prologue()
         return self
 
-    def __exit__(self, *exc):
+    def __exit__(self, *exc: Any) -> None:
```
lengau, line 283: Do we want to register this class as a subclass of `contextlib.AbstractContextManager`?

craft_parts/executor/step_handler.py
```diff
@@ -158,14 +158,14 @@ def _builtin_stage(self) -> StepContents:
         srcdir = str(self._part.part_install_dir)
         files, dirs = filesets.migratable_filesets(stage_fileset, srcdir)
 
-        def pkgconfig_fixup(file_path):
+        def pkgconfig_fixup(file_path: str) -> None:
```
lengau, line 161:
```suggestion
        def pkgconfig_fixup(file_path: Union[os.PathLike, str]) -> None:
```

craft_parts/overlays/overlay_manager.py
```diff
@@ -127,7 +129,8 @@ def refresh_packages_list(self) -> None:
 
         mount_dir = self._project_info.overlay_mount_dir
         # Ensure we always run refresh_packages_list by resetting the cache
-        packages.Repository.refresh_packages_list.cache_clear()
+        if hasattr(packages.Repository.refresh_packages_list, "cache_clear"):
```
lengau, line 132: Possibly more of a question for @cmatsuoka, but is there any reason to not make the abstract `BaseRepository` have an `lru_cache`?

cmatsuoka: The `packages` package code was copied from Snapcraft and it contains many areas that could be improved (or fully refactored). I can't say much about this specific case without verifying the code in detail.

craft_parts/packages/snaps.py
```diff
@@ -339,12 +339,12 @@ def _get_parsed_snap(snap: str) -> Tuple[str, str]:
...
 
 
-def get_snapd_socket_path_template():
+def get_snapd_socket_path_template() -> str:
     """Return the template for the snapd socket URI."""
     return "http+unix://%2Frun%2Fsnapd.socket/v2/{}"
 
 
-def _get_local_snap_file_iter(snap_name: str, *, chunk_size: int):
+def _get_local_snap_file_iter(snap_name: str, *, chunk_size: int) -> Iterator:
```
lengau, line 347: The [typeshed](https://github.com/python/typeshed/blob/main/stubs/requests/requests/models.pyi#L109) for requests agrees with what's written here, but I'm pretty sure the `Response` object's `iter_content` will return either an `Iterator[str]` if `decode_unicode=True` or an `Iterator[bytes]` if `decode_unicode=False`.

(Likely relevant typeshed issue: https://github.com/python/typeshed/issues/9550)

craft_parts/sources/base.py
```diff
@@ -77,7 +77,7 @@ def __init__(
         self.source_branch = source_branch
         self.source_depth = source_depth
         self.source_checksum = source_checksum
-        self.source_details = None
+        self.source_details: Any = None
```
lengau, line 80:
```suggestion
        self.source_details: Optional[Dict[str, Optional[str]]] = None
```

Are other types of sources planning to provide something else? (Right now only `git_source` provides details)

craft_parts/sources/git_source.py
```diff
@@ -164,12 +164,13 @@ def __init__(  # pylint: disable=too-many-arguments
...
-    def _fetch_origin_commit(self):
+    def _fetch_origin_commit(self) -> None:
         """Fetch from origin, using source-commit if defined."""
+        command_str = cast(str, self.command)
         command = [
-            self.command,
+            command_str,
             "-C",
-            self.part_src_dir,
+            str(self.part_src_dir),
```
lengau, line 173: Can you add this to lines 187 and 318?

craft_parts/overlays/overlay_manager.py
lengau: In that case this can be a mystery for another day.

lengau (approved):
Everything I've commented has been the rambling thoughts of someone new to the codebase, so take them for what they're worth there

craft_parts/executor/step_handler.py
tigarmo: but then it'd fail on line 164:
```python
            if not file_path.endswith(".pc"):
                return
```

craft_parts/packages/snaps.py
tigarmo: Thanks, update to `Iterator[bytes]`
good call

craft_parts/sources/base.py
tigarmo: Updated to `Optional[Dict[str, Optional[str]]]`, if a subclass changes this in the future we can revisit it

tigarmo:
@lengau the comments are much appreciated, thanks. Updated the PR based on your suggestions

lengau (approved):

craft_parts/executor/executor.py
lengau: I don't really have a strong reasoning either way. I'm just in a habit of doing stuff like that because it's saved me lots of debugging with IDE tools in the past. In this case I meant the question literally and didn't mean to imply it as a suggestion.

craft_parts/executor/step_handler.py
lengau: I read the function twice and still managed to miss that. Sorry!

craft_parts/executor/executor.py
tigarmo: oh alright! I could have done that but I merged the PR after this was marked as resolved. Oops!

PR #340, opened 2022-12-22
plugins: add maven plugin

cmatsuoka:
Port the Snapcraft V1 Maven plugin and use the common java plugin infrastructure shared with the Ant plugin.

This is currently stacked on top of PR #339.

- [ ] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?

-----

PR #351, opened 2023-01-17
linters: add PLR2004 to ruff ignore

tigarmo:
PLR2004 is "Magic value used in comparison, consider using a constant". This is triggering for things like expected messages in tests, so there were over 600 instances in craft-parts.

-----


cmatsuoka (approved):
A variation of this fix was also added to PR #340.

PR #357, opened 2023-02-01
tests: Ensure maven test sets environment

lengau:

-----

I came across this while running the integration tests. As is, the maven test succeeds when running all tests in order, but fails when run without running the ant test. The change in `test_maven.py` fixes the test so it can run independently, and the change in `__init__.py` ensures that cached functions get their caches cleared after each integration test, keeping them more independent.


tigarmo (approved):
good catch!

PR #359, opened 2023-02-01
chore(lint): update linters and fix linting issues with new rules

lengau:

-----

### DIFF
--- a/tests/unit/executor/test_organize.py
+++ b/tests/unit/executor/test_organize.py
@@ -29,130 +29,130 @@ from craft_parts.executor.organize import organize_files
     "data",
     [
         # simple_file
-        dict(
-            setup_dirs=[],
-            setup_files=["foo"],
-            organize_map={"foo": "bar"},
-            expected=[(["bar"], "")],
-            expected_message=None,
-            expected_overwrite=None,
-        ),
+        {
+            "setup_dirs": [],
+            "setup_files": ["foo"],
+            "organize_map": {"foo": "bar"},
+            "expected": [(["bar"], "")],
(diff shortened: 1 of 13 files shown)


tests/unit/executor/test_organize.py
```diff
@@ -29,130 +29,130 @@
...
-        dict(
-            setup_dirs=[],
-            setup_files=["foo"],
-            organize_map={"foo": "bar"},
-            expected=[(["bar"], "")],
-            expected_message=None,
-            expected_overwrite=None,
-        ),
+        {
+            "setup_dirs": [],
```
tigarmo, line 33: I'm not necessarily against this lint but I wonder why it exists? This version is (very minorly) less readable to me

tigarmo (approved):

tests/unit/executor/test_organize.py
lengau: I'm conflicted about it in this specific case, because it's tests. In production code I prefer literals, but I agree that the previous version was more readable.

Pros of dict literals: 
1. They're faster. In this particular case, the entire literal is turned into a memory area at compile time, so at runtime the VM can effectively just do a `memcpy`, whereas before it would actually have to call the function. This even extends to when it's not purely a literal (such as where we have `os.path.join` there), as cpython is smart enough to leave a null pointer and fill it in.
2. Consistency with dict literals vs. other literals.

Cons of dict literals:
1. Yeah they're not as nice to read as the function call IMO.

It's not a hill I'm willing to die on in either direction, so the path of least resistance was to make it meet those linting rules rather than changing them.

PR #361, opened 2023-02-06
plugins: implement strict/offline plugins

tigarmo:
The first commit has @cmatsuoka 's original code with some minor linting by me. The second commit has the tests I added.
One thing we should maybe discuss is this `strict`/`offline` "duality". Really what a "strict" plugin currently means is that it does offline builds, but is that definition something that we envision changing in the future? Should we use "strict" throughout the codebase?

(CRAFT-1587)

### DIFF
--- a/craft_parts/executor/step_handler.py
+++ b/craft_parts/executor/step_handler.py
@@ -26,7 +26,7 @@ import tempfile
 import textwrap
 import time
 from pathlib import Path
-from typing import List, Optional, Set, TextIO, Union
+from typing import List, Optional, Set, TextIO, Union, cast
 
 from craft_parts import errors, packages
 from craft_parts.infos import StepInfo
@@ -36,6 +36,7 @@ from craft_parts.sources import SourceHandler
 from craft_parts.steps import Step
 from craft_parts.utils import file_utils
 
+from ..plugins.base import StrictPlugin
 from . import filesets
 from .filesets import Fileset
 from .migration import migrate_files
@@ -108,6 +109,26 @@ class StepHandler:
     def _builtin_pull(self) -> StepContents:
         if self._source_handler:
             self._source_handler.pull()
+
+        # Plugin prepare commands.
+        if self._step_info.offline_build:
+            # If offline build is enabled we're using only strict plugins
+            plugin = cast(StrictPlugin, self._plugin)
+            pull_prepare_commands = plugin.get_pull_prepare_commands()
+
+            try:
+                _create_and_run_script(
+                    pull_prepare_commands,
+                    script_path=self._part.part_run_dir.absolute() / "pull.sh",
+                    cwd=self._part.part_src_subdir,
+                    stdout=self._stdout,
+                    stderr=self._stderr,
+                )
+            except subprocess.CalledProcessError as process_error:
+                raise errors.PluginPullError(
+                    part_name=self._part.name
+                ) from process_error
+
         return StepContents()
 
     @staticmethod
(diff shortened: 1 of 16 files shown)


setup.py
```diff
@@ -73,7 +73,7 @@ def is_rtd() -> bool:
     "codespell",
     "coverage",
     "isort",
-    "mypy",
+    "mypy==0.991",
```
tigarmo, line 76: pinned this temporarily because of this error with mypy 1.0.0:

```
$ make test-mypy 
mypy setup.py craft_parts tests
setup.cfg:17: error: Error importing plugin "pydantic.mypy": could not convert string to float: '1.0.0'  [misc]
Found 1 error in 1 file (errors prevented further checking)
make: *** [Makefile:92: test-mypy] Error 2

Comment (lengau) [reply setup.py:76]: I think we need to pin all the linters on all our apps anyway and let Renovate handle updates

File: craft_parts/plugins/base.py [hunk 2]
```diff
@@ -81,6 +82,20 @@ def set_action_properties(self, action_properties: ActionProperties) -> None:
...
 
 
+class StrictPlugin(Plugin):
+    """Base class for strict plugins.
+
+    Strict plugins will retrieve all dependencies from remote locations during
+    the pull step, allowing the build step to run offline.
+    """
+
+    offline = True
```
cmatsuoka, line 92: I think something like `allow_offline_build` would be more descriptive here since being offline depends on changes in the environment (i.e., we're not intrinsically offline, but we will still work if the build step happens on a host that's offline).

sergiusens (approved):

sergiusens:
We probably want a future: pull-environment, not sure it is needed at all now though

craft_parts/plugins/nil_plugin.py
```diff
@@ -43,7 +43,7 @@ def unmarshal(cls, data: Dict[str, Any]) -> "NilPluginProperties":
         return cls()
 
 
-class NilPlugin(Plugin):
+class NilPlugin(StrictPlugin):
```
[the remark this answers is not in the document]
sergiusens: I think not, what we can do for a "strict" plugin is disallow override-pull

cmatsuoka:
> Really what a "strict" plugin currently means is that it does offline builds, but is that definition something that we envision changing in the future?

What the "strict" plugin does is consistent with the step names: everything is pulled in the pull step, and then only build happens in the build step (so this is the proper way the plugin should behave, hence the "strict" name in opposition to a "relaxed" implementation that allows step roles to be mixed).

craft_parts/executor/step_handler.py
```diff
@@ -108,6 +109,26 @@ def run_builtin(self) -> StepContents:
     def _builtin_pull(self) -> StepContents:
         if self._source_handler:
             self._source_handler.pull()
+
+        # Plugin prepare commands.
+        if self._step_info.offline_build:
```
sergiusens, line 114: This does not make much sense to me, if the plugin is strict and we are offline, how will we pull?

cmatsuoka: `offline_build` means that the build step will work on an offline environment, the pull step still requires network access to download content.

craft_parts/plugins/base.py
sergiusens: Check the type instead of using an attrib

lengau (approved):

tigarmo:
@cmatsuoka @sergiusens new version with today's discussion is up

sergiusens (approved):
so much nicer

PR #365, opened 2023-02-07
docs: include overlays in the lifecycle documentation.

lengau:

-----

CRAFT-1597


lengau:
Lint issue is worked around in https://github.com/canonical/craft-parts/pull/366

docs/explanation/lifecycle.rst
```diff
@@ -2,7 +2,62 @@
 Lifecycle details
 *****************
 
+Each part is built in five separate steps:
+
+#. ``PULL`` — The components for the part are retrieved from their stated
```
tigarmo, line 7: afaik the PULL step includes both the fetching of the package dependencies and the fetching of the source, if relevant, right? So it might be a good idea to add that here (that this is where the part's `source` is fetched)

docs/explanation/lifecycle.rst
```diff
@@ -2,7 +2,62 @@
...
+   location and placed into a package cache area.
+#. ``OVERLAY`` — Any overlay packages are incorporated into the staging area
```
tigarmo, line 9: What is the `staging area` referenced here?

docs/explanation/lifecycle.rst
```diff
@@ -2,7 +2,62 @@
...
+   and the overlay script is run. Finally, any overlay filters are applied.
+#. ``BUILD`` — The part is built according to the particular part plugin and
+   build override.
+#. ``STAGE`` — The specified outputs from the ``BUILD`` stage are copied into
+   a unified staging area for all parts.
+#. ``PRIME`` — The specified files are copied from the staging area to the
+   priming area for use in the final project. This is distinct from ``STAGE``
```
tigarmo, line 16: The term `project` here feels a bit odd. In fact I often search for a term to use for "the set of files that are included in the actual resulting snap/rock/charm", and sometimes I use "payload" for that. Maybe we should make a term (any term, but I vote for payload as it's application-agnostic) "official"?

docs/explanation/lifecycle.rst
```diff
@@ -2,7 +2,62 @@
 Lifecycle details
 *****************
 
+Each part is built in five separate steps:
```
tigarmo, line 5: here's a suggestion from something that came up in a ROCKs discussion yesterday: users often confuse the directories that are involved in the build process, which they need to write a correct e.g. `build-override` script. Consider cross-referencing this section with the [Step Execution Environment](https://craft-parts.readthedocs.io/en/latest/reference/parts_steps.html#step-execution-environment) somehow; like saying `and the built artifacts are placed in $CRAFT_PART_INSTALL` on the `BUILD` description

docs/explanation/lifecycle.rst
```diff
@@ -2,7 +2,62 @@
...
+
+Step order
+----------
+
+While each part's steps are guaranteed to run in the order above, they are
+not necessarily run immediately following each other, especially if multiple
+parts are included in a project. While specifics are implementation-dependent,
+the general rules for combining parts are:
+
+#. ``PULL`` all parts before running further steps
```
tigarmo, line 29:
```suggestion
#. ``PULL`` all parts before running further steps.
```

docs/explanation/lifecycle.rst
```diff
@@ -2,7 +2,62 @@
...
+
 Lifecycle processing diagram
 ----------------------------
 
 .. image:: /images/lifecycle_logic.png
+
+Further Information
+-------------------
+
+Further information can be found in the `Snapcraft parts lifecycle documentation
```
tigarmo, line 58: I guess there's a lot of overlap between the two docs now huh? Should we just copy over all that content here?

tigarmo (commented):
looking good!

docs/explanation/lifecycle.rst
lengau: Yeah, I referenced that documentation quite a bit while writing this. It has a fair amount of snapcraft-specific stuff that'd need to be removed, and there appears to be stuff that's changed too (addition of the overlay step, renaming `SNAPCRAFT_*` environment variables to `CRAFT_*`, etc.)

I'm concerned about copying it wholesale because of those differences (and it's a quite out of scope of my task to bring in full lifecycle docs here — I just needed to do this since my overlay documentation depends very heavily on understanding lifecycles), but I do agree.

I've created CRAFT-1600 for this purpose.

docs/explanation/lifecycle.rst
lengau: Considering that I was referencing that very section while writing this, that should've been obvious to me!

I've xref'd the parts steps and added a list of step outputs in the parts and steps reference.

docs/explanation/lifecycle.rst
lengau: Yeah I wasn't too happy with that terminology either. Payload is much better - incorporated!

docs/explanation/lifecycle.rst
lengau: This is supposed to mean the overlay directory. I've rephrased it, please lmk what you think now.

docs/explanation/lifecycle.rst
lengau: I've clarified this — have a read and LMK what you think

tigarmo (approved):
thanks!

docs/explanation/lifecycle.rst
```diff
@@ -2,7 +2,65 @@
...
+
+#. ``PULL`` — The source and external dependencies (such as package
+   dependencies) for the part are retrieved from their stated location and
+   placed into a package cache area.
+#. ``OVERLAY`` — Any overlay packages are installed in an overlay of the
+   filesystem base, and the overlay script is run. Finally, any overlay filters
+   are applied.
+#. ``BUILD`` — The part is built according to the particular part plugin and
+   build override.
+#. ``STAGE`` — The specified outputs from the ``BUILD`` stage are copied into
```
sergiusens, line 16:
```suggestion
#. ``STAGE`` — The specified outputs from the ``BUILD`` step are copied into
```

sergiusens: We call the stages "steps" to avoid ambiguity with the stage step itself

sergiusens (approved):

PR #370, opened 2023-02-08
docs: explanation section for overlays

lengau:

-----

This depends on #365 and incorporates its changes. I will unmark as draft once that PR has been merged.

Fixes CRAFT-1560


docs/explanation/overlays.rst
```diff
@@ -0,0 +1,185 @@
+****************
+``OVERLAY`` Step
+****************
+
+It is sometimes necessary to modify the base filesystem used in when generating
```
sergiusens, line 5: > the base filesystem used in when generating

this construct sounds strange to me.

I would also add a ref link to base (if it exists).

docs/explanation/overlays.rst
```diff
@@ -0,0 +1,185 @@
...
+          rm -f ${CRAFT_OVERLAY}/usr/bin/vi ${CRAFT_OVERLAY}/usr/bin/vim*
+          rm -f ${CRAFT_OVERLAY}/usr/bin/emacs*
+          rm -f ${CRAFT_OVERLAY}/bin/nano
+        overlay:
+          - bin
+          - usr/bin
+
+After running this part, the overlay layer (and the final package) will only
+contain ``ed`` as an editor, with vi/vim, emacs, and nano all having been
+removed.
```
sergiusens, line 38: ed is monospaced while the other editors are not, what is the criteria?

docs/explanation/overlays.rst
```diff
@@ -0,0 +1,185 @@
...
+build.
+
+Parts are ordered deterministically for a project, guaranteeing that order will
+be preserved between runs. The current ordering rules are as follows:
+
+#. Parts are ordered alphabetically by name
+#. Any part that is required before another part is moved ahead of it
+
+This means that renaming parts and adding or removing ``after`` sections
+for parts can modify the overlay order.
```
sergiusens, line 79: should this not be under an Ordering title? Visibility seems like a stretch

docs/explanation/overlays.rst
lengau: Yeah that's because it's just all kinds of awkward :-) I've changed it to "modify the base filesystem when generating a parts-based payload," though I'm open to other suggestions for better ways to introduce overlays.

Unfortunately there's no base ref in the craft-parts doc yet and I couldn't see anything in other projects that would be relevant.

tigarmo: maybe we should explicitly call-out rockcraft here? Something like "Some Craft applications, like Rockcraft, include entire base filesystems in addition to the usual part-generated payload. For those applications, it is often desirable to modify this base filesystem..."
that's not _quite_ right, but something along those paths?

docs/explanation/overlays.rst
```diff
@@ -0,0 +1,185 @@
...
+
+Overlay Parameters
+------------------
+
+A part has three parameters that can be used to adjust how the overlay step
+works: ``overlay-packages``, ``overlay-script`` and ``overlay-filter``.
+``overlay-packages`` and ``overlay`` (the filter parameter) behave much the
+same way as the related parameters on the ``STAGE`` step. ``overlay-script``
+likewise behaves similarly to ``override-stage``, including having access to
+the ``partsctl`` command.
```
tigarmo, line 17: do you mean `craftctl`?

docs/explanation/overlays.rst
```diff
@@ -0,0 +1,185 @@
...
+        overlay-script: |
+          echo "${CRAFT_OVERLAY}" > "${CRAFT_PART_INSTALL}/direct"
+      indirect-overlay-access:
+        plugin: nil
+        after: [direct-overlay-access]
+        override-build: |
+          echo "${CRAFT_OVERLAY}" > "${CRAFT_PART_INSTALL}/indirect"
+
+Because ``no-overlay-access`` has no access to the overlay directory, the
+``no-overlay-access`` step will fail, as the ``CRAFT_OVERLAY`` environment
```
tigarmo, line 68:
```suggestion
``override-build`` step will fail, as the ``CRAFT_OVERLAY`` environment
```
or possible `build`, as you refer to it as a "step"

docs/explanation/overlays.rst
lengau: I've moved this over to an expanded part processing order section in the lifecycle explanation, adding a couple of examples. It's now xref'd in the Layers section below.

docs/explanation/overlays.rst
lengau: The criteria comes from me writing the two halves of the sentence 10 minutes apart. Fixed!

docs/explanation/overlays.rst
```diff
@@ -0,0 +1,176 @@
...
+
+Each part has an overlay layer, which acts on a shared storage area in the
+processing order of parts. If the part doesn't specify any overlay parameters,
+this overlay is empty. The overlay's integrity is checked with a checksum
+defined by the following diagram:
+
+.. image:: /images/overlay_checksum.svg
+   :alt: Diagram for generating the overlay checksum.
+
+Each layer combines its parameters with the checksum for the layer below it
```
tigarmo, line 85: This sentence is a little awkward with the many "checksums". Here's a suggestion: `Each layer's checksum is the combination of the layer's properties and the checksum of the layer below it.`

docs/explanation/overlays.rst
```diff
@@ -0,0 +1,176 @@
...
+and checksums that in order to form its own checksum. The base layer is
+checksummed to feed the layer for the first part. As a result, a change to
+any layer will require the recalculation of the overlay for all layers
+above. The order of layers is determined by the :ref:`part_processing_order`.
+
+Filesystem Mutations
+====================
+
+Mutations to the filesystem may include changes caused by Installation of OS packages (including package database updates), the execution of user scripts in the overlay filesystem context, or file filters.
```
tigarmo, line 94: Long line here

tigarmo, line 94:
```suggestion
Mutations to the filesystem may include changes caused by the installation of OS packages (including package database updates), the execution of user scripts in the overlay filesystem context, or file filters.
```

docs/explanation/overlays.rst
```diff
@@ -0,0 +1,176 @@
...
+
+The outcome of the overlay step for each part includes solely the modifications
+made by that part to underlying layers.. As a consequence, if no modifications
```
tigarmo, line 97:
```suggestion
made by that part to underlying layers. As a consequence, if no modifications
```

docs/explanation/overlays.rst
```diff
@@ -0,0 +1,176 @@
...
+are made, the result of the overlay step is empty and the result is the same as
+the four-step lifecycle without overlays. Subtractive changes such as file
+removals are allowed and handled through special whiteout files conforming to
+the `OCI image layer specification <oci_image_layers>`_.
```
tigarmo, line 101: :+1:

docs/explanation/overlays.rst
```diff
@@ -0,0 +1,176 @@
...
+   script if defined.
+#. Disable the mechanism that handles filesystem layering.
+#. Generate the overlay step state.
+
+Staging Overlay Files
+=====================
+
+When executing the stage step for a part that declares overlay content, the
+consolidated content generated in the overlay step is added to the common stage
+area along with artifacts resulting from the part build step. Files from overlay
```
tigarmo, line 127:
```suggestion
area along with artifacts resulting from the part's build step. Files from overlay
```

tigarmo (dismissed):
nice content!

docs/explanation/overlays.rst
```diff
@@ -0,0 +1,176 @@
+****************
+``OVERLAY`` Step
+****************
+
+It is sometimes necessary to modify the base filesystem when generating a
```
tigarmo, line 5: I guess what I'm missing here is some kind of definition of "base filesystem" here. Maybe a distinction between this "base" filesystem and the build filesystem where the lifecycle runs?

docs/explanation/overlays.rst
lengau: Good catch

docs/explanation/overlays.rst
lengau: I've incorporated your suggestion from Sergio's comment above, which should fix both :-)

docs/explanation/overlays.rst
lengau: *Content blatantly ripped from Claudio's specification document*

docs/explanation/overlays.rst
lengau: That's a significant improvement. I slightly adjusted your suggestion and then improved the rest of the paragraph too.

docs/explanation/overlays.rst
lengau: I like that — it anchors the reason for the overlay step to a specific real-world use case so the reader can more quickly decide whether the overlay step is relevant to them.

sergiusens (approved):

PR #371, opened 2023-02-08
packages: MVP for a yum-based RPMRepository

facundobatista:
CRAFT-1581

### DIFF
--- a/craft_parts/packages/__init__.py
+++ b/craft_parts/packages/__init__.py
@@ -20,7 +20,7 @@ from typing import TYPE_CHECKING, Type
 from . import errors  # noqa: F401
 from . import snaps  # noqa: F401
 from .normalize import fix_pkg_config  # noqa: F401
-from .platform import is_deb_based, is_rpm_based
+from .platform import is_deb_based, is_yum_based
 
 if TYPE_CHECKING:
     from .base import BaseRepository
@@ -34,10 +34,10 @@ def _get_repository_for_platform() -> Type["BaseRepository"]:
 
         return Ubuntu
 
-    if is_rpm_based():
-        from .rpm import RPMRepository
+    if is_yum_based():
+        from .yum import YUMRepository
 
-        return RPMRepository
+        return YUMRepository
 
     from .base import DummyRepository
 
--- a/craft_parts/packages/__init__.py
+++ b/craft_parts/packages/__init__.py
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright 2017-2021 Canonical Ltd.
+# Copyright 2017-2023 Canonical Ltd.
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
@@ -20,7 +20,7 @@ from typing import TYPE_CHECKING, Type
 from . import errors  # noqa: F401
 from . import snaps  # noqa: F401
 from .normalize import fix_pkg_config  # noqa: F401
-from .platform import is_deb_based
+from .platform import is_deb_based, is_rpm_based
 
 if TYPE_CHECKING:
     from .base import BaseRepository
@@ -33,6 +33,11 @@ def _get_repository_for_platform() -> Type["BaseRepository"]:
         from .deb import Ubuntu
 
         return Ubuntu
+
+    if is_rpm_based():
+        from .rpm import RPMRepository
+
+        return RPMRepository
 
     from .base import DummyRepository
 
--- a/craft_parts/packages/platform.py
+++ b/craft_parts/packages/platform.py
@@ -15,7 +15,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 """Helpers to determine the repository for the platform."""
-from typing import Optional, List
+from typing import List, Optional
 
 from craft_parts import errors
 from craft_parts.utils import os_utils
@@ -24,10 +24,10 @@ _DEB_BASED_PLATFORM = ["ubuntu", "debian", "elementary OS", "elementary", "neon"
 _RPM_BASED_PLATFORM = ["fedora", "centos"]
 
 
-def _check(distro: str, platform_distros: List[str]) -> bool:
+def _check(distro: Optional[str], platform_distros: List[str]) -> bool:
     """Check if the indicated distro is included in the specified platform distros.
 
-    If the indicated `distro` is None it will be retrived from OsRelease
+    If the indicated `distro` is None it will be retrieved from OsRelease
     ("unknown" on error).
     """
     if not distro:
--- a/craft_parts/packages/platform.py
+++ b/craft_parts/packages/platform.py
@@ -25,10 +25,10 @@ _RPM_BASED_PLATFORM = ["fedora", "centos"]
 
 
 def _check(distro: Optional[str], platform_distros: List[str]) -> bool:
-    """Check if the indicated distro is included in the specified platform distros.
+    """Check if `distro` is included in the specified platform distros.
 
-    If the indicated `distro` is None it will be retrieved from OsRelease
-    ("unknown" on error).
+    If the indicated `distro` is None it will be retrieved from OsRelease or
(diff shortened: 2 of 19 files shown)


craft_parts/packages/rpm.py
```diff
@@ -0,0 +1,200 @@
...
+    @classmethod
+    @functools.lru_cache(maxsize=1)
+    def refresh_packages_list(cls) -> None:
+        """Refresh the list of packages available in the repository."""
+        # Return early when integration testing
+        if os.getenv("CRAFT_PARTS_PACKAGE_REFRESH", "1") == "0":
+            return
+
+        try:
+            cmd = ["yum", "update", "-y"]
```
tigarmo, line 89: I guess `dnf` doesn't come installed by default in centos 7?

craft_parts/packages/rpm.py
```diff
@@ -0,0 +1,200 @@
...
+
+from craft_parts.utils import os_utils
+
+from . import errors
+from .base import BaseRepository
+
+logger = logging.getLogger(__name__)
+
+
+class RPMRepository(BaseRepository):
```
tigarmo, line 31: I know this is an MVP but can we test this class somehow? How about, for example, an integrated test that mocks the OsRelease and the `yum` calls and then does a full lifecycle with a `build-package`? Would it work?

craft_parts/packages/rpm.py
sergiusens: it does not

craft_parts/packages/__init__.py
```diff
@@ -34,6 +34,11 @@ def _get_repository_for_platform() -> Type["BaseRepository"]:
 
         return Ubuntu
 
+    if is_rpm_based():
+        from .rpm import RPMRepository
+
+        return RPMRepository
```
sergiusens, line 40: I would change this to:

```suggestion
    if is_yum_based():
        from .yum import YUMRepository

        return YUMRepository
```

craft_parts/packages/platform.py
```diff
@@ -15,24 +15,44 @@
...
 
 """Helpers to determine the repository for the platform."""
-from typing import Optional
+from typing import List, Optional
 
 from craft_parts import errors
 from craft_parts.utils import os_utils
 
 _DEB_BASED_PLATFORM = ["ubuntu", "debian", "elementary OS", "elementary", "neon"]
+_RPM_BASED_PLATFORM = ["fedora", "centos"]
```
sergiusens, line 24: Fedora today only support DNF, so I would remove it for now

craft_parts/packages/platform.py
```diff
@@ -15,24 +15,44 @@
...
-def is_deb_based(distro: Optional[str] = None) -> bool:
-    """Verify the distribution packaging system.
-
-    :param distro: The distribution name.
+def _check(distro: Optional[str], platform_distros: List[str]) -> bool:
+    """Check if the indicated distro is included in the specified platform distros.
 
-    :return: Whether the distribution uses .deb packages.
+    If the indicated `distro` is None it will be retrieved from OsRelease
+    ("unknown" on error).
```
sergiusens, line 31:
```suggestion
    If `distro` is None it will be retrieved from OsRelease or return "unknown" on error.
```

craft_parts/packages/platform.py
```diff
@@ -15,24 +15,44 @@
...
 _DEB_BASED_PLATFORM = ["ubuntu", "debian", "elementary OS", "elementary", "neon"]
+_RPM_BASED_PLATFORM = ["fedora", "centos"]
 
 
-def is_deb_based(distro: Optional[str] = None) -> bool:
-    """Verify the distribution packaging system.
-
-    :param distro: The distribution name.
+def _check(distro: Optional[str], platform_distros: List[str]) -> bool:
+    """Check if the indicated distro is included in the specified platform distros.
```
sergiusens, line 28:
```suggestion
    """Check if `distro` is included in the specified platform distros.
```

sergiusens (commented):
Looks good, but I would change RPM to DNF, it is mostly a happy coincidence that Debian and derivatives use apt as the resolver; in retrospect, the is_deb_based, should have been is_apt_based since this would have supported the in no longer existing Conectiva Linux :-)

craft_parts/packages/rpm.py
facundobatista: No, and when I read a little about it it was a little hacky to install, so I just moved forward with `yum`.

craft_parts/packages/rpm.py
```diff
@@ -0,0 +1,200 @@
...
+
+    @classmethod
+    def _check_if_all_packages_installed(cls, package_names: List[str]) -> bool:
+        """Check if all given packages are installed.
+
+        Will check versions if using <pkg_name>=<pkg_version> syntax parsed by
+        get_pkg_name_parts().  Used as an optimization to skip installation
+        and cache refresh if dependencies are already satisfied.
+
+        :return True if _all_ packages are installed (with correct versions).
```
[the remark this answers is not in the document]
facundobatista: good catch! fixed!

craft_parts/packages/rpm.py
facundobatista: FTR, decided to maybe do this in a future (not for this PR).

craft_parts/packages/platform.py
facundobatista: Improved.

craft_parts/packages/platform.py
facundobatista: Improved.

facundobatista:
> Looks good, but I would change RPM to DNF,

You mean "change RPM to YUM", right?

tests/unit/packages/test_platform.py
```diff
@@ -63,8 +58,35 @@ def test_is_deb_based_error(mocker):
...
+
+def test_is_rpm_based():
+    for distro in _all_distros + ["other"]:
+        assert platform.is_rpm_based(distro) == (distro in _rpm_distros)
+
+
+def test_is_rpm_based_default(mocker):
+    for distro in _all_distros:
+        mocker.patch("craft_parts.utils.os_utils.OsRelease.id", return_value=distro)
+        assert platform.is_rpm_based() == (distro in _rpm_distros)
```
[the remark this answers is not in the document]
lengau: (I'm also fine with leaving this the way it is right now for consistency with the deb ones)

craft_parts/packages/rpm.py
```diff
@@ -0,0 +1,200 @@
...
+        package_names = sorted(package_names)
+        logger.debug("Requested packages: %s", package_names)
+
+        install_required = False
+        if not cls._check_if_all_packages_installed(package_names):
+            install_required = True
+
+        if not list_only:
+            if refresh_package_cache and install_required:
+                cls.refresh_packages_list()
```
lengau, line 148: Related to my comment above, I'm not sure refreshing the packages list explicitly is necessary here.

lengau (commented):
This is great!

craft_parts/packages/platform.py
facundobatista: > > Looks good, but I would change RPM to DNF,
> 
> You mean "change RPM to YUM", right?

Confirmed through chat. Rename done.

craft_parts/packages/rpm.py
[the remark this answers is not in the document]
facundobatista: You are right! Something confused me while exploring centos, then. I remember this happening to me: not being able to install something, then doing `yum update`, then later being able to install it ok. But something else should have been wrong there.

So I converted this operation into a NOOP and documented it properly.

craft_parts/packages/rpm.py
facundobatista: Right! Also added a comment about it.

tests/unit/packages/test_platform.py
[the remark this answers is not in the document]
facundobatista: (exactly my thought... I was tempted to refactor the other first, but this branch was big already)

sergiusens (approved):

PR #372, opened 2023-02-09
plugins: check if plugin properties are dirty

cmatsuoka:
Add plugin properties to the part state and allow the plugin to list properties to be used to verify if the pull or build steps are dirty.

This implementation ensures compatibility with the current plugin API and doesn't require changes in existing plugins. By default, all plugin properties will be verified against the previous build step state. This can be overridden by each plugin if needed.

The implementation is also designed to work with strict plugins where plugin properties changes may require a repull. In this case, each plugin must return the list of properties relevant to the pull step according to the offline build mode.

-----

### DIFF
--- a/tests/unit/state_manager/test_build_state.py
+++ b/tests/unit/state_manager/test_build_state.py
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright 2021 Canonical Ltd.
+# Copyright 2021-2023 Canonical Ltd.
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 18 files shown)


tests/unit/state_manager/test_build_state.py
```diff
@@ -117,3 +117,31 @@ def test_property_changes(self, properties):
...
+        for prop in augmented_properties.keys():
+            other = augmented_properties.copy()
+            other[prop] = "new value"
+
+            diff = state.diff_properties_of_interest(
+                other, also_compare=["extra-property"]
+            )
+            if prop in relevant_properties:
+                # relevant project options changed
+                assert diff == {prop}
```
lengau, line 144: This seems like a case where checking the full difference and providing all failures could lead to a more obvious resolution of any bugs that appear rather than failing on the first problem.

sergiusens (approved):

lengau (approved):

tests/unit/state_manager/test_build_state.py
cmatsuoka: In lifecycle context all dirty properties are listed, but the test checks property changes individually to make sure each one can be caught separately. We can add more tests covering different scenarios.

PR #232, opened 2022-05-29
sources: add 7z source handler

cmatsuoka:
Allow parts to specify 7z sources. Based on the implementation
from Snapcraft.

Co-authored-by: Tim Süberkrüb <dev@timsueberkrueb.io>
Co-authored-by: Kyle Fazzari <kyrofa@ubuntu.com>
Co-authored-by: Sergio Schvezov <sergio.schvezov@canonical.com>
Signed-off-by: Claudio Matsuoka <claudio.matsuoka@canonical.com>

- [x] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?

-----

PR #377, opened 2023-02-23
sources: add an rpm source handler

lengau:

-----

CRAFT-1583

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -95,6 +95,8 @@ jobs:
           # Remove newer go and install regular version for 20.04
           sudo snap remove go
           sudo apt install -y golang
+          # Install RPM dependencies for RPM tests
+          sudo apt install rpm
       - name: specify node version
         uses: actions/setup-node@v3
         with:
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -97,6 +97,13 @@ jobs:
           sudo apt install -y golang
           # Install RPM dependencies for RPM tests
           sudo apt install rpm
+          # Ensure we don't have dotnet installed, to properly test dotnet-deps
+          # Based on https://github.com/actions/runner-images/blob/main/images/linux/scripts/installers/dotnetcore-sdk.sh
+          sudo apt remove -y dotnet-* || true
+          # Remove manually-installed dotnet from tarballs
+          sudo rm -rf /usr/share/dotnet
+          # Remove dotnet tools
+          rm -rf $HOME/.dotnet
       - name: specify node version
         uses: actions/setup-node@v3
         with:
--- a/craft_parts/sources/errors.py
+++ b/craft_parts/sources/errors.py
@@ -58,6 +58,27 @@ class InvalidSourceOption(SourceError):
         super().__init__(brief=brief, resolution=resolution)
 
 
+# TODO: Merge this with InvalidSourceOption above
+class InvalidSourceOptions(SourceError):
+    """A source option is not allowed for the given source type.
+
+    :param source_type: The part's source type.
+    :param options: The invalid source options.
+    """
+
+    def __init__(self, *, source_type: str, options: List[str]):
+        self.source_type = source_type
+        self.options = options
+        humanized_options = formatting_utils.humanize_list(options, "and")
+        brief = (
+            f"Failed to pull source: {humanized_options} cannot be used "
+            f"with a {source_type} source."
+        )
+        resolution = "Make sure sources are correctly specified."
+
+        super().__init__(brief=brief, resolution=resolution)
+
+
 class IncompatibleSourceOptions(SourceError):
     """Source specified options that cannot be used at the same time.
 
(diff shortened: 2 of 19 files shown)


craft_parts/sources/rpm_source.py
```diff
@@ -0,0 +1,115 @@
...
+    @override
+    def provision(
+        self,
+        dst: Path,
+        keep: bool = False,
+        src: Optional[Path] = None,
+    ) -> None:
+        """Extract rpm file contents to the part source dir."""
+        rpm_path = src or self.part_src_dir / os.path.basename(self.source)
+        command = ["rpm2archive", "-"]
```
cmatsuoka, line 102: Any special reason to choose rpm2archive instead of rpm2cpio? The latter is probably faster/lighter given the data format used in the rpm file.

lengau: Good question, and one I debated internally given precisely the fact that current RPMs are typically CPIO-based. 

In the end I decided on it for future-proofing and compatibility reasons. While the default format with rpm v4 is still a CPIO file, there are cases ([example](https://kojipkgs.fedoraproject.org//packages/python-graph-tool/2.29/2.fc32/aarch64/python3-graph-tool-2.29-2.fc32.aarch64.rpm)) where rpm2cpio fails and suggests rpm2archive intsead. Additionally, the rpm v6 spec looks as though it's not going to be compatible with rpm2cpio.

My initial version included the `--nocompression` parameter for rpm2archive, but alas that was only added in rpm 4.17 and focal only has 4.14. I considered adding detection logic (`rpm2cpio`, fall back to `rpm2archive --nocompression`, finally fall back to `rpm2archive`), but decided the extra complexity probably wasn't worth the performance gain.

craft_parts/sources/errors.py
```diff
@@ -58,6 +58,27 @@ def __init__(self, *, source_type: str, option: str):
         super().__init__(brief=brief, resolution=resolution)
 
 
+# TODO: Merge this with InvalidSourceOption above
```
tigarmo, line 61: do you intend to do this on this PR?

tests/integration/sources/test_rpm.py
```diff
@@ -0,0 +1,110 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2022 Canonical Ltd.
```
tigarmo, line 3:
```suggestion
# Copyright 2023 Canonical Ltd.
```

tests/integration/sources/test_rpm.py
```diff
@@ -0,0 +1,110 @@
...
+    ]
+
+    subprocess.run(
+        ["rpmbuild", "-bb", "--verbose"] + rpmbuild_params + [str(spec_file)],
+        check=True,
+        text=True,
+        capture_output=True,
+    )
+
+    # The asterisks in this glob allow this to be
```
tigarmo, line 79: which glob?

.github/workflows/tests.yaml
```diff
@@ -95,6 +95,8 @@ jobs:
           # Remove newer go and install regular version for 20.04
           sudo snap remove go
           sudo apt install -y golang
+          # Install RPM dependencies for RPM tests
+          sudo apt install rpm
```
tigarmo, line 99: @cmatsuoka do SourceHandlers have a way to declare dependencies that they need in order to actually handle the sources? For example, how are "7z" files handled?

tigarmo (approved):
nice ;)

tests/integration/sources/test_rpm.py
lengau: The one I ended up removing from the test anyway! Fixed :-)

.github/workflows/tests.yaml
sergiusens: https://github.com/canonical/craft-parts/blob/main/craft_parts/packages/deb.py#L372

lengau: Do we know if `rpm2cpio` is even a valid source type anymore? It doesn't seem to be, but I hesitate to remove it here. (I've otherwise updated things though)

craft_parts/sources/errors.py
lengau: No, because it'd include touching a bunch of things not related to this. I've got another branch open for that though

craft_parts/sources/rpm_source.py
cmatsuoka: It would be interesting to add a comment about that otherwise the next person to look at that code may want to change it to rpm2cpio.

lengau: Good point. Adding!

.github/workflows/tests.yaml
cmatsuoka: 7z sources are currently not handled, but I have an old PR to add it: https://github.com/canonical/craft-parts/pull/232 :cake:

cmatsuoka (approved):
Thanks Alex it looks great. Just add a comment explaining the decision of not using rpm2cpio, otherwise someone could change it in the future.

craft_parts/sources/errors.py
lengau: Well I was going to submit that PR tomorrow, but I came up with this after dinner as what might be a better way to do it generically. LMK what you think: https://github.com/canonical/craft-parts/pull/394/commits/015b8592d4c44f815560e45d5bfddede9f0e5ecc

PR #383, opened 2023-02-23
tests: fix dotnet integration tests

### LINKED ISSUE #380: dotnet integration test not fully working
The dotnet integration test is always succeeding on CI because the CI has dotnet installed by default. When it's not installed, it fails with:

```
craft_parts.errors.PluginEnvironmentValidationError: Environment validation failed for part 'foo': 'dotnet' not found and part 'foo' does not depend on a part named 'dotnet-deps' that would satisfy the dependency.
```

We should ensure that dotnet is not installed on the CI and that the succeeds without dotnet.

lengau:
Fixes #380

-----

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -98,9 +98,10 @@ jobs:
           # Ensure we don't have dotnet installed, to properly test dotnet-deps
           # Based on https://github.com/actions/runner-images/blob/main/images/linux/scripts/installers/dotnetcore-sdk.sh
           sudo apt remove dotnet-* || true
-          while $(which dotnet); do
-            sudo rm -f $(which dotnet)
-          done
+          # Remove manually-installed dotnet from tarballs
+          sudo rm -rf /usr/share/dotnet
+          # Remove dotnet tools
+          rm -rf $HOME/.dotnet
       - name: specify node version
         uses: actions/setup-node@v3
         with:
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -95,6 +95,12 @@ jobs:
           # Remove newer go and install regular version for 20.04
           sudo snap remove go
           sudo apt install -y golang
+          # Ensure we don't have dotnet installed, to properly test dotnet-deps
+          # Based on https://github.com/actions/runner-images/blob/main/images/linux/scripts/installers/dotnetcore-sdk.sh
+          sudo apt remove dotnet-* || true
+          while $(which dotnet); do
+            sudo rm -f $(which dotnet)
+          done
       - name: specify node version
         uses: actions/setup-node@v3
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yaml
```diff
@@ -95,6 +95,12 @@ jobs:
           # Remove newer go and install regular version for 20.04
           sudo snap remove go
           sudo apt install -y golang
+          # Ensure we don't have dotnet installed, to properly test dotnet-deps
+          # Based on https://github.com/actions/runner-images/blob/main/images/linux/scripts/installers/dotnetcore-sdk.sh
+          sudo apt remove dotnet-* || true
+          while $(which dotnet); do
```
tigarmo, line 101: weird, do you know where this "other" dotnet is coming from?

tigarmo (changes requested):

.github/workflows/tests.yaml
lengau: I'm not entirely sure, but it looks like a downloaded/extracted tarball like you'd get from the `setup-dotnet` action. Maybe they're using that tarball to bring in versions that aren't in apt?

tests/integration/plugins/test_dotnet.py
```diff
@@ -79,3 +79,45 @@ def test_dotnet_plugin(new_dir):
...
+def test_dotnet_plugin_no_dotnet(new_dir, mocker):
+    """Test the dotnet plugin while pretending dotnet isn't installed."""
+    real_subprocess_run = subprocess.run
+    mock_method = mocker.patch.object(subprocess, "run")
+
+    # Patch the first call to subprocess.run to throw an exception, mimicking not
+    # having dotnet installed, but allow future calls to succeed.
+    first_call = True
+
+    def patch_first_call(*args, **kwargs):
```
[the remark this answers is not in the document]
lengau: We could, but we'd still need to limit it to the first call since after doing the dotnet install in the `dotnet-deps` part it runs the command again.

.github/workflows/tests.yaml
lengau: Okay, I can't replicate getting a random `dotnet` in places other than where it's extracted to `/usr/share/dotnet` in the installer, so I'm just removing the dotnet tools installed via apt and extracted into `/usr/share/dotnet` in the SDK installer script for the runner. (I'm also removing ~/.dotnet since that'll contain tools that require the SDK and gets added to $PATH)

tests/integration/plugins/test_dotnet.py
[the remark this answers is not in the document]
lengau: Looking at it again, it's probably better if I patch the validator rather than `subprocess.run`, so I've done that instead.

lengau:
@tigarmo this version should be better

PR #384, opened 2023-02-23
plugins: refactor python plugin for subclassing

cmatsuoka:
The location of the Python interpreter executable can depend on application-specific parameters that cannot be determined by a generic parts plugin. To address that, allow applications to subclass and override methods in the Python plugin in order to use custom logic to set the interpreter location.

Co-authored-by: Tiago Nobrega <tiago.nobrega@canonical.com>

-----
CRAFT-1629

### DIFF
--- a/craft_parts/plugins/python_plugin.py
+++ b/craft_parts/plugins/python_plugin.py
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright 2020-2021 Canonical Ltd.
+# Copyright 2020-2023 Canonical Ltd.
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
@@ -18,7 +18,7 @@
 
 import shlex
 from textwrap import dedent
-from typing import Any, Dict, List, Set, cast
+from typing import Any, Dict, List, Optional, Set, cast
 
 from overrides import override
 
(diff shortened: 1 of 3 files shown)


craft_parts/plugins/python_plugin.py
```diff
@@ -154,44 +154,91 @@ def get_build_commands(self) -> List[str]:
...
-            """
+                eval "${{opts_state}}"
+                ln -sf "${{symlink_target}}" "${{PARTS_PYTHON_VENV_INTERP_PATH}}"
+                """
             )
         )
 
         return build_commands
+
+    def _should_remove_symlinks(self) -> bool:
```
lengau, line 219: What's the motivation behind this being a method rather than an attribute or property?

craft_parts/plugins/python_plugin.py
```diff
@@ -154,44 +154,91 @@ def get_build_commands(self) -> List[str]:
...
         # Now fix shebangs.
+        script_interpreter = self._get_script_interpreter()
         build_commands.append(
             dedent(
                 f"""\
-            find "{self._part_info.part_install_dir}" -type f -executable -print0 | xargs -0 \
-                sed -i "1 s|^#\\!${{PARTS_PYTHON_VENV_INTERP_PATH}}.*$|#\\!/usr/bin/env ${{PARTS_PYTHON_INTERPRETER}}|"
-            """
+                find "{self._part_info.part_install_dir}" -type f -executable -print0 | xargs -0 \\
+                    sed -i "1 s|^#\\!${{PARTS_PYTHON_VENV_INTERP_PATH}}.*$|{script_interpreter}|"
```
cmatsuoka, line 162: Is escaping needed for `!` in this case?

craft_parts/plugins/python_plugin.py
cmatsuoka: This allows application-specific subclasses to use complex logic based on part info to decide what to do (instead of a simple yes or no),

craft_parts/plugins/python_plugin.py
tigarmo: (I don't know)

craft_parts/plugins/python_plugin.py
```diff
@@ -154,44 +154,91 @@ def get_build_commands(self) -> List[str]:
...
+    def _get_system_python_interpreter(self) -> Optional[str]:
+        """Obtain the path to the system-provided python interpreter.
+
+        This method can be overridden by application-specific subclasses. It
+        returns the path to the Python that bin/python3 should be symlinked to
+        if Python is not part of the payload.
+        """
+        return '$(readlink -f "$(which "${PARTS_PYTHON_INTERPRETER}")")'
+
+    def _get_script_interpreter(self) -> str:
```
tigarmo, line 238: I'm wondering if we should leave this part specifically (the shebang fix) alone for now. During our last meeting I showed that ugly hack that works for installed snaps but not during the build, so this method is either insufficient or it's going to get way uglier.
If I remember correctly the suggestion was made to add full-on scripts that do the full lookup of the correctly Python to use. So for example, when fixing "black" we would rename it to (say) "_black" and then add new "black" file as a shell script that calls "_black". Of course, this strategy still needs to be validated/prototyped.
what do you think? @cmatsuoka @sergiusens

tigarmo (commented):
Thanks a lot for improving my messy prototype!

craft_parts/plugins/python_plugin.py
cmatsuoka: We could leave the method there and not override it on the Snapcraft-specific plugin for now as it doesn't seem to be causing immediate problems. We can work on it again after reexamining the scenario and possible alternatives.

cmatsuoka:
Rebased on main.

craft_parts/plugins/python_plugin.py
tigarmo: Alright, if you're fine with possibly re-working this again in the future then I'm fine with it too

tigarmo (approved):
thanks!

craft_parts/plugins/python_plugin.py
cmatsuoka: Maybe not needed, maybe it's there for a reason. I'll leave it there for now.

sergiusens (approved):

PR #385, opened 2023-02-24
ci: make the CI run in a standard Ubuntu container

lengau:

-----

The switch to node 18 broke our CI, which is what prompted this.


lengau:
Ughhh nope this would basically end up a ground-up redo

PR #402, opened 2023-03-10
docs: add basic environment setup information

dboddie:

-----

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,10 +26,11 @@ by running the following command:
 ./tools/freeze-requirements.sh
 ```
 
-This creates a virtual environment that can be entered by sourcing the
-`activate` file in a bash shell:
+This creates a set of requirements for a virtual environment. Create and
+enter this environment with these commands:
 
 ```
+python3 -m venv venv
 . venv/bin/activate
 ```
 
--- a/README.md
+++ b/README.md
@@ -17,6 +17,34 @@ Free software: GNU Lesser General Public License v3
 
 https://canonical-craft-parts.readthedocs-hosted.com/en/latest/
 
+# Setting up a development environment
+
+To run tests and build documentation, set up a development environment
+by running the following command:
+
+```
+./tools/freeze-requirements.sh
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,6 +17,34 @@ Free software: GNU Lesser General Public License v3
 
 https://canonical-craft-parts.readthedocs-hosted.com/en/latest/
 
+# Setting up a development environment
+
+To run tests and build documentation, set up a development environment
+by running the following command:
+
+```
+./tools/freeze-requirements.sh
```
cmatsuoka, line 26: Freeze-requirements creates a virtual environment, but it also removes this (temporary) environment at the end after the requirements file is created. Maybe creating a new venv and doing what the github ci workflow does could be a better alternative? wdyt @lengau?

lengau: For now I think what @cmatsuoka says is the best way to go about it. 

In the longer term, using `tox` to do it is probably the cleanest way. Once we bring in the starbase configuration here, we'll have the same tox configuration as [craft-archives](https://github.com/canonical/craft-archives/blob/main/tox.ini#L103), so you'll be able to run `tox run rundocs` to run `sphinx-autobuild` on your documentation files. (You can also activate that same venv by sourcing its `activate` file  or just run `tox exec -e sphinx-build bash` to get into the venv in a fresh shell.) Ideally I think we want automatic management (including updating) of our venvs, which tox will provide, but it's nontrivial to set up the tox config for our existing projects.

cmatsuoka: Just to clarify, what the github workflow does is:
```
sudo apt install -y libapt-pkg-dev
pip install -U https://launchpad.net/ubuntu/+archive/primary/+sourcefiles/python-distutils-extra/2.43/python-distutils-extra_2.43.tar.xz https://launchpad.net/ubuntu/+archive/primary/+sourcefiles/python-apt/2.0.0ubuntu0.20.04.6/python-apt_2.0.0ubuntu0.20.04.6.tar.xz .[dev]
```
So this would be executed after creating the venv, and running the freeze-requirements script won't be necessary.

PR #406, opened 2023-03-16
steps: handle overlay feature in step lists

cmatsuoka:
When the overlay feature is disabled, the lists of next and previous steps for any given step must not list the overlay step.

-----

### DIFF
--- a/tests/unit/executor/test_executor.py
+++ b/tests/unit/executor/test_executor.py
@@ -204,8 +204,6 @@ class TestExecutionContext:
         with ExecutionContext(executor=e):
             assert not mock_mount.called
 
-        Features.reset()
-
     def test_capture_stdout(self, capfd, new_dir):
         def cbf(info):
             print(f"prologue {info.custom}")
--- a/tests/unit/executor/test_executor.py
+++ b/tests/unit/executor/test_executor.py
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright 2021 Canonical Ltd.
+# Copyright 2021-2023 Canonical Ltd.
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
@@ -21,6 +21,7 @@ import pytest
 from craft_parts import callbacks, overlays
 from craft_parts.actions import Action
 from craft_parts.executor import ExecutionContext, Executor
+from craft_parts.features import Features
 from craft_parts.infos import ProjectInfo
 from craft_parts.parts import Part
 from craft_parts.steps import Step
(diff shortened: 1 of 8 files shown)


tests/unit/executor/test_part_handler.py
```diff
@@ -37,6 +38,15 @@
 # pylint: disable=too-many-lines
 
 
+def setup_module():
+    Features.reset()
+    Features(enable_overlay=True)
+
+
+def teardown_module():
+    Features.reset()
```
cmatsuoka, line 47: Part handling should be tested with and without overlay enabled. This will be addressed in CRAFT-1604.

tests/unit/executor/test_executor.py
```diff
@@ -199,6 +203,8 @@ def test_prologue_overlay_packages(self, new_dir, mocker):
         with ExecutionContext(executor=e):
             assert not mock_mount.called
 
+        Features.reset()
```
tigarmo, line 206: this is tricky because if the test fails before this line is reached it will never be called, which might make _other_ tests that come after this fail and that's usually very annoying to debug.

You could wrap the whole thing in a try/finally but I think the `request` fixture would be cleaner:

```python
    def test_prologue_overlay_packages(self, new_dir, mocker, request):
        """Check that the overlay package cache is not touched if the part doesn't have overlay packages"""
        Features.reset()
        Features(enable_overlay=True)
        request.addfinalizer(Features.reset)

File: tests/unit/overlays/__init__.py [hunk 3]
```diff
@@ -0,0 +1,26 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from craft_parts.features import Features
+
+
+def setup_module():
```
tigarmo, line 20: I think when we tackle CRAFT-1604 we should also remove this setup/teardown duplication across many test files

tests/unit/state_manager/test_state_manager.py
```diff
@@ -271,6 +298,8 @@ def test_clean_part(self, new_dir):
         for step in list(Step):
             assert sm._state_db.get(part_name="p1", step=step) is None
 
+        Features.reset()
```
tigarmo, line 301: same comment here re: `Features.reset()`  not being called on test failure

tigarmo (changes requested):
Nice! I'm "Requesting changes" here just because I think we should address the `Features.reset()` thing in tests now, but otherwise looks great

tests/unit/overlays/__init__.py
cmatsuoka: My plan is to have tests that require the overlay feature to be enabled under `tests/unit/features/overlay` and then enable it globally for the entire package. The only exception would be the `tests/unit/overlay` because it has a corresponding package under `craft_parts`.

tests/unit/executor/test_executor.py
cmatsuoka: Done as suggested.

tests/unit/state_manager/test_state_manager.py
cmatsuoka: Changed as suggested.

tigarmo (approved):
nice! Remember to remove the leftover `Features.reset()` calls

PR #407, opened 2023-03-17
features: reorganize overlay feature tests

cmatsuoka:
Subsystems affected by the overlay feature should be tested with and without the feature enabled. Reorganize the unit test layout so that regular tests run with the feature disabled, and again with overlay enabled in `tests/unit/feature/overlay`, including feature-specific tests.

Also fix the part environment setup so that the `CRAFT_OVERLAY` variable is only set when the overlay feature is enabled.

-----
CRAFT-1604


tigarmo (approved):
nice, thanks!

PR #410, opened 2023-03-24
Update filesets documentation

dboddie:

-----

(CRAFT-1613)


cmatsuoka:
Craft Parts currently doesn't support filesets, so the feature can't be documented. We have an idea to change filesets to be more flexible, but it's not implemented yet.

docs/explanation/parts.rst
```diff
@@ -174,6 +176,10 @@ can use the :ref:`after` property to define this relationship. This property
 specifies a list containing the names of parts that it will be built after.
 The parts in the list will be *built and staged* before the part is built.
 
+By default, Craft Parts uses the defined build order to determine which
+parts can be built in parallel. This can be disabled by setting the
+:ref:`disable_parallel` property to ``True``.
+
```
cmatsuoka, line 182: Part building is always serialized, no parts are built in parallel. `disable_parallel` is used to disable parallel builds inside a part, but it's currently not implemented. I think we can remove this paragraph for now.

docs/explanation/filesets.rst
```diff
@@ -0,0 +1,107 @@
...
+Fileset names
+-------------
+
+Internally, Craft Parts uses the ``overlay``, ``stage`` and ``prime`` filesets
+to migrate files from all parts into the corresponding steps. For example,
+the *stage* fileset refers to all the files and directories that will be moved
+into the staging area ready for the *stage* step to be run.
+
+Tools that use Craft Parts can define filesets with arbitrary names and they
+can also expose the feature to users. Names are only limited by the format
```
mr-cal, line 25: > Names are only limited by the format used to describe them.

This took me a minute to comprehend.  Is this saying that there are no naming conventions or restrictions for filesets other than the restrictions of the config file format?  (as in yaml's syntax rules when filesets are defined in a snapcraft.yaml file?)

Pedantic: I would reword this to be more to-the-point.  Something like "there are no restrictions on fileset names."  Your statement is true, but it's true for any data written in any format; it's distracting in this context.

mr-cal (approved):
Nice work!

docs/explanation/filesets.rst
```diff
@@ -0,0 +1,107 @@
...
+
+Fileset names
+-------------
+
+Internally, Craft Parts uses the ``overlay``, ``stage`` and ``prime`` filesets
+to migrate files from all parts into the corresponding steps. For example,
+the *stage* fileset refers to all the files and directories that will be moved
+into the staging area ready for the *stage* step to be run.
+
+Tools that use Craft Parts can define filesets with arbitrary names and they
```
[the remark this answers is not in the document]
sergiusens: this is fine as it is not talking about the keyword itself

docs/explanation/parts.rst
sergiusens, line 182:
```suggestion
```

docs/explanation/parts.rst
```diff
@@ -173,7 +175,6 @@ If a part depends on other parts in a project as build dependencies then it
 can use the :ref:`after` property to define this relationship. This property
 specifies a list containing the names of parts that it will be built after.
 The parts in the list will be *built and staged* before the part is built.
```
sergiusens, line 177:
```suggestion
The parts in the list will be *built and staged* before the part is built.

```

sergiusens (approved):

PR #415, opened 2023-04-06
tests: cast type expectation for organize assertions

sergiusens:

-----

### DIFF
--- a/tests/unit/executor/test_organize.py
+++ b/tests/unit/executor/test_organize.py
@@ -17,7 +17,7 @@
 import os
 import re
 from pathlib import Path
-from typing import Any, List
+from typing import Any, List, Tuple, cast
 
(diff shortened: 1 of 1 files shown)


tests/unit/executor/test_organize.py
```diff
@@ -220,6 +220,7 @@ def _organize_and_assert(
             base_dir=base_dir,
             overwrite=overwrite,
         )
+        expected = cast(List[Tuple[List[str], str]], expected)
```
cmatsuoka, line 223: It seems that `expected` declared as `List[Any]` in the function declaration is also bad because it can be an exception too?

cmatsuoka (approved):
This should solve problems for now, but we can refactor this test later, it looks like a port from a different testing framework.

PR #418, opened 2023-04-13
docs: avoid warnings by using autosummary instead of apidoc

dboddie:

-----

### DIFF
--- a/docs/reference/reference.rst
+++ b/docs/reference/reference.rst
@@ -2,6 +2,8 @@
 Package reference
 *****************
 
-.. toctree::
+.. autosummary::
+   :toctree:
+   :recursive:
 
-   gen/craft_parts
+   craft_parts
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -115,5 +115,5 @@ def run_apidoc(_):
     main(["-e", "-o", output_dir, module, "--no-toc", "--force"])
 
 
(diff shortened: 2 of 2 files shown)


docs/reference/reference.rst
```diff
@@ -2,6 +2,8 @@
 Package reference
 *****************
 
-.. toctree::
+.. autosummary::
+   :toctree:
+   :recursive:
 
-   gen/craft_parts
```
tigarmo, line 7: Ah I see now why the new version looks cleaner; `gen/` has the result of apidoc generating an individual rst file from each python module, with the proper `automodule` directives. So with this change the documentation is actually missing a lot.
For example, compare the [current docs](https://canonical-craft-parts.readthedocs-hosted.com/en/latest/reference/gen/craft_parts.plugins.ant_plugin.html) for the ant module with the [new one](https://dboddie-test-org.github.io/craft-parts-docs/reference/craft_parts.plugins.ant_plugin.html#module-craft_parts.plugins.ant_plugin) (the one with these changes).

dboddie: Yes, that's missing a lot. I'll see if there's another way to keep the reference hierarchy while preventing warnings about duplicate documentation.

dboddie:
I'll try a different approach in a new PR.

PR #435, opened 2023-05-02
plugins: add support for pyproject.toml projects (lp:#2008237)

toabctl:
Python projects move to pyproject.toml[0] instead of using setup.py so add support to craft-parts so snaps can be build using the python plugin.

[0]
https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/

-----

### DIFF
--- a/craft_parts/plugins/python_plugin.py
+++ b/craft_parts/plugins/python_plugin.py
@@ -59,7 +59,7 @@ class PythonPlugin(Plugin):
     It can be used for python projects where you would want to do:
 
         - import python modules with a requirements.txt
-        - build a python project that has a setup.py
+        - build a python project that has a setup.py or pyproject.toml
         - install packages straight from pip
 
     This plugin uses the common plugin keywords as well as those for "sources".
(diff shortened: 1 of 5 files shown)


tigarmo (approved):
nice! thank you

toabctl:
@tigarmo how long does it roughly take to get this merged & used by snapcraft?

craft_parts/plugins/python_plugin.py
```diff
@@ -153,7 +153,9 @@ def get_build_commands(self) -> List[str]:
             requirements_cmd = f"{pip} install {constraints} -U {requirements}"
             build_commands.append(requirements_cmd)
 
-        build_commands.append(f"[ -f setup.py ] && {pip} install {constraints} -U .")
+        build_commands.append(
+            f"[ -f setup.py ] || [ -f pyproject.toml ] && {pip} install {constraints} -U ."
```
lengau, line 157:
```suggestion
            f"[ -f setup.py -o -f pyproject.toml ] && {pip} install {constraints} -U ."
```

This is a bit cleaner IMO and allows an easier update to include `setup.cfg` files if we choose.

lengau (approved):

jacobszpz:
Did this change get reverted?

mr-cal:
> Did this change get reverted?

@jacobszpz - Not in the latest releases.  It was mistakenly backported to the 1.19.x release series then reverted because it is a feature, not a bug fix.

jacobszpz:
I see. Thank you

jacobszpz:
> > Did this change get reverted?
> 
> @jacobszpz - Not in the latest releases.  It was mistakenly backported to the 1.19.x release series then reverted because it is a feature, not a bug fix.

I see, thank you

PR #439, opened 2023-05-05
tests: protect fixture against chdir mocks

cmatsuoka:
Always call the original os.chdir in the new_dir fixture to prevent
unexpected behavior if os.chdir is mocked in a test.

-----

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -40,16 +40,24 @@ def pytest_configure(config):
 
 
 @pytest.fixture
-def new_dir(tmpdir):
+def new_dir(monkeypatch, tmpdir):
     """Change to a new temporary directory."""
-
-    original_chdir = os.chdir
-    cwd = os.getcwd()
-    os.chdir(tmpdir)
-
+    monkeypatch.chdir(tmpdir)
     yield tmpdir
 
-    original_chdir(cwd)
+
+@pytest.fixture
+def mock_chdir(monkeypatch):
+    mock_fn = mock.Mock(spec=os.chdir)
+    monkeypatch.setattr(os, "chdir", mock_fn)
+    yield mock_fn
+
+
+@pytest.fixture
+def mock_chroot(monkeypatch):
+    mock_fn = mock.Mock(spec=os.chdir)
+    monkeypatch.setattr(os, "chroot", mock_fn)
+    yield mock_fn
 
 
 @pytest.fixture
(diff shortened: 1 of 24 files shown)


tests/conftest.py
```diff
@@ -43,12 +43,13 @@ def pytest_configure(config):
...
     """Change to a new temporary directory."""
 
+    original_chdir = os.chdir
     cwd = os.getcwd()
     os.chdir(tmpdir)
 
     yield tmpdir
 
-    os.chdir(cwd)
+    original_chdir(cwd)
```
lengau, line 52: This could still fail in weird ways if another fixture mocks `chdir` before this fixture gets started. I'd recommend using [`monkeypatch.chdir`](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.MonkeyPatch.chdir) instead.

lengau (changes requested):

tests/conftest.py
tigarmo: I'm looking at the [source code](https://docs.pytest.org/en/7.1.x/reference/reference.html#_pytest.monkeypatch.MonkeyPatch.chdir) of `monkeypatch.chdir()` (not sure if I'm looking at the right version but, eh) and it looks like it just calls `os.getcwd()`  and `os.chdir()`; does it have any special "magic" to get around the fixture-ordering issue?

lengau: Internal pytest fixtures always get closed last. But also, if we use `monkeypatch` rather than `mocker` to patch things from within `os`, we can be guaranteed that the monkeypatches will be undone before changing directories (see `undo` def [here](https://docs.pytest.org/en/7.1.x/_modules/_pytest/monkeypatch.html#monkeypatch)).

cmatsuoka: I prefer a solution that doesn't require changes in the test itself, because sooner or later someone will write a test mocking os.chdir using standard mock or mocker.

lengau: I understand that desire, but a test that has side-effects outside of its own space should be considered problematic as well.  I would propose that we provide a fixture to nip this ahead of time: https://github.com/cmatsuoka/craft-parts/pull/2/files

lengau (approved):

tests/conftest.py
cmatsuoka: Proposal merged, thanks!

PR #440, opened 2023-05-05
requirements: pin urllib3 version to prevent request problems

cmatsuoka:

-----

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,6 +47,7 @@ install_requires = [
     "pyxdg",
     "requests",
     "requests-unixsocket",
+    "urllib3<2",  # keep compatible API
 ]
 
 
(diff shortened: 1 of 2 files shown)


lengau (approved):
Looks like this is related to `requests_unixsocket` and someone has opened a PR on it: https://github.com/msabramo/requests-unixsocket/pull/69

Could you include that link in the comment so it's easy in future to see why we pinned urllib3 <2.?

syu-w:
Maybe we can stop use that `pyxdg` package?

lengau:
@syu-w probably... The project seems to be pretty inactive on freedesktop.org (https://gitlab.freedesktop.org/xdg/pyxdg), so I'm not sure whether it's considered actively maintained. The lack of response from maintainers on pull requests leads me towards thinking it's not, but I've [bumped the xdg mailing list](https://lists.freedesktop.org/archives/xdg/2023-May/014655.html) about it. 

For what we use it for, implementing the standard directly ourselves probably wouldn't be that bad, but from a "health of the community" perspective I'd rather see the package updated.

cmatsuoka:
Also including the xdg pyright workaround to avoid dependency deadlock.

PR #442, opened 2023-05-09
executor: fix subdir in pull and build steps

cmatsuoka:
The working directories for the pull and build steps should
also be set accordingly when source-subdir is set.

-----

### DIFF
--- a/craft_parts/executor/part_handler.py
+++ b/craft_parts/executor/part_handler.py
@@ -198,7 +198,7 @@ class PartHandler:
         self._run_step(
             step_info=step_info,
             scriptlet_name="override-pull",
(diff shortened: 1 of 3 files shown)


lengau (approved):

tigarmo (approved):
Nice! Do you think this change will "break" anyone? I'm thinking that cases using `source-subdir: some/path` and with something like `override-build: cd some/path ...` (to get around  the bug) will now start failing, but the fix in this case should be straightforward enough

cmatsuoka:
It's always a possibility that someone worked around this bug, but then removing the workaround should be trivial.

sergiusens:
this might have broken Snapcraft

PR #449, opened 2023-05-15
plugins: better checking for needed payload python

tigarmo:
This commit handles the case where the venv symlinks must be removed, but a part-provided Python interpreter *must* be present. This is done by deferring the symlink removal to the end of the build, so that the existing check for a valid interpreter can be performed.

This is to address the "Rockcraft/ubuntu" case in the following table of existing scenarios:
image

That case is particularly bad because the "failure to have the build fail" results in a ROCK that is completely unusable, because of an additional (usrmerge) bug that will be addressed in a separate PR/issue.

CRAFT-1725

### DIFF
--- a/craft_parts/plugins/python_plugin.py
+++ b/craft_parts/plugins/python_plugin.py
@@ -125,20 +125,7 @@ class PythonPlugin(Plugin):
                 """
             )
         )
-
-        # And check if venv symlinks should be removed.
-        if self._should_remove_symlinks():
-            build_commands.append(
-                dedent(
-                    f"""\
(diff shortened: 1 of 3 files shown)


tigarmo:
CI is having a rough day today as the archives are very slow, but I've verified that this fix works for Rockcraft with [a new spread test](https://github.com/canonical/rockcraft/pull/257/files).

cmatsuoka:
@lengau fixed the pyright issue in main, you may want to merge or rebase.

tigarmo:
thanks, done!

cmatsuoka (approved):
This follows the solution discussed in Prague.

PR #455, opened 2023-05-17
deps: make python-apt dependency optional

lengau:

-----

OK, so this is a different mechanism for hopefully making the python-apt dependency easier to deal with.

1. `python-apt` is made an optional dependency with a graceful notification about it.
2. Several optional extras are added to specify a specific package for developer convenience.

Unfortunately, the second would not work without the first, at least not in any way I could figure out, as `python-apt` without a specified location would get an attempted install before the version with the specified location (rather than being deduplicated).

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -51,6 +51,7 @@ package = editable
 extras =
     dev
     test
+    focal-dev
 env_dir = {work_dir}/linting
 runner = ignore_env_name_mismatch
 
@@ -77,7 +78,7 @@ commands =
 description = Static type checking
 base = testenv, lint
 env_dir = {work_dir}/typing
-extras = dev, types
+extras = dev, types, focal-dev
 labels = lint, type
 allowlist_externals =
     mypy: mkdir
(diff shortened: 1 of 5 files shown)


tox.ini
```diff
@@ -77,7 +78,7 @@ commands =
 description = Static type checking
 base = testenv, lint
 env_dir = {work_dir}/typing
-extras = dev, types
+extras = dev, types, focal-dev
```
mr-cal, line 81: With a clean tox environment, I ran `tox run -e lint-pyright` and it seamlessly installed the focal version of python-apt on my 22.10 system.  Is that OK?

If it is ok, then awesome! It just worked :)

lengau: It'll link against your system's libapt, so AFAICT it'll be fine until a backwards-incompatible libapt release, and at that point it should break at the stage of compiling python-apt. But especially for the linters here it really shouldn't matter, since it won't actually execute any of the bound code.

mr-cal (approved):

PR #463, opened 2023-05-24
plugins: qmake

ScarlettGatelyMoore:

-----

I ran the make tests - but I am going to be honest and couldn't get past the ruff test. It wanted me to change a bunch of files that don't belong to me.


craft_parts/plugins/qmake_plugin.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def get_build_packages(self) -> Set[str]:
+        """Return a set of required packages to install in the build environment."""
+        build_packages = {"g++", "make", "qt5-qmake"}
+
+        return build_packages
+
+    def get_build_environment(self) -> Dict[str, str]:
+        """Return a dictionary with the environment to use in the build step."""
+        return {"QT_SELECT": "qt5"}
```
lengau, line 85: What happens currently if someone tries to build a qt6-only project with this plugin? (I'm fine with it failing, I just want to ensure it fails "nicely".)

tests/unit/plugins/test_qmake_plugin.py
```diff
@@ -0,0 +1,106 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2020-2022 Canonical Ltd.
```
lengau, line 3:
```suggestion
# Copyright 2023 Canonical Ltd.
```

Feel free to add yourself or your org to the copyright block too (I'm guessing the copyright is partially Canonical's since I presume that more than the copyright block was copied from elsewhere - correct as needed :-) ) And same note about the copyright owners on the plugin file too.

lengau (commented):
Thanks @ScarlettGatelyMoore, this looks great! I'd like @cmatsuoka to review this too since he's much more familiar with craft_parts than I am, but all my comments are pretty minor/nitpicky.

I've also enabled the CI runners here for tests and stuff.

lengau:
@ScarlettGatelyMoore I'll take a look at the ruff stuff - it's probably due to problems with the configuration I made + ruff not being stable yet

tests/unit/plugins/test_qmake_plugin.py
ScarlettGatelyMoore: That is correct :) I followed a commit for the cmake plugin and converted to qmake logic. I will add myself.

ScarlettGatelyMoore:
> 

Great, Thanks!

craft_parts/plugins/qmake_plugin.py
lengau: I guess what I'm really just looking for is that the error message be something along the lines of "the qmake plugin doesn't currently support qt6" rather than "error on line 4637453 of some_obscure_file.hpp: Missing }". As long as it's fairly clear to a user I'm not too concerned.

ScarlettGatelyMoore: Instead I added support for qmake6!

ScarlettGatelyMoore:
Who do I need to poke to get these workflows approved?

ScarlettGatelyMoore:
> I made a PR to your branch that fixes the linting issues on my machine (and presumably in CI): [ScarlettGatelyMoore#1](https://github.com/ScarlettGatelyMoore/craft-parts/pull/1)
> 
> I think it'll fix the pytest import error too

Hi, I am having issues trying to fix my tests as I still can't make tests locally due ruff errors on files I don't own. The tests on here require maintainer approval and hasn't had approval for days. Please help.

capecrusader-121:
@lengau  @ScarlettGatelyMoore  Any progress on this or any targeted release ?

ScarlettGatelyMoore:
> @lengau @ScarlettGatelyMoore Any progress on this or any targeted release ?

I got stuck and then time got away from me. I will take a look again this week. Or if another set of eyes can take a look, I don't oppose help.

ScarlettGatelyMoore:
I made progress on this. The tests pass now except pyright fails on a bunch of __doc__ type failures and are unrelated to this.

craft_parts/plugins/qmake_plugin.py
```diff
@@ -0,0 +1,145 @@
...
+                'QMAKE_CFLAGS+="${CFLAGS:-}"',
+                'QMAKE_CXXFLAGS+="${CXXFLAGS:-}"',
+                'QMAKE_LFLAGS+="${LDFLAGS:-}"',
+            ] + options.qmake_parameters
+
+        if options.qmake_project_file:
+            qmake_configure_command.extend(
+                [
+                    f'"{self._part_info.part_src_dir}"',
+                    "/",
```
cmatsuoka, line 128: Just curious, what does this command-line parameter mean?

craft_parts/plugins/qmake_plugin.py
```diff
@@ -0,0 +1,145 @@
...
+                    f'"{options.qmake_project_file}"',
+                ]
+            )
+        else:
+            qmake_configure_command.append(f'"{self._part_info.part_src_dir}"')
```
cmatsuoka, line 133: Since this argument is the same in both cases, maybe we could add it before the conditional?

craft_parts/plugins/qmake_plugin.py
```diff
@@ -0,0 +1,145 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright (C) 2023 Canonical Ltd
+# Copyright 2023 Scarlett Moore <sgmoore@kde.org>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
```
cmatsuoka, line 7: Craft-parts plugins should be LGPL.

craft_parts/plugins/qmake_plugin.py
ScarlettGatelyMoore: Sorry, not clear what you are asking here.  I have an optional qmake_project_file: str = "" property so that one can define a non standard .pro file as demonstrated in the test 


    def test_get_build_commands_qmake_project_file(self, setup_method_fixture, new_dir):
        plugin = setup_method_fixture(
            new_dir, properties={"qmake_project_file": "hello.pro"}

cmatsuoka: I asked because this list seems to evaluate to three different command-line arguments (src dir, /, and the project file) and I was wondering what the / could mean here. Or should they assemble a single path such as `srcdir/hello.pro`?

ScarlettGatelyMoore: Correct, assemble a single path

cmatsuoka: But then, when joining with a space in line 137 below, wouldn't that result in three different arguments?

cmatsuoka: One way to address this could be something like:
```
qmake_configure_command.append(str(self._part_info.part_src_dir / options.qmake_project_file))
```
in line 126 so that this specific entry would be a single argument.

ScarlettGatelyMoore: Thank you, a big problem is I can't seem to run make tests locally, ruff tests fail with things way beyond my level of knowledge. Is there a way to only run unit tests?

cmatsuoka: You can run `make test-units` or `make test-integrations` to run all of each kind or `pytest -vv tests/some_file.py` to run a single test.

lengau:
I've made [a PR to your branch](https://github.com/ScarlettGatelyMoore/craft-parts/pull/2) that should fix the relevant issues.

lengau (approved):
LGTM! Thanks for all the work on this!

cmatsuoka (approved):
Thanks Scarlett and Alex! L.et's land this on main and test it -- if everything looks fine we can backport it to 1.x

PR #464, opened 2023-05-25
callbacks: add callback to configure overlay package layer

tigarmo:
This new callback is called by the Executor after the overlay's package cache layer is mounted, but before the package list is actually refreshed. The main purpose is to give applications a chance to configure the Apt installation in the overlay (e.g. add/remove package repositories).

-----

### DIFF
--- a/craft_parts/callbacks.py
+++ b/craft_parts/callbacks.py
@@ -19,6 +19,7 @@
 import itertools
 import logging
 from collections import namedtuple
+from pathlib import Path
 from typing import Callable, Iterable, List, Optional, Set, Union
 
 from craft_parts import errors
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -131,9 +131,9 @@ def fake_snapd():
     socket_path_patcher = mock.patch(
         "craft_parts.packages.snaps.get_snapd_socket_path_template"
     )
+    escaped_path = snapd_fake_socket_path.replace("/", "%2F")
     mock_socket_path = socket_path_patcher.start()
-    mock_socket_path.return_value = f'\
```
tigarmo, line 135: this is unrelated but pyright started complaining about an "unterminated string" here

craft_parts/callbacks.py
```diff
@@ -54,6 +59,21 @@ def register_stage_packages_filter(func: FilterCallback) -> None:
     _STAGE_PACKAGE_FILTERS.append(CallbackHook(func, None))
 
 
+def register_configure_overlay(func: ConfigureOverlayCallback) -> None:
```
tigarmo, line 62: I'm not settled on the callback name (configure_overlay), so if there are better suggestions I'm all ears

craft_parts/callbacks.py
```diff
@@ -54,6 +59,21 @@ def register_stage_packages_filter(func: FilterCallback) -> None:
...
+    """Register a callback function to configure the mounted overlay.
+
+    This "hook" is called after the overlay's package cache layer is mounted, but
+    *before* the package list is refreshed. It can be used to configure the
+    overlay's system, typically to install extra package repositories for Apt.
+    Note that when hook is called the overlay is mounted by *not* chroot'ed into.
```
sergiusens, line 68: can you do a grammar check on this last sentence?

sergiusens (approved):
looks good, please take a look at the documentation comment

cmatsuoka (approved):
+1, but please check the registration function docstring.

craft_parts/callbacks.py
tigarmo: thanks for catching this, looks like my hands and my brain got out of sync when writing that sentence

PR #469, opened 2023-06-02
many: remove todos what won't be done

cmatsuoka:
The codebase contained a number of outdated, unnecessary, or otherwise
impractical to implement TODOs that would most likely never be done.
Remove those, and keep the ones that we will actually work on.

-----

### DIFF
--- a/craft_parts/executor/organize.py
+++ b/craft_parts/executor/organize.py
@@ -60,8 +60,6 @@ def organize_files(
 
             if os.path.isdir(src) and "*" not in key:
                 file_utils.link_or_copy_tree(src, dst)
-                # TODO create alternate organization location to avoid
-                # deletions.
                 shutil.rmtree(src)
                 continue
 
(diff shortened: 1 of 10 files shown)


craft_parts/executor/organize.py
```diff
@@ -60,8 +60,6 @@ def organize_files(
 
             if os.path.isdir(src) and "*" not in key:
                 file_utils.link_or_copy_tree(src, dst)
-                # TODO create alternate organization location to avoid
```
mr-cal, line 63: I was confused by this TODO just this week!

mr-cal (approved):
:+1: I don't have context on all these TODOs but I don't see anything suspicious and I trust your judgment that these are no longer relevant or practical

craft_parts/executor/organize.py
cmatsuoka: Frankly I don't know what this means.

PR #483, opened 2023-06-28
docs: make the part properties reference reusable

dboddie:

-----
CRAFT-1713
Needed for a corresponding PR in Rockcraft.
Documentation: https://canonical-craft-parts--483.com.readthedocs.build/en/483/


docs/base/part_properties.rst
```diff
@@ -0,0 +1,372 @@
+.. Ideally, this would be automatically generated.
+
+.. _after:
+
+after
+-----
+**Type:** array of unique strings with at least 1 item
+**Step:** build
```
sergiusens, line 8: why is there no newline on this one?

docs/base/part_properties.rst
```diff
@@ -0,0 +1,372 @@
...
+
+Specifies a list of parts that a given part will be built *after*.
+
+.. _build_attributes:
+
+build-attributes
+----------------
+**Type:** array of unique strings with at least 1 item from "core22-step-dependencies", "enable-patchelf", "no-patchelf", "no-install", "debug", "keep-execstack".
```
sergiusens, line 16: core22-step-dependencies is only relevant to snapcraft; build-attributes as a whole might only be relevant to snapcraft

docs/base/part_properties.rst
```diff
@@ -0,0 +1,372 @@
...
+
+filesets
+--------
+**Type:** dictionary mapping strings to lists of strings
+
+**Step:** all
+
+Defines named lists of paths to files and directories that can be referred to
+by name in keys that accept lists of paths.
+
```
sergiusens, line 76: this concept only exists in old snapcraft, should not be mentioned

docs/base/part_properties.rst
tigarmo: my suggestion is that we address that separately, because if we remove it from this document we won't have anywhere to put it right?

tigarmo (approved):

docs/base/part_properties.rst
sergiusens: we can put it somewhere once we consume this documentation from Snapcraft, which we don't today, this can confuse our rocks users too

dboddie: I've conditionally excluded this for now.

PR #487, opened 2023-07-04
plugins: be more specific in search for payload Python

tigarmo:
This commit updates the Python plugin to be more restrictive when looking for the interpreter in the parts' payload: if the venv is created using "python3.10", then that exact version of Python must be found in the payload in order to be usable.

In addition to avoiding issues from incompatible versions, this handles cases like the Python 3.6 package (from the deadsnakes ppa) that has both a "python3.6" and a "python3.6m" binaries by default.

-----


tigarmo:
I'm investigating the mypy breakage separately

cmatsuoka (approved):

PR #489, opened 2023-07-05
partitions: add feature and validation

mr-cal:

-----

- The partitions feature can be enabled
- LifecycleManager can accept a new parameter, `partitions`
- The list of partitions is validated and errors are appropriately raised
- Integration tests that only ensures craft-parts works as expected when partitions are enabled (no changes from default behavior)

I tried my best to follow how overlay testing is done (how test directories are laid out, which tests are reused, etc). 

(CRAFT-1843)


tests/unit/features/partitions/test_lifecycle_manager.py
```diff
@@ -0,0 +1,138 @@
...
+                parts_data,
+                application_name="test_manager",
+                cache_dir=new_dir,
+                partitions=partitions,
+            )
+
+        assert str(raised.value) == "First partition must be 'default'."
+
+    @settings(suppress_health_check=[HealthCheck.function_scoped_fixture])
+    @given(
```
mr-cal, line 96: Inspired by https://github.com/canonical/craft-application/pull/20/files#diff-8d3623d9812e6ea2183ad0ed6faa2bcf4ca22b81c744497e4340ef8affe67614 (cc @lengau for review of this test)

craft_parts/features.py
```diff
@@ -27,9 +27,16 @@
 
 @dataclasses.dataclass(frozen=True)
 class Features(metaclass=Singleton):
-    """Configurable craft-parts features."""
+    """Configurable craft-parts features.
+
+    The overlay and partitions features are mutually exclusive.
```
cmatsuoka, line 32: What happens if we try to use them at the same time?

craft_parts/errors.py
```diff
@@ -49,6 +49,17 @@ def __str__(self) -> str:
         return "\n".join(components)
 
 
+class FeatureEnabled(PartsError):
```
cmatsuoka, line 52: Can we consolidate this and `FeatureDisabled` into a single `FeatureError` or similarly named error class?

tests/integration/lifecycle/test_lifecycle_partitions.py
```diff
@@ -0,0 +1,390 @@
...
+        plugin: nil"""
+)
+
+
+@pytest.fixture(autouse=True)
+def setup_feature(enable_partitions_feature):
+    yield
+
+
+def test_basic_lifecycle_actions(new_dir, mocker):
```
cmatsuoka, line 47: If this is the same as the regular lifecycle test with only the partitions parameters changed, maybe we could devise a way to use a common implementation in a future test refactoring round.

tests/integration/lifecycle/test_lifecycle_partitions.py
mr-cal, line 1: This file is a copy/paste of `tests/integration/lifecycle/test_lifecycle.py but, with partitions.

tests/integration/lifecycle/test_lifecycle_partitions.py
```diff
@@ -0,0 +1,390 @@
...
+
+        foo_state_dir = Path("parts/foo/state")
+        bar_state_dir = Path("parts/bar/state")
+
+        self._lifecycle.clean(part_names=["foo"])
+
+        assert Path("parts/foo/src/foo.txt").is_file() is False
+        assert Path("parts/foo/install/foo.txt").is_file() is False
+        assert Path("stage/foo.txt").is_file() is False
+        assert Path("prime/foo.txt").is_file() is False
```
cmatsuoka, line 312: The test file could be inside a partition to reflect the real scenario.

craft_parts/features.py
mr-cal: Theoretically, they should be compatible so I've removed this mutually exclusive requirement.

However there is no use case for enabling both features, so I will not exhaustively test craft-parts with both features enabled.

craft_parts/errors.py
mr-cal: Good idea.  I consolidated these two errors to `FeatureError`

tests/integration/lifecycle/test_lifecycle_partitions.py
mr-cal: +1.  As discussed offline, it isn't worth a grandiose refactoring for two features.  When we add a third feature, we can revisit how to best refactor the tests.

tests/integration/lifecycle/test_lifecycle_partitions.py
mr-cal: +1.  I will move this into a partition in a subsequent PR when more partition functionality is added.

(this entire integration test will grow with upcoming PRs)

tests/unit/features/partitions/test_lifecycle_manager.py
lengau: Looks good to me! I've got my tweezers out for nitpicking, but it's purely informational, so feel free to go

tests/unit/features/partitions/test_lifecycle_manager.py
```diff
@@ -0,0 +1,138 @@
...
+        partitions=strategies.lists(
+            strategies.text(strategies.sampled_from([*ascii_lowercase]), min_size=1),
```
lengau, line 98: `sampled_from` will take any ordered collection, of which a string is one, so you don't technically need to convert it to a list.

Since an empty string is presumably an invalid partition name (`min_size=1` tipped me off), it might be good to test for a partition set of `["default", ""]`

tests/unit/features/partitions/test_lifecycle_manager.py
```diff
@@ -0,0 +1,138 @@
...
+            LifecycleManager(
+                parts_data,
+                application_name="test_manager",
+                cache_dir=new_dir,
+                partitions=partitions,
+            )
+
+        assert str(raised.value) == "First partition must be 'default'."
+
+    @settings(suppress_health_check=[HealthCheck.function_scoped_fixture])
```
lengau, line 95: Ahh yes, function-scoped fixtures. The bane of my hypothesis-usage...

These aren't a problem here for what you're doing here AFAICT, but strong caveat on disabling the health check since it won't detect additional fixtures added later, and then you'll get frustrated about the new fixture you add and why you're getting such weird results on this test, and you'll end up waking up at 2 AM and running to your laptop because you realise the problem.

Overly specific? Personal experience? Nah of course not!

tests/unit/features/partitions/test_lifecycle_manager.py
```diff
@@ -0,0 +1,138 @@
...
+    @given(
+        partitions=strategies.lists(
+            strategies.text(strategies.sampled_from([*ascii_lowercase]), min_size=1),
+            min_size=1,
```
lengau, line 99: If this test ends up taking a long time, it may be worth capping `max_size` to a reasonable value (100?). Hypothesis can and will provide stupidly large lists.

tests/unit/features/partitions/test_lifecycle_manager.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+class TestPartitionsSupport:
+    """Verify LifecycleManager supports partitions."""
+
+    @pytest.fixture
+    def parts_data(self) -> Dict[str, Any]:
+        return {"parts": {"foo": {"plugin": "nil"}}}
+
+    @pytest.mark.parametrize("partitions", [["default"], ["default", "kernel"]])
```
lengau, line 36: How about `["default", "default"]`

tests/unit/features/partitions/test_lifecycle_manager.py
lengau: Also as a side note: This is a fantastic use case for hypothesis!

tests/unit/features/partitions/test_lifecycle_manager.py
lengau: (And as a follow-up, if duplicates aren't allowed, you probably want set your hypothesis strategy below to be unique and filter it to not contain `"default"`

tests/unit/test_errors.py
```diff
@@ -53,8 +53,16 @@ def test_part_dependency_cycle():
     assert err.resolution == "Review the parts definition to remove dependency cycles."
 
 
+def test_feature_enabled():
+    err = errors.FeatureError("bummer")
```
lengau, line 57: :rofl:

tests/unit/features/partitions/test_lifecycle_manager.py
```diff
@@ -0,0 +1,138 @@
...
+        assert info.project_name == "project"
+        assert info.target_arch == "arm64"
+        assert info.arch_triplet == "aarch64-linux-gnu"
+        assert info.parallel_build_count == 16
+        assert info.dirs.parts_dir == new_dir / "work_dir" / "parts"
+        assert info.dirs.stage_dir == new_dir / "work_dir" / "stage"
+        assert info.dirs.prime_dir == new_dir / "work_dir" / "prime"
+        assert info.custom_args == ["custom"]
+        assert info.custom == "foo"
+        assert info.partitions == partitions
```
lengau, line 62: This might be a good opportunity to use `pytest-check`

tests/unit/features/partitions/test_lifecycle_manager.py
cmatsuoka: Also `["default", "kernel", "kernel"]` is a possibility.

cmatsuoka (approved):
Looks good!

tests/unit/features/partitions/test_lifecycle_manager.py
mr-cal: I improved the strategy and added a test for `["default", ""]`

tests/unit/features/partitions/test_lifecycle_manager.py
mr-cal: Ha!   I added a comment explaining why it is safe to suppress the health check here.

tests/unit/features/partitions/test_lifecycle_manager.py
mr-cal: It takes less than half a second on my computer, so I think it's fast enough.

tests/unit/features/partitions/test_lifecycle_manager.py
mr-cal: Great catch.  I added a check in the code and more tests.

tests/unit/features/partitions/test_lifecycle_manager.py
mr-cal: Nice suggestion, added!

tests/unit/features/partitions/test_lifecycle_manager.py
lengau: Thanks! In general I do like to include human-readable explanations whenever I override something (I'm sure you've seen my additional comments when I #noqa something), so thank you for doing so!

lengau (approved):
Great addition, thank you!

PR #493, opened 2023-07-12
partitions: validate usage of partitions in parts

mr-cal:

-----

Validate usage of partitions in filepaths listed in the `organize`, `overlay`, `stage`, and `prime` keywords for parts.

(CRAFT-1844)

### DIFF
--- a/craft_parts/lifecycle_manager.py
+++ b/craft_parts/lifecycle_manager.py
@@ -30,7 +30,7 @@ from craft_parts.dirs import ProjectDirs
 from craft_parts.features import Features
 from craft_parts.infos import ProjectInfo
 from craft_parts.overlays import LayerHash
-from craft_parts.parts import Part, part_by_name
+from craft_parts.parts import Part, part_by_name, validate_partition_usage
 from craft_parts.state_manager import states
 from craft_parts.steps import Step
 
@@ -149,6 +149,9 @@ class LifecycleManager:
             part_list.append(_build_part(name, spec, project_dirs, strict_mode))
 
         self._has_overlay = any(p.has_overlay for p in part_list)
+
+        if partitions:
+            validate_partition_usage(part_list, partitions)
 
         # a base layer is mandatory if overlays are in use
         if self._has_overlay:
--- a/craft_parts/parts.py
+++ b/craft_parts/parts.py
@@ -16,8 +16,9 @@
 
 """Definitions and helpers to handle parts."""
 
+import re
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Sequence, Set
+from typing import Any, Dict, Iterable, List, Optional, Sequence, Set
 
 from pydantic import BaseModel, Field, ValidationError, root_validator, validator
 
@@ -28,6 +29,7 @@ from craft_parts.packages import platform
 from craft_parts.permissions import Permissions
 from craft_parts.plugins.properties import PluginProperties
 from craft_parts.steps import Step
+from craft_parts.utils.formatting_utils import humanize_list
 
 
 class PartSpec(BaseModel):
@@ -320,6 +322,61 @@ class Part:
         """Return whether this part declares overlay content."""
         return self.spec.has_overlay
 
+    def check_partition_usage(self, partitions: List[str]) -> List[str]:
+        """Check if partitions are properly used in a part.
+
+        :param partitions: The list of valid partitions.
+
+        :returns: A list of invalid uses of partitions in the part.
+        """
+        error_list: List[str] = []
+
+        if not Features().enable_partitions:
+            return error_list
+
+        for fileset_name, fileset in [
+            ("overlay", self.spec.overlay_files),
+            # only the destination of organize filepaths use partitions
+            ("organize", self.spec.organize_files.values()),
+            ("stage", self.spec.stage_files),
+            ("prime", self.spec.prime_files),
+        ]:
+            error_list.extend(
+                self._check_partitions_in_filepaths(fileset_name, fileset, partitions)
+            )
+
+        return error_list
+
+    def _check_partitions_in_filepaths(
+        self, fileset_name: str, fileset: Iterable[str], partitions: List[str]
+    ) -> List[str]:
+        """Check if partitions are properly used in a fileset.
+
+        If a filepath begins with a parentheses, then the text inside the parentheses
+        must be a valid partition.
+
+        :param fileset_name: The name of the fileset being checked.
+        :param fileset: The list of filepaths to check.
+        :param partitions: The list of valid partitions.
+
+        :returns: A list of invalid uses of partitions in the fileset.
+        """
+        error_list = []
+
+        for filepath in fileset:
+            match = re.match("^-?\\((?P<partition>.*?)\\)", filepath)
+            if match:
+                partition = match.group("partition")
+                if partition not in partitions:
+                    error_list.append(
+                        f"    unknown partition {partition!r} in {filepath!r}"
+                    )
+
+        if error_list:
+            error_list.insert(0, f"  parts -> {self.name} -> {fileset_name}")
+
+        return error_list
+
 
 def part_by_name(name: str, part_list: List[Part]) -> Part:
     """Obtain the part with the given name from the part list.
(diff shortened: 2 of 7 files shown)


craft_parts/parts.py
```diff
@@ -320,6 +322,61 @@ def has_overlay(self) -> bool:
...
+            match = re.match("^-?\\((?P<partition>.*?)\\)", filepath)
+            if match:
+                partition = match.group("partition")
+                if partition not in partitions:
+                    error_list.append(
+                        f"    unknown partition {partition!r} in {filepath!r}"
+                    )
+
+        if error_list:
+            error_list.insert(0, f"  parts -> {self.name} -> {fileset_name}")
```
cmatsuoka, line 376: IIRC the snapcraft.yaml documentation in the forum uses a dot as a member separator, we may want to be consistent with that in the error message.

mr-cal:
Test failure appears to be unrelated (see #494 )

craft_parts/parts.py
mr-cal: Reworded.  Let me know what you think.

I really would have liked to leverage the [PartSpecificationError](https://github.com/canonical/craft-parts/blob/b7c20f179afeaa8ec8cba88ce5316dce6d26a240/craft_parts/errors.py#L118), but it's designed for a single part, not a list of parts.

tests/unit/features/partitions/test_parts.py
```diff
@@ -0,0 +1,206 @@
...
+from textwrap import dedent
+
+import pytest
+
+from craft_parts import parts
+from craft_parts.parts import Part
+from tests.unit import test_parts
+
+
+class TestPartSpecs(test_parts.TestPartSpecs):
```
tigarmo, line 26: This (and the other "no-op" classes) is testing that the existing behavior is still working if the partitions feature is enabled but the project doesn't use partitions, right? Can you add a comment near here saying that? To remind our future selves.

tests/unit/features/partitions/test_parts.py
```diff
@@ -0,0 +1,206 @@
...
+            "organize": {
+                f"test-{index}": item for index, item in enumerate(valid_filesets)
+            },
+            "stage": valid_filesets,
+            "prime": valid_filesets,
+        }
+
+        part_list = [Part("a", part_data), Part("b", part_data)]
+
+        assert not parts.validate_partition_usage(part_list, partition_list)
```
tigarmo, line 114:
```suggestion
        assert parts.validate_partition_usage(part_list, partition_list) is None
```

This is another nit, feel free to ignore: `not parts.validate_partition_usage(part_list, partition_list)` makes me think that the function is returning `False`, meaning that the validation failed, which is the opposite of the true meaning

tests/unit/features/partitions/test_parts.py
```diff
@@ -0,0 +1,206 @@
...
+        }
+
+        part_list = [Part("part-a", part_data), Part("part-b", part_data)]
+
+        with pytest.raises(ValueError) as raised:
+            parts.validate_partition_usage(part_list, partition_list)
+
+        assert str(raised.value) == dedent(
+            """\
+            Error: Invalid usage of partitions:
```
tigarmo, line 156: nice, very thorough :+1:

tests/unit/features/test_parts.py
```diff
@@ -0,0 +1,203 @@
...
+from textwrap import dedent
+
+import pytest
+
+from craft_parts import parts
+from craft_parts.parts import Part
+
+
+@pytest.mark.usefixtures("enable_all_features")
+class TestPartPartitionUsage:
```
tigarmo, line 26: are these the same tests as in `unit/features/partitions/test_parts.py`?

tigarmo (commented):
very nice ;)

tests/unit/features/partitions/test_parts.py
```diff
@@ -157,59 +154,53 @@ def test_part_invalid_partition_usage_complex(
...
+            - unknown partition '123' in '(123)/test/(default)' in field 'parts.part-b.stage'
+            - unknown partition '123' in '(123)/test/(kernel)' in field 'parts.part-b.stage'
+            - unknown partition '' in '()' in field 'parts.part-b.prime'
+            - unknown partition 'foo' in '(foo)' in field 'parts.part-b.prime'
+            - unknown partition 'bar' in '(bar)/test' in field 'parts.part-b.prime'
+            - unknown partition 'baz' in '(baz)' in field 'parts.part-b.prime'
+            - unknown partition '123' in '(123)/test' in field 'parts.part-b.prime'
+            - unknown partition '123' in '(123)/test/(456)' in field 'parts.part-b.prime'
+            - unknown partition '123' in '(123)/test/(default)' in field 'parts.part-b.prime'
+            - unknown partition '123' in '(123)/test/(kernel)' in field 'parts.part-b.prime'
```
cmatsuoka, line 204: Maybe we could keep the structured view of the previous approach. It seems easier to read, wdyt?

craft_parts/parts.py
mr-cal: Conversation continued [here](https://github.com/canonical/craft-parts/pull/493#discussion_r1263083886)

craft_parts/lifecycle_manager.py
```diff
@@ -150,6 +150,9 @@ def __init__(
 
         self._has_overlay = any(p.has_overlay for p in part_list)
 
+        if partitions:
```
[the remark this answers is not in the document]
mr-cal: Done!

craft_parts/parts.py
```diff
@@ -320,6 +322,59 @@ def has_overlay(self) -> bool:
...
+        :param fileset_name: The name of the fileset being checked.
+        :param fileset: The list of filepaths to check.
+        :param partitions: The list of valid partitions.
+
+        :returns: A list of invalid uses of partitions in the fileset.
+        """
+        error_list = []
+
+        for filepath in fileset:
+            match = re.match("^-?\\((?P<partition>.*?)\\)", filepath)
```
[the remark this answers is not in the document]
mr-cal: Optimized :zap:

craft_parts/parts.py
```diff
@@ -320,6 +322,59 @@ def has_overlay(self) -> bool:
...
+            if match:
```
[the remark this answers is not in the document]
mr-cal: I think it's a good idea.

@cmatsuoka - do you know about 3.8+ support in craft-parts?  (I see 3.7 in [setup.py](https://github.com/canonical/craft-parts/blob/dcdab5926caecd1848b6c3c0db9decf8f8ed02cd/setup.py#L143))

Also, cc @syu-w or @lengau - wasn't there something recently regarding python 3.6 support for charmcraft?  What was that about?

tests/unit/features/partitions/test_lifecycle_manager.py
```diff
@@ -174,3 +180,47 @@ def test_partitions_duplicates(self, new_dir, parts_data, partitions):
             )
 
         assert str(raised.value) == "Partitions must be unique."
+
+    def test_partitions_usage_valid(self, new_dir, partition_list):
```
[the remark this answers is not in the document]
mr-cal: Via the [`__init__` module](https://github.com/canonical/craft-parts/blob/479df38eec7d0208c94d364d5a8bd4889e5bfe94/tests/unit/features/partitions/__init__.py#L20-L22) (the same way it's done for [overlay](https://github.com/canonical/craft-parts/blob/479df38eec7d0208c94d364d5a8bd4889e5bfe94/tests/unit/features/overlay/__init__.py#L20-L22))

tests/unit/features/partitions/test_parts.py
mr-cal: Good idea, I added a comment.

You are correct about everything except for "the partitions feature is enabled but the project doesn't use partitions".  So to be precise (I italicized the small difference):

As you said these no-op classes verify you can enable the partitions feature and create Parts with no change in behavior.  Now, once your application gets to the point of initializing a LifecycleManager and you provide it the list of parts, then the application _must provide a list of partitions to the LifecycleManager when the feature is enabled_.

tests/unit/features/partitions/test_parts.py
mr-cal: Sounds good.  That's a better way to assert that a function that doesn't return anything also does not raise an exception.

tests/unit/features/test_parts.py
mr-cal: Similar except:
1. `overlay` feature enabled
2.  partitions are used and tested in the `overlay` fileset in parts

tests/unit/features/partitions/test_parts.py
mr-cal: Yes, I liked it too.  I went back to the previous structured view but with the dot (.) notation you suggested in an earlier comment.

craft_parts/parts.py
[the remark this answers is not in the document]
syu-w: That one I think is in rockcraft? https://github.com/canonical/rockcraft/pull/311, I am not sure if that is for their program or override *craft all python. I had already use `:= ` in charmcraft.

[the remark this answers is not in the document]
tigarmo: I'm fine with leaving as-is for now

tigarmo (approved):
great! thanks

PR #497, opened 2023-07-14
partitions: add utility to make partition compatible filepaths

mr-cal:

-----

Adds a utility function that takes a filepath and returns a partition-compatible filepath.

(CRAFT-1845)

### DIFF
--- a/craft_parts/utils/partition_utils.py
+++ b/craft_parts/utils/partition_utils.py
@@ -18,7 +18,7 @@
 
 import re
 from pathlib import PurePath, PurePosixPath
-from typing import TypeVar, Union
+from typing import TypeVar, Union, cast
 
 from craft_parts.features import Features
 
@@ -51,8 +51,7 @@ def get_partition_compatible_filepath(filepath: FlexiblePath) -> FlexiblePath:
 
     str_path = str(filepath)
 
-    # Anything that starts with a glob should be assumed to remain that same glob.
-    if str_path.startswith("*"):
+    if str_path == "*":
         return filepath
 
     if _has_partition(str_path):
@@ -67,4 +66,4 @@ def get_partition_compatible_filepath(filepath: FlexiblePath) -> FlexiblePath:
         inner_path = str_path
 
     new_filepath = PurePosixPath(partition, inner_path)
-    return filepath.__class__(new_filepath)
+    return cast(FlexiblePath, filepath.__class__(new_filepath))
--- a/craft_parts/utils/partition_utils.py
+++ b/craft_parts/utils/partition_utils.py
@@ -17,17 +17,23 @@
 """Partition helpers."""
 
 import re
-from pathlib import Path
-from typing import Union
+from pathlib import PurePath, PurePosixPath
+from typing import TypeVar, Union
 
 from craft_parts.features import Features
 
-
-# XXX: for context, you can see where this function in the prototype:
-# https://github.com/mr-cal/craft-parts/commit/e2d15f1bb1da3207463ad3015b68509537b007c3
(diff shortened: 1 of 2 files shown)


craft_parts/utils/partition_utils.py
```diff
@@ -0,0 +1,71 @@
...
+
+    :returns: A filepath that is compatible with the partitions feature.
+    """
+    if not Features().enable_partitions:
+        return filepath
+
+    str_path = str(filepath)
+
+    # Anything that starts with a glob should be assumed to remain that same glob.
+    if str_path.startswith("*"):
```
lengau, line 56: @mr-cal I changed this a bit as I never got a satisfactory regex that met the additional requirements. Please review and lmk whether this is an acceptable alternative.

craft_parts/utils/partition_utils.py
```diff
@@ -0,0 +1,70 @@
...
+        return filepath
```
syu-w, line 56: File path could start with `*/xxx`? What's this used for?

tests/unit/utils/test_partition_utils.py
```diff
@@ -0,0 +1,107 @@
...
+@pytest.mark.parametrize(
+    ("full_path", "expected"),
+    [
+        ("some/path", False),
+        ("(default)", True),
+        ("(default)/", True),
+        ("(part)/some/path", True),
+        ("(nota)partition", False),
+    ],
+)
```
syu-w, line 59: `(not/a)partition`
`(not/a)/partition`
`(NOTA)/partition`
`(not1)/partition`

We really not allowed use number and capital case?

lengau: Correct — we had a conversation over it at https://docs.google.com/document/d/1LeJvrUJAgyoGtbNtjQH6HTFnRMaWxO2Swq_BbvyrhP0/edit?disco=AAAAyZyrkx4

Thanks for the additional test cases! Added!

syu-w: OK then, I think upper case and auto upper case, with number, maybe more better for differentiating partitions. Given that this is a historical legacy that's all for now.

craft_parts/utils/partition_utils.py
lengau: Some of the defaults are strings that are just `"*"`, used for overlay/stage/prime fields. I'd want to double-check with others, but it's feasible that someone might want to prime a file in all partitions.

mr-cal: If I understand correctly, the question is if we should glob partitions (`*/xxx` -> `(default)/xxx, (kernel)/xxx`) or if it should apply the glob inside the default partition (`*/xxx` -> `(default)/*/xxx`).

Both options are fine, but I lean towards the second option (where the glob is applied inside the default partition).  

Two reasons:
1. If filepaths beginning with a glob behave differently, then we'll need some extra documentation (I could see this being confusing).
2. If a user wants the same file in all partitions, they can explicitly list the partitions (`stage: ["(default)/xxx", "(kernel)/xxx"]`)

Perhaps @flotter has an opinion here?

craft_parts/utils/partition_utils.py
mr-cal: Yeah, looking good!  I added comments elsewhere so I'll mark this as resolved.

tests/unit/utils/test_partition_utils.py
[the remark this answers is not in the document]
lengau: This will happen in tests for CRAFT-1846

cmatsuoka (approved):
Looks good!

PR #509, opened 2023-08-02
sequencer: only load project vars in adopting part

cmatsuoka:
Project variables exist only in the part scope and should be ignored
in other parts' states. In the existing implementation these values
are persisted in each part, which leads to double assignment errors
when a step is executed for parts in a order that's different from
the previous execution (see example in LP #1831135).

There are different ways to address this issue, the current fix
tries to be minimally disruptive by just ignoring persisted values
when reloading the state. We can refactor this code later to have
a cleaner implementation.

-----
CRAFT-1088


lengau:
Linting failure should be fixed with #511

tigarmo (approved):

cmatsuoka:
@kubiko Confirms that kernels are building correctly after this change.

PR #511, opened 2023-08-03
deps: freeze pydantic version

lengau:

-----

CI fixed in combination with #512. This is visible in the CI results from #513.


lengau:
Jammy test failure fixed in #512

sergiusens (approved):

PR #516, opened 2023-08-09
sequencer: only load project vars in adopting part (#509)

cmatsuoka:
Project variables exist only in the part scope and should be ignored
in other parts' states. In the existing implementation these values
are persisted in each part, which leads to double assignment errors
when a step is executed for parts in a order that's different from
the previous execution (see example in LP #1831135).

There are different ways to address this issue, the current fix
tries to be minimally disruptive by just ignoring persisted values
when reloading the state. We can refactor this code later to have
a cleaner implementation.

-----

### DIFF
--- a/craft_parts/infos.py
+++ b/craft_parts/infos.py
@@ -174,6 +174,11 @@ class ProjectInfo:
         return self._project_name
 
     @property
+    def project_vars_part_name(self) -> Optional[str]:
+        """Return the name of the part that can set project vars."""
+        return self._project_vars_part_name
+
+    @property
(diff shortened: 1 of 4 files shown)


tigarmo (approved):
This is just a cherry-pick right? LGTM

lengau (approved):
I've made PRs to fix the lint/test failures by doing the same cherry-picks we needed for the 1.19 branch. See #519 and #520

cmatsuoka:
Rebased on top of hotfix/1.23.1

PR #518, opened 2023-08-10
executor: add prologue INFO-level log messages

### LINKED ISSUE #508: Add INFO-level log messages in prologue
### What needs to get done

A lot of things happen during the prologue, and some of them (like installing build packages and snaps) are very time consuming. We should add INFO-level (so, for user-consumption) log messages.

### Why it needs to get done

So that the user has better feedback of what's happening.

tigarmo:
Add informational messages to distinguish when we're installing build-packages, installing build-snaps, updating the overlay base system and fetching stage-packages, as these are potentially time-consuming operations and with the exception of fetching the stage-packages they all happen before we even start executing lifecycle steps.

Fixes #508

-----

### DIFF
--- a/craft_parts/executor/executor.py
+++ b/craft_parts/executor/executor.py
@@ -233,7 +233,7 @@ class Executor:
         if self._extra_build_packages:
             build_packages.update(self._extra_build_packages)
 
-        logger.info("Installing requested build-packages")
+        logger.info("Installing build-packages")
         packages.Repository.install_packages(sorted(build_packages))
 
     def _install_build_snaps(self) -> None:
@@ -257,7 +257,7 @@ class Executor:
                 ", ".join(build_snaps),
             )
         else:
-            logger.info("Installing requested build-snaps")
+            logger.info("Installing build-snaps")
             packages.snaps.install_snaps(build_snaps)
 
     def _verify_plugin_environment(self) -> None:
--- a/craft_parts/executor/executor.py
+++ b/craft_parts/executor/executor.py
@@ -94,6 +94,7 @@ class Executor:
         # update the overlay environment package list to allow installation of
         # overlay packages.
         if any(p.spec.overlay_packages for p in self._part_list):
+            logger.info("Updating base overlay system")
             with overlays.PackageCacheMount(self._overlay_manager) as ctx:
                 callbacks.run_configure_overlay(
                     self._project_info.overlay_mount_dir, self._project_info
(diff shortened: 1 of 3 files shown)


craft_parts/executor/executor.py
```diff
@@ -255,6 +257,7 @@ def _install_build_snaps(self) -> None:
                 ", ".join(build_snaps),
             )
         else:
+            logger.info("Installing requested build-snaps")
```
sergiusens, line 260: Reading this feels a little better to me
```suggestion
        logger.info("Installing build-snaps")
```

Or perhaps required (as they were requested) :man_shrugging:

craft_parts/executor/part_handler.py
```diff
@@ -892,6 +892,7 @@ def _fetch_stage_packages(self, *, step_info: StepInfo) -> Optional[List[str]]:
             return None
 
         try:
+            logger.info("Fetching requested stage-packages")
```
sergiusens, line 895:
```suggestion
            logger.info("Fetching stage-packages")
```

tests/integration/test_logging.py
```diff
@@ -0,0 +1,45 @@
...
+    caplog.set_level(logging.INFO, logger="craft_parts")
+    Path("parts.yaml").write_text(parts_yaml)
+
+    monkeypatch.setattr(sys, "argv", ["build"])
+    monkeypatch.setenv("CRAFT_PARTS_PACKAGE_REFRESH", "0")
+
+    main.main()
+
+    assert "Installing requested build-packages" in caplog.text
+    assert "Fetching requested stage-packages" in caplog.text
```
sergiusens, line 45:
```suggestion
    assert "Installing build-packages" in caplog.text
    assert "Fetching stage-packages" in caplog.text
```

tests/integration/test_logging.py
```diff
@@ -0,0 +1,45 @@
...
+
+from craft_parts import main
+
+parts_yaml = textwrap.dedent(
+    """
+    parts:
+      foo:
+        plugin: nil
+        build-packages: [libc6]
+        stage-packages: [hello]
```
sergiusens, line 29: no build-snaps to not make it slow?

sergiusens: maybe pyright, it will most likely also already be installed :-)

tigarmo: > no build-snaps to not make it slow?

That, and also to be able to run the test locally as a regular user. I'll try pyright

cmatsuoka (approved):
Thanks!

PR #528, opened 2023-08-15
partitions: make partitions optional in ProjectDirs

lengau:

-----

### DIFF
--- a/craft_parts/dirs.py
+++ b/craft_parts/dirs.py
@@ -20,6 +20,8 @@ from pathlib import Path
 from types import MappingProxyType
 from typing import Dict, Optional, Sequence, Union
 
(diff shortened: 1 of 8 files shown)


craft_parts/dirs.py
```diff
@@ -44,6 +46,7 @@ def __init__(
         partitions: Optional[Sequence[str]],
         work_dir: Union[Path, str] = ".",
     ):
+        partition_utils.validate_partition_names(partitions)
```
cmatsuoka, line 49: We're validating it here, but the argument is still required?

lengau:
Linting issues are unrelated and occur with the update from pyright 1.1.322 -> 1.1.323

cmatsuoka (approved):
Thanks!

PR #530, opened 2023-08-16
plugins/rust_plugin: rewrite the plugin to improve UX

liushuyu:

-----

The new plugin will install the Rustup toolchain by default (opt-out possible) and also makes it more intelligent.

Three new options have been added:

- One option for enabling LTO (more optimized binary)
- One option for specifying Rust toolchain channel/version
- One option for disabling default features in the Rust crate

### DIFF
--- a/craft_parts/plugins/rust_plugin.py
+++ b/craft_parts/plugins/rust_plugin.py
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright 2022 Canonical Ltd.
+# Copyright 2022-2023 Canonical Ltd.
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
@@ -14,9 +14,11 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-"""The rust plugin."""
+"""The craft Rust plugin."""
 
 import logging
+import subprocess
+from textwrap import dedent
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, cast
 
 from overrides import override
@@ -30,6 +32,7 @@ logger = logging.getLogger(__name__)
 
 # A workaround for mypy false positives
 # see https://github.com/samuelcolvin/pydantic/issues/975#issuecomment-551147305
+# The proper fix requires Python 3.9+ (needs `typing.Annotated`)
 if TYPE_CHECKING:
     UniqueStrList = List[str]
 else:
@@ -37,12 +40,16 @@ else:
 
 
 class RustPluginProperties(PluginProperties, PluginModel):
-    """The part properties used by the rust plugin."""
+    """The part properties used by the Rust plugin."""
 
     # part properties required by the plugin
     rust_features: UniqueStrList = []
     rust_path: UniqueStrList = ["."]
+    rust_channel: Optional[str] = None
+    rust_use_global_lto: bool = False
+    rust_no_default_features: bool = False
     source: str
+    after: Optional[UniqueStrList] = None
 
     @classmethod
     @override
@@ -56,7 +63,9 @@ class RustPluginProperties(PluginProperties, PluginModel):
         :raise pydantic.ValidationError: If validation fails.
         """
         plugin_data = extract_plugin_properties(
-            data, plugin_name="rust", required=["source"]
+            data,
+            plugin_name="rust",
+            required=["source"],
         )
         return cls(**plugin_data)
 
@@ -76,28 +85,62 @@ class RustPluginEnvironmentValidator(validator.PluginEnvironmentValidator):
 
         :param part_dependencies: A list of the parts this part depends on.
         """
-        for dependency in ["cargo", "rustc"]:
-            self.validate_dependency(
-                dependency=dependency,
-                plugin_name="rust",
-                part_dependencies=part_dependencies,
-            )
+        if "rust-deps" in (part_dependencies or []):
+            options = cast(RustPluginProperties, self._options)
+            if options.rust_channel:
+                raise validator.errors.PluginEnvironmentValidationError(
+                    part_name=self._part_name,
+                    reason="rust-deps can not be used"
+                    "when rust-channel is also specified",
+                )
+            for dependency in ["cargo", "rustc"]:
+                self.validate_dependency(
+                    dependency=dependency,
+                    plugin_name="rust",
+                    part_dependencies=part_dependencies,
+                )
+            options.rust_channel = "none"
 
 
 class RustPlugin(Plugin):
-    """A plugin for rust projects.
+    """A craft plugin for Rust applications.
+
+    This Rust plugin is useful for building Rust based parts.
+
+    Rust uses cargo to drive the build.
 
     This plugin uses the common plugin keywords as well as those for "sources".
     For more information check the 'plugins' topic for the former and the
     'sources' topic for the latter.
 
     Additionally, this plugin uses the following plugin-specific keywords:
+        - rust-channel
+          (string, default "stable")
+          Used to select which Rust channel or version to use.
+          It can be one of "stable", "beta", "nightly" or a version number.
+          If you don't want this plugin to install Rust toolchain for you,
+          you can put "none" for this option.
+
         - rust-features
-          (list of unique strings; default: [])
-          List of rust features to install.
+          (list of strings)
+          Features used to build optional dependencies
+
         - rust-path
-          (list of unique strings; default: ["."])
-          Path of rust project. Currently, only the first path in the list is used.
+          (list of strings, default [.])
+          Build specific crates inside the workspace
+
+        - rust-no-default-features
+          (boolean, default False)
+          Whether to disable the default features in this crate.
+          Equivalent to setting `--no-default-features` on the commandline.
+
+        - rust-use-global-lto
+          (boolean, default False)
+          Whether to use global LTO.
+          This option may significantly impact the build performance but
+          reducing the final binary size.
+          This will forcibly enable LTO for all the crates you specified,
+          regardless of whether you have LTO enabled in the Cargo.toml file
     """
 
     properties_class = RustPluginProperties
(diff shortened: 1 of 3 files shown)


liushuyu:
The CI errors are unrelated (due to new behaviors in the newer version of `pyright`)

craft_parts/plugins/rust_plugin.py
```diff
@@ -111,32 +118,84 @@ def get_build_snaps(self) -> Set[str]:
...
+            rust_install_cmd = self._get_setup_rustup(options.rust_channel)
+        else:
+            logger.info("Switch rustup channel to %s", options.rust_channel)
+            rust_install_cmd = [f"rustup default {options.rust_channel}"]
 
         if options.rust_features:
-            install_command_list.extend(
-                ["--features", f"'{' '.join(options.rust_features)}'"]
+            features_string = " ".join(options.rust_features)
+            config_cmd.extend(["--features", f"'{features_string}'"])
```
lengau, line 165:
```suggestion
            config_cmd.extend(["--features", *options.rust_features])
```

lengau (dismissed):
Thanks for this! It changes the default behaviour in a way that could have surprising results for existing projects, so I've made suggestions about how to handle backwards compatibility. I'm a big fan of the `rustup` integration though

craft_parts/plugins/rust_plugin.py
```diff
@@ -111,32 +118,84 @@ def get_build_snaps(self) -> Set[str]:
...
+        return {"curl", "gcc", "git", "pkg-config", "findutils"}
+
+    def _check_rustup(self) -> bool:
+        try:
+            rustup_version = subprocess.check_output(["rustup", "--version"])
+            return "rustup" in rustup_version.decode("utf-8")
+        except (subprocess.CalledProcessError, FileNotFoundError):
+            return False
+
+    def _get_setup_rustup(self, channel: str) -> List[str]:
```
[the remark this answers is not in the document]
liushuyu: > I'll defer to @cmatsuoka and @mr-cal on this as the maintainer and author, but we may want to default to getting rust from the system package manager.

That's a bad idea. Our in-archive Rust toolchain is usually lagging, and Rust is a very fast-moving language that releases a new compiler version every six weeks.



> To keep backwards compatibility, I would guess the logic could be something like:
> 
>     1. If a `rust-deps` part dependency is defined, no-op (maybe error if a `rust-channel` is defined).
> 
>     2. If a `rust-channel` is defined, use rustup.
> 
>     3. If neither of those, include `rustc` and `cargo` in the build packages.

I chose the current behavior because if you do a quick search on the Snapcraft forum, you will find most users are confused by this (they are migrating from `core18`, which they would expect the plugin to install a `rustup` binary.

craft_parts/plugins/rust_plugin.py
liushuyu: This suggested change does not seem correct.
`cargo` expect all the features to be a string:

`cargo build --features 'a b c'`

instead of:

`cargo build --features a b c`

craft_parts/plugins/rust_plugin.py
```diff
@@ -111,32 +118,84 @@ def get_build_snaps(self) -> Set[str]:
...
-            "--root",
-            '"${CRAFT_PART_INSTALL}"',
-            "--force",
-        ]
+        rust_install_cmd: List[str] = []
+        rust_build_cmd: List[str] = []
+        config_cmd: List[str] = []
+
+        if options.rust_channel == "none":
+            rust_install_cmd = []
```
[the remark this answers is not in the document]
liushuyu: That's a good point. However, I do think if the user chose the update channel instead of a fixed version, they should expect the version to be updated each time the upstream releases a new compiler version.

craft_parts/plugins/rust_plugin.py
```diff
@@ -111,32 +118,84 @@ def get_build_snaps(self) -> Set[str]:
...
+    # virtual workspace is a bit tricky,
+    # we need to build the whole workspace and then copy the binaries ourselves
+    pushd "{crate}"
+    cargo build --workspace --release {config_cmd_string}
+    # install the final binaries
+    # TODO(liushuyu): this will also install proc macros in the workspace,
+    # which the user may not want to keep in the final installation
+    find ./target/release -maxdepth 1 -executable -exec install -Dvm755 {{}} "{self._part_info.part_install_dir}" ';'
+    popd
+fi"""
```
cmatsuoka, line 199: This would look nicer dedented.

liushuyu: Addressed.

craft_parts/plugins/rust_plugin.py
```diff
@@ -14,34 +14,38 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-"""The rust plugin."""
+"""The Snapcraft Rust plugin."""
```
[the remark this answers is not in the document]
liushuyu: Addressed

craft_parts/plugins/rust_plugin.py
```diff
@@ -111,32 +118,84 @@ def get_build_snaps(self) -> Set[str]:
...
+        config_cmd: List[str] = []
+
+        if options.rust_channel == "none":
+            rust_install_cmd = []
+        elif not self._check_rustup():
+            logger.info("Rustup not found, installing it")
+            rust_install_cmd = self._get_setup_rustup(options.rust_channel)
+        else:
+            logger.info("Switch rustup channel to %s", options.rust_channel)
+            rust_install_cmd = [f"rustup default {options.rust_channel}"]
```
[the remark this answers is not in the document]
liushuyu: Rust has a slightly different triplet system than us (and most compilers), so I think a different translation table may need to be used.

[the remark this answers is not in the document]
lengau: Good point — there's also `self._part_info.target_arch` that'll provide the debian architecture name (`deb` in each dict of the mapping) if you'd prefer that to be the source. My naïve thought was to simply replace the first `-` in the triplet with `unknown`, but thinking about it further that neither covers armhf nor is any more reliable than just making a new mapping table.

craft_parts/plugins/rust_plugin.py
lengau: Ahh, you're right — I missed the `'`s. Sorry about that!

craft_parts/plugins/rust_plugin.py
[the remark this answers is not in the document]
lengau: That's a good point. In that case we probably do want to run `rustup update` since I would guess it would be less surprising to get a new version of rust than to come back to a project you haven't touched in a month and get an old version of rust.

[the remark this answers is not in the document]
liushuyu: > That's a good point. In that case we probably do want to run `rustup update` since I would guess it would be less surprising to get a new version of rust than to come back to a project you haven't touched in a month and get an old version of rust.

For Rust, if you are using the `stable` channel (which is the default), it is improbable that the newer compiler would break your project. Even if it does, the Rust compiler usually will provide an option to fix your code automatically.

craft_parts/plugins/rust_plugin.py
```diff
@@ -56,52 +61,55 @@ def unmarshal(cls, data: Dict[str, Any]) -> "RustPluginProperties":
...
-        :param part_dependencies: A list of the parts this part depends on.
-        """
-        for dependency in ["cargo", "rustc"]:
-            self.validate_dependency(
-                dependency=dependency,
-                plugin_name="rust",
-                part_dependencies=part_dependencies,
-            )
+class RustPlugin(Plugin):
+    """A Snapcraft plugin for Rust applications.
```
syu-w, line 72: Not only Snapcraft

craft_parts/plugins/rust_plugin.py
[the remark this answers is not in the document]
syu-w: I would like to allow user to choice use system rust or install from rustup with/without pinned rust versions.

craft_parts/plugins/rust_plugin.py
```diff
@@ -111,32 +118,84 @@ def get_build_snaps(self) -> Set[str]:
...
-            options.rust_path[0],  # TODO: support more than 1 path
-            "--root",
-            '"${CRAFT_PART_INSTALL}"',
-            "--force",
-        ]
+        rust_install_cmd: List[str] = []
+        rust_build_cmd: List[str] = []
+        config_cmd: List[str] = []
+
+        if options.rust_channel == "none":
```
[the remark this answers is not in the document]
syu-w: Not sure if we want auto `cargo update`. We could add an option for that.

craft_parts/plugins/rust_plugin.py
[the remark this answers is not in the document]
lengau: I like the use of rustup (and have dealt with [exactly the lagging rust versions issue](https://github.com/canonical/charmcraft/blob/main/snap/snapcraft.yaml#L74-L94) in charmcraft — though there `focal-updates` getting a new enough rust ended up resolving it). I've also dealt with the confusion of my core22 snap not having rust even when I used the plugin, so I'm all on board with that!

The backwards compatibility issue is within a single `coreX` base rather than between bases. While I personally prefer the core18 behaviour, we can't break existing `core22` snaps (or charms/rocks that use the plugin).

The two major ways people currently use to get rust on core22 are to install it in a `rust-deps` part or to add `cargo` to `build-packages` (as I did with charmcraft, and also like [sccache](https://github.com/mozilla/sccache/blob/main/snap/snapcraft.yaml#L40) and [alacritty](https://github.com/snapcrafters/alacritty/blob/1f478744ea69f1095c5399e65db2888b9d598ebe/snap/snapcraft.yaml#L75), and we need to ensure that neither of these break existing projects.

Perhaps the following behaviour will be more palatable:

1. If a `rust-deps` part dependency is defined, assume that installs rust and error if there's rustup configuration here.
  a) if there is a `rust-deps` part dependency, maybe log the fact that it's overriding the installation so users have a friendly way to know that they need to rename it
2. If `rust-channel` is undefined, check for `rustc` and `cargo` during the pull step. If they exist **and** `rustup` does not, no-op. (This should work since `build-packages` are installed before the pull step).
3. If we don't have `cargo` or `rustc`, do the `rustup` download/update logic.

craft_parts/plugins/rust_plugin.py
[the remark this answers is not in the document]
liushuyu: I think that was a typo, they probably meant `rustup update`.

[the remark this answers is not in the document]
lengau: You're right — I meant to say `cargo fetch`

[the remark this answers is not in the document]
lengau: (Probably with an appropriate `--target`)

craft_parts/plugins/rust_plugin.py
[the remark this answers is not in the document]
lengau: Exactly — and if you're using the `nightly` channel you can (theoretically at least) expect your code to break randomly in CI unless you've set up a `rust-toolchain.toml`.

But in a case like [the alacritty snap](https://github.com/snapcrafters/alacritty/blob/1f478744ea69f1095c5399e65db2888b9d598ebe/snap/snapcraft.yaml), if you have `rust-channel: stable` update the `source-tag`, you don't want it to tell you that rust is out of date and to have to `snapcraft clean` to get it to work — instead, re-pulling should include updating rust.

craft_parts/plugins/rust_plugin.py
liushuyu: Fixed

lengau:
A rebase should now fix the lint issues BTW :smile:

craft_parts/plugins/rust_plugin.py
```diff
@@ -14,36 +14,50 @@
...
 else:
     UniqueStrList = conlist(str, unique_items=True)
 
 
 class RustPluginProperties(PluginProperties, PluginModel):
-    """The part properties used by the rust plugin."""
+    """The part properties used by the Rust plugin."""
 
     # part properties required by the plugin
+    rust_deps: UniqueStrList = []
```
sergiusens, line 45: this is the part, as in
```yaml
parts:
  p1:
    plugin: rust
    source: .
    after: [rust-deps]
  rust-deps:
    plugin: nil
    override-build: |
      rustup...
```

craft_parts/plugins/rust_plugin.py
```diff
@@ -14,36 +14,50 @@
...
+    rust_channel: Optional[str] = None
+    rust_use_global_lto: bool = False
+    rust_no_default_features: bool = False
     source: str
 
+    @validator("rust_channel")
+    @classmethod
+    def check_rust_deps(cls, value: Optional[str], values) -> Optional[str]:
+        """Validate that system Rust and rustup Rust are not used together."""
+        if value and values.get("rust_deps"):
```
sergiusens, line 57:
```yaml
after = values.get("after")
if value and "rust-deps" in after
```

craft_parts/plugins/rust_plugin.py
```diff
@@ -111,32 +128,109 @@ def get_build_snaps(self) -> Set[str]:
...
-        return {"PATH": "${HOME}/.cargo/bin:${PATH}"}
+        return {
+            "PATH": "${HOME}/.cargo/bin:${PATH}",
+        }
+
+    @override
+    def get_pull_commands(self) -> List[str]:
+        """Return a list of commands to run during the pull step."""
+        options = cast(RustPluginProperties, self._options)
+        if not options.rust_channel and self._check_system_rust():
```
sergiusens, line 170: also check if self._part.dependencies has rust-deps

cmatsuoka: `self._part_info` is available to plugins (instead of `self_part`) and it currently doesn't provide dependencies. The information is currently passed only to environment validators. @liushuyu may have an alternative solution for this, otherwise we can add dependency information to `PartInfo`.

liushuyu: I have pushed an alternative solution yesterday:

https://github.com/canonical/craft-parts/pull/530/files#diff-f7845e9499f862b5d9c9c6140de217cea696e6c67f6dfb03cfe1bdf421861486R88-R99

craft_parts/plugins/rust_plugin.py
liushuyu: This is not possible. `after` is not passed to this part.

craft_parts/plugins/rust_plugin.py
liushuyu: Done.

craft_parts/plugins/rust_plugin.py
```diff
@@ -111,32 +153,109 @@ def get_build_snaps(self) -> Set[str]:
...
+            logger.info("Rust is installed on the system, skipping rustup")
+            return []
+
+        rust_channel = options.rust_channel or "stable"
+        if rust_channel == "none":
```
cmatsuoka, line 200: Is "none" a valid channel value or something used for debug? It's not mentioned in the docstring above.

craft_parts/plugins/rust_plugin.py
cmatsuoka: It seems that current solution detects if both the channel and a rust-deps part are available, however if only the rust-deps part is used, rustup is still invoked? `self._check_system_rust()` would fail at this point because the rust-deps part is not staged yet.

liushuyu: > It seems that current solution detects if both the channel and a rust-deps part are available, however if only the rust-deps part is used, rustup is still invoked, `self._check_system_rust()` would fail at this point because the rust-deps part is not staged yet.

Addressed.
Now it will automatically set `"none"` in `validate_environment` if the user has `rust-deps` defined.

cmatsuoka (approved):
Thank you, I think we now cover all the requirements. @sergiusens, @lengau, @mr-cal, @syu-w, is there anything else you want to check?

liushuyu:
CI errors do not seem related to this PR.

craft_parts/plugins/rust_plugin.py
```diff
@@ -111,32 +154,109 @@ def get_build_snaps(self) -> Set[str]:
...
+        return {"curl", "gcc", "git", "pkg-config", "findutils"}
+
+    def _check_system_rust(self) -> bool:
+        """Check if Rust is installed on the system."""
+        try:
+            rust_version = subprocess.check_output(["rustc", "--version"])
+            cargo_version = subprocess.check_output(["cargo", "--version"])
+            return "rustc" in rust_version.decode(
+                "utf-8"
+            ) and "cargo" in cargo_version.decode("utf-8")
```
lengau, line 166: This is a real nitpick, but I think it's more readable to do

```suggestion
            rust_version = subprocess.check_output(["rustc", "--version"], text=True)
            cargo_version = subprocess.check_output(["cargo", "--version"], text=True)
            return "rustc" in rust_version and "cargo" in cargo_version
```

lengau (approved):
This is great, thank you! With some slight adjustments to charmcraft this will also solve most of our rust-dependencies issues in charms.

craft_parts/plugins/rust_plugin.py
liushuyu: Done

sergiusens:
the failing test might be obsolete now and chisel or the chisel DB might now be doing the right thing...
cc @cjdcordeiro for confirmation

cjdcordeiro:
> the failing test might be obsolete now and chisel or the chisel DB might now be doing the right thing... cc @cjdcordeiro for confirmation

If I'm reading it right, the tests are failing on the case where Chisel is `! supported`, and the exception is not being raised. Chisel hasn't changed in quite some time. 

The test seems to be failing for 20.04, which you have marked as `unsupported` in 

https://github.com/canonical/craft-parts/blob/56f25859850f1694ab1d043deaa2712c8b1317c4/tests/integration/lifecycle/test_chisel_lifecycle.py#L32

(it is now supported though! could that be the issue?)

liushuyu:
> > (it is now supported though! could that be the issue?)
> 
> This must be the cause, I'll create a PR to update the test.

I will rebase this pull request should the new pull request fixes the issue.

Saviq:
The documentation could use an update following this:

https://forum.snapcraft.io/t/the-rust-plugin/8588

PR #535, opened 2023-08-25
packages: don't write info logs in overlays

tigarmo:
This commit is a workaround for a current bad interaction between the way we perform overlay operations (a chroot in a multiprocess-created subprocess) and the way craft applications indicate progress on the cli (a spinner in a thread).

The consequence of the bad interaction is that trying to write an info message to the terminal ends up deadlocking the main thread in the subprocess because the thread that could unblock it does not exist.

Instead, log the message as 'debug' instead of 'info', and add a main- process-side 'info' message to communicate that overlay packages are being fetched.

-----

### DIFF
--- a/craft_parts/executor/part_handler.py
+++ b/craft_parts/executor/part_handler.py
@@ -931,6 +931,7 @@ class PartHandler:
 
         try:
             with overlays.PackageCacheMount(self._overlay_manager) as ctx:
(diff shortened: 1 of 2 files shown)


cmatsuoka (approved):
The real fix in craft-cli can be tracked in https://github.com/canonical/craft-cli/issues/172

sergiusens (approved):

PR #557, opened 2023-09-22
chore(deps): update type checkers

lengau:

-----

### DIFF
--- a/craft_parts/parts.py
+++ b/craft_parts/parts.py
@@ -155,7 +155,7 @@ class PartSpec(BaseModel):
             return self.override_build
         if step == Step.STAGE:
             return self.override_stage
-        if step == step.PRIME:
+        if step == Step.PRIME:
             return self.override_prime
 
         raise RuntimeError(f"cannot get scriptlet for invalid step {step!r}")
--- a/craft_parts/plugins/rust_plugin.py
+++ b/craft_parts/plugins/rust_plugin.py
@@ -99,7 +99,7 @@ class RustPluginEnvironmentValidator(validator.PluginEnvironmentValidator):
                     plugin_name="rust",
                     part_dependencies=part_dependencies,
                 )
-            options.rust_channel = "none"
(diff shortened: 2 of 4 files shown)


mr-cal (approved):

craft_parts/parts.py
```diff
@@ -155,7 +155,7 @@ def get_scriptlet(self, step: Step) -> Optional[str]:
             return self.override_build
         if step == Step.STAGE:
             return self.override_stage
-        if step == step.PRIME:
+        if step == Step.PRIME:
```
cmatsuoka, line 158: This is an interesting bug, how did linters let this pass?

craft_parts/plugins/rust_plugin.py
```diff
@@ -99,7 +99,7 @@ def validate_environment(
                     plugin_name="rust",
                     part_dependencies=part_dependencies,
                 )
-            options.rust_channel = "none"
+            options.rust_channel = "none"  # type: ignore[misc]
```
cmatsuoka, line 102: The type is already optional string, why is this necessary?

cmatsuoka (approved):

craft_parts/parts.py
lengau: I'm not entirely sure — I guess because functionally it wasn't a bug (`<instance>.PRIME == <class>.PRIME`, though strictly `<instance>.PRIME is <class>.PRIME.value`)?

craft_parts/plugins/rust_plugin.py
lengau: When I wrote this last week, I thought we had tested the rust plugin to include this, but looking again we haven't.

The complaint was:

```
craft_parts/plugins/rust_plugin.py:103: error: Property "rust_channel" defined in "RustPluginProperties" is read-only  [misc]
```

So this is actually a perfect example of mypy uncovering a bug for us! :tada: 

I've created #559 for that and updated the file with an explanatory comment and a note about removing this ignore comment once we fix it.

PR #560, opened 2023-10-02
rust_plugin: multiple fixes 

### LINKED ISSUE #559: New rust plugin fails to validate with a `rust-deps` part dependency
### Bug Description

We allow a `rust-deps` part dependency in the rust plugin so the user may install rust in the manner they choose.

The updated rust plugin, however, fails to validate when this occurs.

### To Reproduce

Modify the `test_get_pull_commands_compat` test in `test_rust_plugin.py` as follows:

```python
def test_get_pull_commands_compat(part_info):
    properties = RustPlugin.properties_class.unmarshal(
        {"source": ".", "after": ["rust-deps"]}
    )
    plugin = RustPlugin(properties=properties, part_info=part_info)
    plugin.validator_class(
        part_name="project",
        env="",
        properties=properties
    ).validate_environment(part_dependencies=["rust-deps"])
    plugin._check_rustup = lambda: False

    commands = plugin.get_build_commands()
    assert plugin.get_pull_commands() == []
    assert 'cargo install -f --locked --path "."' in commands[0]
```

### part yaml

```shell
parts:
  rust-deps:
    plugin: nil
  project:
    source: .
    after: ["rust-deps"]
```

### Relevant log output

```shell
FAILED   [100%]
tests/unit/plugins/test_rust_plugin.py:184 (test_get_pull_commands_compat)
part_info = <craft_parts.infos.PartInfo object at 0x7fdbcf9c8850>

    def test_get_pull_commands_compat(part_info):
        properties = RustPlugin.properties_class.unmarshal(
            {"source": ".", "after": ["rust-deps"]}
        )
        plugin = RustPlugin(properties=properties, part_info=part_info)
>       plugin.validator_class(
            part_name="project",
            env="",
            properties=properties
        ).validate_environment(part_dependencies=["rust-deps"])

/home/lengau/Work/Code/craft-parts/tests/unit/plugins/test_rust_plugin.py:190: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/home/lengau/Work/Code/craft-parts/craft_parts/plugins/rust_plugin.py:103: in validate_environment
    options.rust_channel = "none"  # type: ignore[misc]
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   TypeError: "RustPluginProperties" is immutable and does not support item assignment

pydantic/main.py:359: TypeError
```

### Other information
When fixing this bug, remove the type  linter ignore comment in `rust_plugin.py` that mentions this bug.

liushuyu:

-----

This pull request fixes the following issues:

- Typing checking issues with the `rust_channel` override. This is unnecessary because some part of the configuration check is done in the validation step, making it impossible to set both `rust-deps` and `rust-channel` to a valid value (which will install two different Rust toolchains).
- Properly removes the compiler plugins. Rust plugin can now identify and remove the compiler plugins installed in the final staging area. Rust compiler will embed two different metadata into the compiler plugin. The craft-parts Rust plugin recognizes the compiler plugin by checking the dynamic symbol inserted into the `.data.rel.ro` section.

---

This pull request fixes #559.

### DIFF
--- a/craft_parts/plugins/rust_plugin.py
+++ b/craft_parts/plugins/rust_plugin.py
@@ -239,8 +239,6 @@ sh -s -- -y --no-modify-path --profile=minimal --default-toolchain {channel}
         for crate in options.rust_path:
             logger.info("Generating build commands for %s", crate)
             config_cmd_string = " ".join(config_cmd)
-            # TODO(liushuyu): the current fallback installation method will also install
-            # compiler plugins into the final Snap, which is likely not what we want.
             # pylint: disable=line-too-long
             rust_build_cmd_single = dedent(
                 f"""\
@@ -254,9 +252,12 @@ sh -s -- -y --no-modify-path --profile=minimal --default-toolchain {channel}
                     pushd "{crate}"
                     cargo build --workspace --release {config_cmd_string}
                     # install the final binaries
-                    # TODO(liushuyu): this will also install proc macros in the workspace,
-                    # which the user may not want to keep in the final installation
                     find ./target/release -maxdepth 1 -executable -exec install -Dvm755 {{}} "{self._part_info.part_install_dir}" ';'
+                    # remove proc_macro objects
+                    for i in "{self._part_info.part_install_dir}"/*.so; do
+                        readelf --wide --dyn-syms "$i" | grep -q '__rustc_proc_macro_decls_[0-9a-f]*__' && \
+                        rm -fv "$i"
+                    done
                     popd
                 fi\
                 """
--- a/craft_parts/plugins/rust_plugin.py
+++ b/craft_parts/plugins/rust_plugin.py
@@ -87,7 +87,7 @@ class RustPluginEnvironmentValidator(validator.PluginEnvironmentValidator):
         """
         if "rust-deps" in (part_dependencies or []):
             options = cast(RustPluginProperties, self._options)
-            if options.rust_channel:
+            if options.rust_channel and options.rust_channel != "none":
                 raise validator.errors.PluginEnvironmentValidationError(
                     part_name=self._part_name,
                     reason="rust-deps can not be used"
@@ -99,12 +99,6 @@ class RustPluginEnvironmentValidator(validator.PluginEnvironmentValidator):
                     plugin_name="rust",
                     part_dependencies=part_dependencies,
                 )
-            # The following type ignore is to make type checking pass while
-            # issue #559 is a bug:
-            # https://github.com/canonical/craft-parts/issues/559
-            # The ignore comment (and this block comment) should be removed
-            # with the fixing of #559.
-            options.rust_channel = "none"  # type: ignore[misc]
 
 
 class RustPlugin(Plugin):
--- a/setup.py
+++ b/setup.py
@@ -88,7 +88,7 @@ test_requires = [
     "isort",
     "hypothesis",
     "pydocstyle",
-    "pylint",
+    "pylint<3.0",
     "pylint-fixme-info",
     "pylint-pytest",
     "pyright==1.1.327",


liushuyu:
It seems like there are some issues when installing `pylint` in the CI runners.

craft_parts/plugins/rust_plugin.py
```diff
@@ -87,7 +87,7 @@ def validate_environment(
         """
         if "rust-deps" in (part_dependencies or []):
             options = cast(RustPluginProperties, self._options)
-            if options.rust_channel:
+            if options.rust_channel and options.rust_channel != "none":
```
lengau, line 90: Based on [the docs](https://canonical-craft-parts.readthedocs-hosted.com/en/latest/reference/plugins/rust_plugin.html), we can set `rust_channel` on `RustPluginProperties` to be `rust_channel: str = "stable"`, right? Then we only need `if options.rust_channel != "none":`

craft_parts/plugins/rust_plugin.py
```diff
@@ -254,9 +246,12 @@ def get_build_commands(self) -> List[str]:
                     pushd "{crate}"
                     cargo build --workspace --release {config_cmd_string}
                     # install the final binaries
-                    # TODO(liushuyu): this will also install proc macros in the workspace,
-                    # which the user may not want to keep in the final installation
                     find ./target/release -maxdepth 1 -executable -exec install -Dvm755 {{}} "{self._part_info.part_install_dir}" ';'
+                    # remove proc_macro objects
```
lengau, line 250: Is there any reason someone might want to keep those? (e.g. during debugging the snap)

I can't come up with one, but if there is we might want to make a way to prevent this from running.

lengau (approved):
Thanks! I have one question and one stylistic comment, but it also looks great as is.

craft_parts/plugins/rust_plugin.py
liushuyu: > Is there any reason someone might want to keep those? (e.g. during debugging the snap)
> 
> I can't come up with one, but if there is we might want to make a way to prevent this from running.

No. Debugging a Rust application Snap also does not need the compiler plugins.
Compiler plugins are only used to generate code during the compilation time.

craft_parts/plugins/rust_plugin.py
liushuyu: This is for backward compatibility. For older `core20` Snaps, setting it as `stable` by default will break the build.

lengau: Ahh, got it. Thank you!

lengau:
Linting issue is due to https://github.com/canonical/craft-parts/issues/562

liushuyu:
> Linting issue is due to #562

I see. I think `ruff` by now has most of the `pylint` rules (and many rules from `pylint` commonly used plugins).

lengau:
@liushuyu it does! I'm happy to merge this PR despite the unrelated pylint failure, but I'll defer to @cmatsuoka

sergiusens:
we agreed as a team to do pylint--

cmatsuoka:
We should land it. @lengau, can you do it?

cmatsuoka (approved):

sergiusens:
@cmatsuoka remember to properly edit the commit when merging/squashing

cmatsuoka:
> @cmatsuoka remember to properly edit the commit when merging/squashing

Yeah, it seems that the android Github app doesn't give you the opportunity to edit the commit message, and it was too late when I noticed that.

PR #563, opened 2023-10-07
fix(tests): replace ineffective call assertions

### LINKED ISSUE #368: The `craft_parts.packages.deb` module is incorrectly tested
There are 10 calls to `assert_has_calls`, which is misused in the tests.

E.g., the test `test_already_installed_no_specified_version` is currently doing...

```
        fake_deb_run.assert_has_calls([])
```

Attempting to validate that there are no calls to `apt` ("already installed!")... however, that line is a noop, as it does not verify any call in the mocker. Probably `assert_not_called` should have been used here (which would fail, as *there are* calls, because the test is buggy).

cmatsuoka:
Tests for the packages.deb module incorrectly use `assert_has_calls`
which are ineffective in certain situations. Normalize to assert
`mock_calls` instead.

Fixes #368

-----


tigarmo (approved):

lengau (approved):

cmatsuoka:
Rebased on main.

Issue #562, opened 2023-10-06
Remove pylint in favour of ruff


lengau:
### What needs to get done

1. Ensure that all of our enabled pylint checks are also enabled in the ruff configuration
2. Remove pylint from CI
3. Remove pylint from dependencies
4. Remove pylint from Makefile

### Why it needs to get done

Pylint 3.0 has caused our CI to break.

https://github.com/canonical/craft-parts/actions/runs/6436125741/job/17478834266?pr=547

cmatsuoka:
The commit that replaced pylint with ruff is 111c8907:

     "chore(linter): enable ruff base on starbase config (#580)" — Author: Sheng Yu, Nov 21
     2023.

     This commit removed the [tool.pylint] configuration from pyproject.toml and expanded the
     ruff configuration. (An earlier commit 487e049f "migrate from flake8 to ruff" introduced
     ruff, but pylint was removed here.)

PR #568, opened 2023-10-20
chore(linter): rebase on starbase linters config

syu-w:
Fixes: https://github.com/canonical/craft-parts/issues/562
CRAFT-2098


pyproject.toml
```diff
@@ -145,8 +117,14 @@ ignore = [
     "Q000", # 2 single-quoted strings - probably accidental
     "RET504", "RET506",  # Return value related.
     "PLR2004", # Magic value used in comparison
+    "S603", # `subprocess` call: check for execution of untrusted input
+    "S607", # Starting a process with a partial executable path
 
 ]
+[tool.ruff.pylint]
+max-args = 24
```
lengau, line 125: Let's keep this the same right now

```suggestion
max-args = 6
```

craft_parts/executor/environment.py
```diff
@@ -281,10 +281,10 @@ def _replace_attr(
 
     if isinstance(attr, dict):
         result: Dict[str, str] = {}
-        for key, value in attr.items():
+        for _key, _value in attr.items():
```
lengau, line 284: Note to other reviewers: These changes are for [PLW2901](https://docs.astral.sh/ruff/rules/redefined-loop-name/), changing the contents of the loop variable.

pyproject.toml
syu-w: OK, then I am going to add all the noqa to the affected functions

pyproject.toml
```diff
@@ -44,30 +44,6 @@ no_implicit_optional = true
...
-
-[tool.pylint.MASTER]
-extension-pkg-whitelist = [
-    "apt_pkg",
-    "pydantic"
-]
-load-plugins = "pylint_fixme_info,pylint_pytest"
-good-names = "id"
-
 [tool.ruff]
```
lengau, line 47: Could you update this section to match the latest starbase? Some of the linter rules we had enabled in pylint (e.g. unused arguments) aren't enabled here.

setup.py
```diff
@@ -88,16 +88,13 @@ def is_rtd() -> bool:
...
-    "pylint-fixme-info",
-    "pylint-pytest",
     "pyright==1.1.327",
     "pytest",
     "pytest-check",
     "pytest-cov",
     "pytest-mock",
     "requests-mock",
-    "ruff==0.0.239",
+    "ruff==0.1.0",
```
lengau, line 97: They already released 0.1.1 since you started this PR 😆

tests/integration/features/overlay_partitions/lifecycle/test_chisel_lifecycle.py
```diff
@@ -16,5 +16,4 @@
 
 # Make pytest run any non-overridden chisel tests here with partitions enabled.
```
lengau, line 18: Any file-wide rule-disables like this in tests can probably go into the `"tests/**.py"` section of ruff's per-file-ignores for the change. (In this case I'm thinking about F401 and F403)

tests/integration/test_main.py
```diff
@@ -71,7 +69,7 @@
 
 
 @pytest.fixture(autouse=True)
-def setup_new_dir(new_dir):  # pylint: disable=unused-argument
+def setup_new_dir(new_dir):
```
lengau, line 72:
```suggestion
@pytest.mark.usefixtures("new_dir")
def setup_new_dir():
```

tests/unit/executor/test_environment.py
```diff
@@ -48,7 +48,7 @@ def get_build_commands(self) -> List[str]:
 
 
 @pytest.fixture(autouse=True)
-def directories(new_dir):  # pylint: disable=unused-argument
+def directories(new_dir):
```
lengau, line 51:
```suggestion
@pytest.mark.usefixtures("new_dir")
def directories():
```

craft_parts/parts.py
```diff
@@ -73,24 +73,23 @@ class Config:
...
     @validator("overlay_packages", "overlay_files", "overlay_script")
-    def validate_overlay_feature(cls, item: Any) -> Any:
+    def validate_overlay_feature(cls, item: Any) -> Any:  # noqa: N805
         """Check if overlay attributes specified when feature is disabled."""
         assert Features().enable_overlay, "overlays not supported"
         return item
 
     @root_validator(pre=True)
-    def validate_root(cls, values: Dict[str, Any]) -> Dict[str, Any]:
+    def validate_root(cls, values: Dict[str, Any]) -> Dict[str, Any]:  # noqa: N805
```
lengau, line 92: This should be fixed by adding the following to `pyproject.toml`:

```
[tool.ruff.pep8-naming]
# Allow Pydantic's `@validator` decorator to trigger class method treatment.
classmethod-decorators = ["pydantic.validator", "pydantic.root_validator"]
```

lengau (changes requested):

pyproject.toml
lengau: Thank you!

I'd rather have noqa comments for things like this since they help make code smells more visible. There's some ongoing discussion about bumping it up further at https://github.com/canonical/starbase/issues/133

tigarmo:
@syu-w can/should we drop this PR?

pyproject.toml
cmatsuoka: 24 is a bit extreme but craft-parts has many calls with more than six arguments, not sure if it's productive to leave it at a value where we'll need too many PLR0913s.

PR #574, opened 2023-10-27
chore(linter): rebase on starbase linters config

syu-w:
Drop isort and pylint
Ignore current pathlib.Path checks

Bump Python version to 3.10

Updated linters to
pyright -> 1.1.333
ruff -> 0.1.3


.github/workflows/tests.yaml
```diff
@@ -55,7 +49,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-20.04]
-        python-version: ["3.7", "3.8", "3.9", "3.10"]
```
lengau, line 58: We might want to keep Python 3.8 until all of our apps are using core22 in their stable releases.

.gitignore
```diff
@@ -100,18 +122,23 @@ ENV/
...
-parts/
-stage/
-prime/
+# Ignore version module generated by setuptools_scm
+/*/_version.py
 
-# Tox environments
-/.tox/
+# Visual Studio Code
+.vscode/
```
lengau, line 144:
```suggestion
.vscode/

```

(newline at EOF)

.github/workflows/tests.yaml
syu-w: Then we can hold for now

craft_parts/permissions.py
```diff
@@ -43,40 +43,38 @@ class Permissions(BaseModel):
...
         """Validate that "owner" and "group" are correctly specified."""
         has_owner = "owner" in values
         has_group = "group" in values
 
-        assert (
-            has_group == has_owner
-        ), 'If either "owner" or "group" is defined, both must be'
+        if has_group != has_owner:
+            raise ValueError(
+                'If either "owner" or "group" is defined, both must be the same.'
```
tigarmo, line 58:
```suggestion
                'If either "owner" or "group" is defined, both must be.'
```

It's not that "owner" and "group" have to have the same value, it's that if the user defines one of them then they must also define the other.

craft_parts/plugins/maven_plugin.py
```diff
@@ -205,5 +206,5 @@ def _create_settings(settings_path: Path) -> None:
 
 
 def _get_no_proxy_string() -> str:
-    no_proxy = [k.strip() for k in os.environ.get("no_proxy", "localhost").split(",")]
+    no_proxy = [k.strip() for k in os.environ.get("NO_PROXY", "localhost").split(",")]
```
tigarmo, line 209: how come?

tests/conftest.py
```diff
@@ -232,3 +234,10 @@ def fake_chown(path, uid, gid, **kwargs):
     mocker.patch.object(os, "chown", side_effect=fake_chown)
 
     return calls
+
+
+@pytest.fixture()
+def _meson():
+    subprocess.run(["pip", "install", "meson"], check=True)
```
tigarmo, line 241: should probably do that on the ci workflow instead

tests/integration/lifecycle/test_lifecycle.py
```diff
@@ -51,7 +50,7 @@ def test_basic_lifecycle_actions(new_dir, partitions, mocker):
     # See https://gist.github.com/sergiusens/dcae19c301eb59e091f92ab29d7d03fc
 
     # first run
-    # command: pull
```
tigarmo, line 54: ?

tigarmo (changes requested):
thanks, this is a lot! Please add any changes as new commits (as opposed to amending the existing commit) so we can tell what's new and what's already been reviewed

Makefile
```diff
@@ -8,7 +8,6 @@ help: ## Show this help.
 
 .PHONY: autoformat
 autoformat: ## Run automatic code formatters.
-	isort $(SOURCES)
```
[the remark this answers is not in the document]
syu-w: `ruff --fix` will be much more than just format

craft_parts/lifecycle_manager.py
```diff
@@ -154,11 +153,10 @@ def __init__(
 
         if partitions:
             validate_partition_usage(part_list, partitions)
+            _validate_partition_usage_in_parts(part_list, partitions)
```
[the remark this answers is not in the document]
syu-w: IIRC `partitions` is empty or None that broke `_validate_partition_usage_in_parts`. You can try move it back and run the tests.

craft_parts/permissions.py
syu-w: OK

tests/conftest.py
syu-w: You mean install `meson` in the CI?

tests/integration/lifecycle/test_lifecycle.py
syu-w: ruff think that is comment out code that I should remove. While I don't want to disable it nor add `noqa`. I had to change it not looks like.

craft_parts/plugins/maven_plugin.py
syu-w: Env usually request to be uppercase. And you get an error for that.

cmatsuoka: Please double check if the name is not lowercase for a reason, and in either case we should also keep checking the lowercase version for backwards compatibility.

craft_parts/plugins/rust_plugin.py
```diff
@@ -33,27 +33,24 @@
 # A workaround for mypy false positives
 # see https://github.com/samuelcolvin/pydantic/issues/975#issuecomment-551147305
 # The proper fix requires Python 3.9+ (needs `typing.Annotated`)
-if TYPE_CHECKING:
-    UniqueStrList = List[str]
-else:
-    UniqueStrList = conlist(str, unique_items=True)
+UniqueStrList = list[str] if TYPE_CHECKING else conlist(str, unique_items=True)  # type: ignore[valid-type]
```
cmatsuoka, line 36: The original formatting seems a bit easier to read, can we keep it?

craft_parts/plugins/maven_plugin.py
sergiusens: this should be lowercase, you will break launchpad if you change this, it is a client interface

craft_parts/sources/tar_source.py
```diff
@@ -74,20 +73,15 @@ def __init__(
...
 
-    # pylint: enable=too-many-arguments
-
     @overrides
     def provision(
         self,
         dst: Path,
-        keep: bool = False,
-        src: Optional[Path] = None,
+        keep: bool = False,  # noqa: FBT001, FBT002
```
cmatsuoka, line 80: I think we should make `keep` and `src` keyword arguments (sources are not supposed to be used directly by applications so it should be safe)

craft_parts/utils/__init__.py
```diff
@@ -27,9 +27,9 @@ def package_name() -> str:
 class Singleton(type):
     """Singleton metaclass."""
 
-    _instances: Dict = {}
+    _instances: dict = {}  # type: ignore[type-arg]
 
-    def __call__(cls, *args: Any, **kwargs: Any) -> Any:
+    def __call__(cls, *args: Any, **kwargs: Any) -> Any:  # noqa: ANN401
```
cmatsuoka, line 32: Maybe use `dict[str, Any]` to avoid the warning instead of ignoring it?

craft_parts/sources/snap_source.py
```diff
@@ -40,19 +40,19 @@ class SnapSource(FileSourceHandler):
...
-        source_branch: Optional[str] = None,
-        source_depth: Optional[int] = None,
-        source_checksum: Optional[str] = None,
-        **kwargs,
+        source_tag: str | None = None,
+        source_commit: str | None = None,
+        source_branch: str | None = None,
+        source_depth: int | None = None,
+        source_checksum: str | None = None,
+        **kwargs: Any,  # noqa: ANN401
```
cmatsuoka, line 55: Maybe use `dict[str, Any]` to avoid the warning instead of ignoring it?

pyproject.toml
```diff
@@ -1,22 +1,41 @@
...
+exclude = [
+    "**/.*",
+    "**/__pycache__",
+    # pyright might not like the annotations generated by setuptools_scm
+    "**/_version.py",
+]
 
 [tool.mypy]
-python_version = "3.8"
+python_version = "3.10"
```
cmatsuoka, line 38: We may want to keep 3.8 for now, and import future annotations

craft_parts/overlays/chroot.py
```diff
@@ -32,7 +32,12 @@
 logger = logging.getLogger(__name__)
 
 
-def chroot(path: Path, target: Callable, *args: Any, **kwargs: Any) -> Any:
+def chroot(
+    path: Path,
+    target: Callable,  # type: ignore[type-arg]
+    *args: Any,  # noqa: ANN401
+    **kwargs: Any,  # noqa: ANN401
```
cmatsuoka, line 39: We can add proper typing for `args` and `kwargs` instead of disabling the warning.

craft_parts/overlays/chroot.py
```diff
@@ -66,9 +71,9 @@ def chroot(path: Path, target: Callable, *args: Any, **kwargs: Any) -> Any:
 def _runner(
     path: Path,
     conn: Connection,
-    target: Callable,
-    args: Tuple,
-    kwargs: Dict,
+    target: Callable,  # type: ignore[type-arg]
+    args: tuple[Any, ...],
+    kwargs: dict[str, Any],
```
cmatsuoka, line 76: Like this.

cmatsuoka (changes requested):
Thanks Sheng, that was a lot of work. Did a pass and added some comments.

tests/conftest.py
tigarmo: yeap. We already install [a lot of](https://github.com/canonical/craft-parts/blob/main/.github/workflows/tests.yaml#L85) build tools needed by the tests

craft_parts/sources/tar_source.py
syu-w: It already keyword arguments, unless I misunderstand the word

tigarmo: I think he meant mandatory kwargs, e.g.

```python
    def provision(
        self,
        dst: Path,
        *,
        keep: bool = False,
        src: Optional[Path] = None,
```

Makefile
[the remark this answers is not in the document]
lengau: We can turn off the commented out code checks if we choose

craft_parts/plugins/maven_plugin.py
```diff
@@ -206,5 +207,15 @@ def _create_settings(settings_path: Path) -> None:
...
 
 def _get_no_proxy_string() -> str:
-    no_proxy = [k.strip() for k in os.environ.get("NO_PROXY", "localhost").split(",")]
+    no_proxy = []
+
+    for address in os.environ.get("NO_PROXY", "localhost").split(","):
+        if address:
+            no_proxy.append(address.strip())
+
+    # backwards compatibility
```
tigarmo, line 216: I don't understand why we're mixing functional changes with the linting work. Can't we just keep the original code?

craft_parts/sources/local_source.py
```diff
@@ -43,10 +44,10 @@ class LocalSource(SourceHandler):
 
     def __init__(
         self,
-        *args: Any,  # noqa: ANN401
+        *args: tuple[Any, ...],
         project_dirs: ProjectDirs,
         copy_function: Callable[..., None] = file_utils.link_or_copy,
-        **kwargs: Any,  # noqa: ANN401
+        **kwargs: dict[str, Any],
```
tigarmo, line 50: I'm a bit surprised that this works, I thought for kwargs you only had to annotate the value type (since the key type is always `str`). So this would mean that `kwargs` is a `dict[str, dict[str, Any]]`
but well, if it works it works ;)

pyproject.toml
```diff
@@ -32,131 +51,131 @@ strict_equality = true
...
 disallow_subclassing_any = true
-#disallow_untyped_decorators = true
-#disallow_any_generics = true
+disallow_untyped_decorators = true
+disallow_any_generics = true
 
 [[tool.mypy.overrides]]
 module = ["craft-parts"]
-#disallow_untyped_defs = true
+disallow_untyped_defs = true
```
mr-cal, line 59: This PR is enormous and encompasses a lot of changes (mostly typing changes but I see lots of subtle changes too that I fear are going to get lost in the size of this PR).  

For the ticket size and acceptance criteria, I don't think enabling more mypy rules should be in this PR.  

Perhaps this PR should just do the work in the ticket:

> Acceptance Criteria
Update to the latest ruff that supports all the pylint features that we are currently not checking with ruff

craft_parts/sources/local_source.py
syu-w: After check PEP, the correct typing for this are `*args: Any` and `**kwargs: Any`.
Since `*args: tuple(T, ...)` and `**kwargs: dict[str, T]`, if we don't what's in it, it has to be `Any`.
@cmatsuoka

craft_parts/plugins/maven_plugin.py
syu-w: Sure, but I changed it because the linting error:
```
craft_parts/plugins/maven_plugin.py:210:51: SIM112 Use capitalized environment variable `NO_PROXY` instead of `no_proxy`
```

tigarmo:
But won't this new version cause the same linting error?
In any case, I think the fix is to add the lint ignore here

tigarmo:
Nooo @syu-w please don't amend commits while the code is under review, it makes it basically impossible to tell what changed
I'm dismissing my "request changes" review because I'm off for the next few days

syu-w:
@tigarmo It's not ready for review yet since I am testing CI 😂

syu-w:
pydantic does not support using `from __future__ import annotations` in the model, and it got python exception like try to slice `list` class. So to support older Python, I had to revert the typing style.

cmatsuoka:
> pydantic does not support using `from __future__ import annotations` in the model, and it got python exception like try to slice `list` class. So to support older Python, I had to revert the typing style.

I think for now it's the thing to do (unfortunately), at least for the pydantic models.

tests/unit/utils/test_file_utils.py
```diff
@@ -159,10 +159,10 @@ def test_with_permissions(self, mock_chown):
         file_utils.link_or_copy("foo/2", "qux/2", permissions=permissions)
 
         # Check that the copied file has the correct permission bits and ownership
-        assert stat.S_IMODE(os.stat("qux/2").st_mode) == 0o755
+        assert stat.S_IMODE(os.stat("qux/2").st_mode) == 0o755  # noqa: PLR2004
```
lengau, line 162: I'm not asking for action here, but I opened a discussion about these: https://github.com/canonical/starbase/issues/148

tests/unit/test_parts.py
```diff
@@ -504,13 +510,15 @@ def test_get_parts_with_overlay(self, partitions):
 class TestPartValidation:
     """Part validation considering plugin-specific attributes."""
 
-    def test_part_validation_data_type(self, partitions):
+    @pytest.mark.usefixtures("partitions")
+    def test_part_validation_data_type(self):
         with pytest.raises(TypeError) as raised:
-            parts.validate_part("invalid data")  # type: ignore
+            parts.validate_part("invalid data")  # type: ignore[reportGeneralTypeIssues]
```
lengau, line 516: :heart:

tests/unit/test_infos.py
```diff
@@ -201,7 +203,7 @@ def test_project_info_set_project_raw_write():
 def test_project_info_set_project_var_bad_name():
     info = ProjectInfo(application_name="test", cache_dir=Path())
 
-    with pytest.raises(ValueError) as raised:
+    with pytest.raises(ValueError) as raised:  # noqa: PT011
```
lengau, line 206: I'd like @cmatsuoka's opinion here. This is complaining that we don't have a `match=` in the context manager. We're comparing the string below instead. I think the options are:

1. Keep it as is, so the lint error reminds us to check equality
2. Move to using `match`
3. Silence PT011 for `craft-parts` (and possibly in all our repos)

tests/unit/state_manager/test_prime_state.py
```diff
@@ -51,7 +51,9 @@ def test_marshal_unmarshal(self):
 
     def test_unmarshal_invalid(self):
         with pytest.raises(TypeError) as raised:
-            PrimeState.unmarshal(False)  # type: ignore
+            PrimeState.unmarshal(
+                False  # noqa: FBT003 # type: ignore[reportGeneralTypeIssues]
```
lengau, line 55: I'd rather not silence this - the parameter name makes this much clearer

lengau:
Per external discussion, we're going to redo this

PR #579, opened 2023-11-14
plugins: ant: add proxy support

kubiko:
Add proxy support to ant plugin

Currently, this is a limitation to running remote builds in the launchpad as those builds require proxy support.

### DIFF
--- a/craft_parts/plugins/ant_plugin.py
+++ b/craft_parts/plugins/ant_plugin.py
@@ -18,6 +18,7 @@
 
 import logging
 import os
+import shlex
 from typing import Any, Dict, Iterator, List, Optional, Set, cast
 from urllib.parse import urlsplit
 
@@ -152,7 +153,7 @@ class AntPlugin(JavaPlugin):
         ant_opts.extend(_get_proxy_options("http"))
         ant_opts.extend(_get_proxy_options("https"))
         if ant_opts:
-            opts = " ".join(opt.replace("'", "'\\''") for opt in ant_opts)
+            opts = shlex.join(ant_opts)
             return {"ANT_OPTS": opts}
         return {}
 
--- a/craft_parts/plugins/ant_plugin.py
+++ b/craft_parts/plugins/ant_plugin.py
@@ -17,7 +17,9 @@
 """The Ant plugin."""
 
 import logging
-from typing import Any, Dict, List, Optional, Set, cast
+import os
+from typing import Any, Dict, Iterator, List, Optional, Set, cast
+from urllib.parse import urlsplit
 
 from overrides import override
 
@@ -142,6 +144,16 @@ class AntPlugin(JavaPlugin):
     @override
     def get_build_environment(self) -> Dict[str, str]:
         """Return a dictionary with the environment to use in the build step."""
+        # Getting ant to use a proxy requires a little work; the JRE doesn't
+        # help as much as it should.  (java.net.useSystemProxies=true ought
+        # to do the trick, but it relies on desktop configuration rather
+        # than using the standard environment variables.)
+        ant_opts = []  # type: List[str]
+        ant_opts.extend(_get_proxy_options("http"))
+        ant_opts.extend(_get_proxy_options("https"))
+        if ant_opts:
+            opts = " ".join(opt.replace("'", "'\\''") for opt in ant_opts)
+            return {"ANT_OPTS": opts}
         return {}
 
     @override
@@ -160,3 +172,17 @@ class AntPlugin(JavaPlugin):
         command.extend(options.ant_build_targets)
 
         return [" ".join(command), *self._get_java_post_build_commands()]
+
+
+def _get_proxy_options(scheme) -> Iterator[str]:
+    proxy = os.environ.get(f"{scheme}_proxy")
+    if proxy:
+        parsed = urlsplit(proxy)
+        if parsed.hostname is not None:
+            yield f"-D{scheme}.proxyHost={parsed.hostname}"
+        if parsed.port is not None:
+            yield f"-D{scheme}.proxyPort={parsed.port}"
+        if parsed.username is not None:
+            yield f"-D{scheme}.proxyUser={parsed.username}"
+        if parsed.password is not None:
+            yield f"-D{scheme}.proxyPassword={parsed.password}"
--- a/craft_parts/plugins/ant_plugin.py
+++ b/craft_parts/plugins/ant_plugin.py
@@ -153,8 +153,7 @@ class AntPlugin(JavaPlugin):
         ant_opts.extend(_get_proxy_options("http"))
         ant_opts.extend(_get_proxy_options("https"))
         if ant_opts:
-            opts = shlex.join(ant_opts)
-            return {"ANT_OPTS": opts}
+            return {"ANT_OPTS": _shlex_join(ant_opts)}
         return {}
 
     @override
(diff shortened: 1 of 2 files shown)


tests/unit/plugins/test_ant_plugin.py
```diff
@@ -124,6 +127,53 @@ def test_get_build_environment(part_info):
...
+        ant_opts = ""
+        if host != "":
+            ant_opts = f"{ant_opts} -D{protocol}.proxyHost={host}"
+        if port != "":
+            ant_opts = f"{ant_opts} -D{protocol}.proxyPort={port}"
+        if username != "":
+            ant_opts = f"{ant_opts} -D{protocol}.proxyUser={username}"
+        if password != "":
+            ant_opts = f"{ant_opts} -D{protocol}.proxyPassword={password}"
+        assert plugin.get_build_environment() == {"ANT_OPTS": ant_opts.strip()}
```
lengau, line 174: This feels like it's repeating the same logic as the plugin. 

It's way less pretty, but this seems safer to me:

```suggestion
@pytest.mark.parametrize(
    ("env", "opts"),
    [
        ({"http_proxy": "http://my-host"}, "-Dhttp.proxyHost=my-host"),
        (
            {"http_proxy": "http://my-host:3128"},
            "-Dhttp.proxyHost=my-host -Dhttp.proxyPort=3128",
        ),
        (
            {"http_proxy": "http://user@my-host:3128"},
            "-Dhttp.proxyHost=my-host -Dhttp.proxyPort=3128 -Dhttp.proxyUser=user",
        ),
        (
            {"http_proxy": "http://user:pass@my-host:3128"},
            "-Dhttp.proxyHost=my-host -Dhttp.proxyPort=3128 -Dhttp.proxyUser=user -Dhttp.proxyPassword=pass",
        ),
        (
            {"http_proxy": "http://user:pass@my-host"},
            "-Dhttp.proxyHost=my-host -Dhttp.proxyUser=user -Dhttp.proxyPassword=pass",
        ),
        ({"https_proxy": "https://my-host"}, "-Dhttps.proxyHost=my-host"),
        (
            {"https_proxy": "https://my-host:3128"},
            "-Dhttps.proxyHost=my-host -Dhttps.proxyPort=3128",
        ),
        (
            {"https_proxy": "https://user@my-host:3128"},
            "-Dhttps.proxyHost=my-host -Dhttps.proxyPort=3128 -Dhttps.proxyUser=user",
        ),
        (
            {"https_proxy": "https://user:pass@my-host:3128"},
            "-Dhttps.proxyHost=my-host -Dhttps.proxyPort=3128 -Dhttps.proxyUser=user -Dhttps.proxyPassword=pass",
        ),
        (
            {"https_proxy": "https://user:pass@my-host"},
            "-Dhttps.proxyHost=my-host -Dhttps.proxyUser=user -Dhttps.proxyPassword=pass",
        ),
    ],
)
def test_get_build_environment_proxy(part_info, env, opts):
    properties = AntPlugin.properties_class.unmarshal({"source": "."})
    plugin = AntPlugin(properties=properties, part_info=part_info)

    with mock.patch.dict(os.environ, env):
        assert plugin.get_build_environment() == {"ANT_OPTS": opts}
```

craft_parts/plugins/ant_plugin.py
```diff
@@ -142,6 +144,16 @@ def get_build_packages(self) -> Set[str]:
...
+        # Getting ant to use a proxy requires a little work; the JRE doesn't
+        # help as much as it should.  (java.net.useSystemProxies=true ought
+        # to do the trick, but it relies on desktop configuration rather
+        # than using the standard environment variables.)
+        ant_opts = []  # type: List[str]
+        ant_opts.extend(_get_proxy_options("http"))
+        ant_opts.extend(_get_proxy_options("https"))
+        if ant_opts:
+            opts = " ".join(opt.replace("'", "'\\''") for opt in ant_opts)
+            return {"ANT_OPTS": opts}
```
lengau, line 156: The maven plugin deals with this by updating an XML file. Could we unify these by putting this implementation into the `JavaPlugin` base class and setting the `_JAVA_OPTIONS` environment variable instead?

craft_parts/plugins/ant_plugin.py
```diff
@@ -142,6 +144,16 @@ def get_build_packages(self) -> Set[str]:
...
         """Return a dictionary with the environment to use in the build step."""
+        # Getting ant to use a proxy requires a little work; the JRE doesn't
+        # help as much as it should.  (java.net.useSystemProxies=true ought
+        # to do the trick, but it relies on desktop configuration rather
+        # than using the standard environment variables.)
+        ant_opts = []  # type: List[str]
+        ant_opts.extend(_get_proxy_options("http"))
+        ant_opts.extend(_get_proxy_options("https"))
+        if ant_opts:
+            opts = " ".join(opt.replace("'", "'\\''") for opt in ant_opts)
```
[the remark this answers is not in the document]
kubiko: Thank you, I have changed this now. I have admin I have picked this directly from v1 version of the ant plugin :)

tests/unit/plugins/test_ant_plugin.py
kubiko: Thank you for the suggestion @lengau!
Indeed same logic from plugin was repeated here, your approach seems cleaner and it prevents someone from simply copying logic from a plugin to fix the test.

craft_parts/plugins/ant_plugin.py
kubiko: I would ask this to be done as a separate PR.
I have created this PR as the bug I opened in Snapcraft was not getting any attention and with 18.04 now being in ESM, we need to provide an upgrade path to any `base: core18` snaps to move to `base core22`. Currently, any ant builds fail in Launchpad because of missing proxy support.

kubiko:
@lengau For some reason when I moved `shlex` my tests are now failing, but only when run on GitHub, when I run those tests locally all works fine. Any idea?

tigarmo:
@kubiko I don't know how it worked before, but `shlex.join()` is only available in Python >=3.8

kubiko:
@tigarmo Well it did not as it was not used before :)
I have moved to `shlex.join()` only after a review comment from @lengau but now I cannot get my tests to pass.

tigarmo:
I think the version of craft-parts that will eventually have this fix might need to keep Python 3.7 support, so maybe we should do something like [this fix](https://github.com/aajanki/yle-dl/commit/9d6e9c54dd7cf964d1737793dca5880d6c866d19)?

kubiko:
> I think the version of craft-parts that will eventually have this fix might need to keep Python 3.7 support, so maybe we should do something like [this fix](https://github.com/aajanki/yle-dl/commit/9d6e9c54dd7cf964d1737793dca5880d6c866d19)?

Thank you @tigarmo! that's good idea
added this now

kubiko:
@tigarmo @lengau Finally I managed to pass all the checks, please re-review and thank you for your help with this!

cmatsuoka (approved):

cmatsuoka:
Thanks!

cmatsuoka:
We'll need to backport this if you need the feature to be available in Snapcraft 7.x or 8.0.

craft_parts/plugins/ant_plugin.py
cmatsuoka: So it seems that this also needs to be added as a hotfix to 1.26.

PR #580, opened 2023-11-17
chore(linter): enable ruff base on starbase config

syu-w:

PR #581, opened 2023-11-17
chore(linter): ruff auto fix base on starbase config

syu-w:
All auto fix.

Depends on https://github.com/canonical/craft-parts/pull/580


pyproject.toml
lengau, line 1: Note to other reviewers: This is the only file that was changed other than using `ruff check --fix .`

lengau (approved):

craft_parts/callbacks.py
```diff
@@ -33,7 +33,10 @@
 StepCallback = Callable[[StepInfo], bool]
 ConfigureOverlayCallback = Callable[[Path, ProjectInfo], None]
 Callback = Union[
-    FilterCallback, ExecutionCallback, StepCallback, ConfigureOverlayCallback
+    FilterCallback,
```
tigarmo, line 36: did we change the desired line length?

craft_parts/executor/part_handler.py
```diff
@@ -743,7 +772,8 @@ def _migrate_overlay_files_to_prime(self) -> None:
         for part in reversed(parts_with_overlay):
             logger.debug("migrate part %r layer to prime", part.name)
             visible_files, visible_dirs = overlays.visible_in_layer(
-                part.part_layer_dir, part.prime_dir
+                part.part_layer_dir,
```
tigarmo, line 775: do you know which rule is splitting lines like this one?

syu-w: that come from the `ruff format', which I think has no rule code

craft_parts/callbacks.py
syu-w: No, but the `ruff format` seems like break long lines.

cmatsuoka (changes requested):
Replacing other linters with ruff looks fine, however I think we should wait for ruff formatting to get stable and use black for now (instead of always splitting argument lists).

craft_parts/executor/part_handler.py
syu-w: OK, after some tests, this kind of format was done by `black`, not `ruff format`.

syu-w:
Now without `COM812 # Trailing comma missing`

pyproject.toml
```diff
@@ -138,19 +138,16 @@ ignore = [
     # Temporary ignored due to API change
     "PTH", # Use pathlib.Path
 
+    "COM812", # Trailing comma missing
```
sergiusens, line 141: why did you move this from the original diff?

this is still in starbase, so keeping it below seems appropriate

craft_parts/packages/apt_cache.py
```diff
@@ -235,7 +235,7 @@ def get_installed_version(
                 return installed.version
         return None
 
-    def fetch_archives(self, download_path: Path) -> List[Tuple[str, str, Path]]:
+    def fetch_archives(self, download_path: Path) -> list[tuple[str, str, Path]]:
```
tigarmo, line 238: I was going to comment on the `List` vs `list`, but apparently this file has had the `__future__.annotations` import for over two years now! Weird

tigarmo (approved):
thanks!

tests/conftest.py
```diff
@@ -211,13 +210,13 @@ def create_dependency_fixture(
...
 
-    yield create_dependency_fixture
+    return create_dependency_fixture
 
 
 ChmodCall = collections.namedtuple("ChmodCall", ["owner", "group", "kwargs"])
 
 
-@pytest.fixture
+@pytest.fixture()
```
cmatsuoka, line 219: Any special reason not to use `dict` in the line below?

tests/integration/executor/test_callbacks.py
```diff
@@ -132,7 +131,7 @@ def test_prologue_callback(tmpdir, capfd):
 
 
 def _my_step_callback(info: StepInfo) -> bool:
-    msg = getattr(info, "message")
+    msg = info.message
```
cmatsuoka, line 134: We can remove the assignment and print `info.message` directly.

tests/integration/executor/test_callbacks.py
```diff
@@ -330,7 +329,7 @@ def test_invalid_update_callback_post(tmpdir, step):
 
 
 def _my_exec_callback(info: ProjectInfo) -> None:
-    print(f"{getattr(info, 'message')}")
+    print(f"{info.message}")
```
cmatsuoka, line 332: Any special reason not to print directly instead of using the f-string?

tests/integration/plugins/test_python.py
```diff
@@ -20,12 +20,11 @@
...
 from typing import Optional
 
+import craft_parts.plugins.plugins
 import pytest
 import yaml
-from overrides import override
-
-import craft_parts.plugins.plugins
 from craft_parts import LifecycleManager, Step, errors, plugins
+from overrides import override
```
cmatsuoka, line 27: Interesting order swap here, shouldn't third-party imports be in the middle block and project-local imports in the bottom block?

tests/unit/plugins/test_ant_plugin.py
```diff
@@ -15,13 +15,12 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import pytest
-from pydantic import ValidationError
-
 from craft_parts import Part, PartInfo, ProjectInfo, errors
 from craft_parts.plugins.ant_plugin import AntPlugin
+from pydantic import ValidationError
```
cmatsuoka, line 20: Same third-party vs local reordering here. Is isort happy with that? Otherwise it could be a ruff bug or configuration.

tests/unit/plugins/test_autotools_plugin.py
```diff
@@ -17,11 +17,10 @@
 from pathlib import Path
 
 import pytest
-from pydantic import ValidationError
-
 from craft_parts.infos import PartInfo, ProjectInfo
 from craft_parts.parts import Part
 from craft_parts.plugins.autotools_plugin import AutotoolsPlugin
+from pydantic import ValidationError
```
cmatsuoka, line 23: Ditto.

tests/unit/utils/test_file_utils.py
```diff
@@ -129,7 +128,7 @@ def test_link_file_ioerror(self, mocker):
 
         def link_and_ioerror(a, b, **kwargs):  # pylint: disable=unused-argument
             orig_link(a, b)
-            raise IOError()
+            raise OSError()
```
cmatsuoka, line 131: Maybe we could rename the function to `link_and_oserror`.

cmatsuoka (commented):
Nice! Added a few questions and nitpicks but overall it looks pretty good!

cmatsuoka (approved):

tests/integration/plugins/test_python.py
syu-w: I guess the test cases do not count as project-local imports

tests/unit/plugins/test_ant_plugin.py
syu-w: isort moved it to the top

syu-w:
There could be some current ignored rules make the import order changed. Let me see if following PRs will fix that or not.

pyproject.toml
syu-w: I will remove the ignored rules below in the following PRs that fix them, so I put this one on top that should not be removed.

craft_parts/packages/apt_cache.py
syu-w: I removed `__future__.annotations` for now to keep the style the same with other files.

tests/conftest.py
syu-w: Just keep the old style for now

PR #587, opened 2023-11-27
docs: add rockcraft chisel notes

syu-w:


docs/base/part_properties.rst
```diff
@@ -311,6 +311,15 @@ The packages to install in the staging area for deployment with the build
 products. These provide libraries and executables to support the deployed
 part.
 
+For Rockcraft:
```
cmatsuoka, line 314: It works regardless of the application in use, any special need to mention Rockcraft? If this is valid only for certain applications we can say so and refer to specific application documentation.

syu-w: OK, I removed Rockcraft, it was said only for rockcraft in its docs.

docs/base/part_properties.rst
```diff
@@ -311,6 +311,14 @@ The packages to install in the staging area for deployment with the build
 products. These provide libraries and executables to support the deployed
 part.
 
+This keyword also support chiselled packages.
```
tigarmo, line 314:
```suggestion
This keyword also supports `Chisel <https://github.com/canonical/chisel>`_ slices.
```

tigarmo (approved):
Please don't force-push commits during the reviewwwww
:cry: :sob: :crying_cat_face:

cmatsuoka (approved):
Thanks!

PR #599, opened 2023-12-06
feat(partitions): validate namespaced partitions

### LINKED ISSUE #595: Parse and validate namespaced partitions
### What needs to get done

The LifecycleManager needs to parse and validate namespaced partitions.  

Namespaces can only contain lowercase alphabetical characters.  Namespaced partitions can contain lowercase alphabetical characters and hyphens.  They cannot start or end with a hyphen.  

A new error will inform the user of invalid namespaced partitions.  This means a single regex check will not be sufficient.  

#### Existing error - invalid partition

This error already exists:
```
Partition 'test!' must only contain lowercase letters.
```

#### New error - invalid namespaced partition

This error occurs when a namespaced partitions contains invalid characters.

```
Namespaced partition 'test!/foo' is invalid.  Namespaced partitions are formatted 
as `<namespace>/<partition>`.  

Namespaces must only contain lowercase letters.

Nmespaced partitions must only contain lowercase letters and hyphens and 
cannot start or end with a hyphen.  
```

Unit tests only, no integration tests will be written for this task.

### Why it needs to get done

To support components in snapcraft

mr-cal:

-----

Validate namespaced partitions when intializing the `LifecycleManager`.

Fixes #595 
(CRAFT-2032)


tests/unit/features/partitions/test_lifecycle_manager.py
```diff
@@ -154,7 +197,42 @@ def test_partitions_invalid(self, new_dir, parts_data, partitions):
...
+            ["default", "/a/b"],
+            ["default", "a//b"],
+            ["default", "/a/"],
+            ["default", "a/Test"],
+            ["default", "a/TEST"],
+            ["default", "a/test1"],
+            ["default", "Test/a"],
+            ["default", "TEST/a"],
+            ["default", "test1/a"],
+            ["default", "te-st/a"],
```
cmatsuoka, line 228: Shouldn't we also test for namespaced partition names starting or ending with dash such as `a/-b`, `a/b-`, or even just `-`, `-/b`, etc?

tests/unit/features/partitions/test_lifecycle_manager.py
```diff
@@ -46,6 +46,33 @@ def valid_partitions_strategy():
     return strategy.filter(lambda partitions: "default" not in partitions[1:])
 
 
+def valid_namespaced_partitions_strategy():
```
cmatsuoka, line 49: Does hypothesis allow generating an invalid list from the valid definition? It could be an interesting test too.

craft_parts/utils/partition_utils.py
```diff
@@ -51,8 +59,40 @@ def validate_partition_names(partitions: Optional[Sequence[str]]) -> None:
...
 
+    namespaces: Set[Tuple[str, str]] = set()
+    regular_partitions: Set[str] = set()
+
     for partition in partitions:
-        if not re.fullmatch("[a-z]+", partition):
+        # validate regular partitions
+        if re.fullmatch("[a-z]+", partition):
+            regular_partitions.add(partition)
+        # validate namespaced partitions
```
cmatsuoka, line 69: Maybe add a check against duplicate entries? We can have strange side effects during processing if we allow a namespaced partition listed twice.

tests/unit/features/partitions/test_lifecycle_manager.py
mr-cal: @lengau, do you know?  I was wondering the same thing and it seemed like hypothesis is not designed for that.

craft_parts/utils/partition_utils.py
```diff
@@ -51,8 +59,40 @@ def validate_partition_names(partitions: Optional[Sequence[str]]) -> None:
     if len(partitions) != len(set(partitions)):
         raise errors.FeatureError("Partitions must be unique.")
 
+    namespaces: Set[Tuple[str, str]] = set()
+    regular_partitions: Set[str] = set()
+
     for partition in partitions:
-        if not re.fullmatch("[a-z]+", partition):
+        # validate regular partitions
+        if re.fullmatch("[a-z]+", partition):
```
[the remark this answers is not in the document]
lengau: Since we're using this regex (and the regex below) multiple times, it might make the most sense to compile them as constants up at the top of the file.

As a side-effect, that also gives us an easy way to use hypothesis to check the regex below, which is sufficiently complex that we might want to test it.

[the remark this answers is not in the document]
mr-cal: I pulled the details into smaller functions but am keeping the regex because `islower()` and  `isalpha()` include a lot of characters we probably don't want.  For example:
```
>>> "𞥋helloʔ𞥋".islower() and "𞥋helloʔ𞥋".isalpha()
True
```

tests/unit/features/partitions/test_lifecycle_manager.py
mr-cal: Good idea.  I added many more invalid test cases with hyphens.

tests/unit/features/partitions/test_lifecycle_manager.py
lengau: I don't think Hypothesis has a way to invert strategies. (I'm not sure how one could actually achieve this in the general case, though for text it might be more feasible.)

craft_parts/utils/partition_utils.py
```diff
@@ -51,8 +59,81 @@ def validate_partition_names(partitions: Optional[Sequence[str]]) -> None:
...
+
+
+def _is_valid_partition(partition: str) -> bool:
+    """Check if a partition name is valid.
+
+    :param partition: partition to check
+
+    :returns: true if the namespaced partition is valid
+    """
+    return bool(re.fullmatch("[a-z]+", partition))
```
lengau, line 74: If we compile the regex at import time and make it a constant:

```suggestion
_VALID_PARTITION_REGEX = re.compile("[a-z]+", re.ASCII)

def _is_valid_partition(partition: str) -> bool:
    """Check if a partition name is valid.

    :param partition: partition to check

    :returns: true if the namespaced partition is valid
    """
    return bool(_VALID_PARTITION_REGEX.fullmatch(partition))
```

We can make it faster for repeated use (and optionally do direct testing of the regex using hypothesis.)

cmatsuoka: In this specific case `name.isalpha() and name.islower()` could be more efficient but at this scale the difference would be negligible.

craft_parts/utils/partition_utils.py
```diff
@@ -51,8 +59,81 @@ def validate_partition_names(partitions: Optional[Sequence[str]]) -> None:
...
+            regular_partitions.add(partition)
+        else:
+            namespaced_partitions.add(partition)
+
+    for regular_partition in regular_partitions:
+        for namespaced_partition in namespaced_partitions:
+            if namespaced_partition.startswith(regular_partition + "/"):
+                raise errors.FeatureError(
+                    f"Partition {regular_partition!r} conflicts with the namespace of "
+                    f"partition {namespaced_partition!r}"
```
cmatsuoka, line 138: Set operations would be nice but I see that the nested loops result in a better error message.

craft_parts/utils/partition_utils.py
```diff
@@ -51,8 +59,81 @@ def validate_partition_names(partitions: Optional[Sequence[str]]) -> None:
...
+def _validate_partition_names(partitions: Sequence[str]) -> None:
+    """Validate names of partitions.
+
+    :param partitions: Sequence of partitions to validate.
+
+    :raises FeatureError: if a partition name is not valid
+    """
     for partition in partitions:
-        if not re.fullmatch("[a-z]+", partition):
+        if _is_valid_partition(partition) or _is_valid_namespaced_partition(partition):
```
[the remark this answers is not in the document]
mr-cal: Renamed.

tests/unit/features/partitions/test_lifecycle_manager.py
```diff
@@ -154,7 +205,53 @@ def test_partitions_invalid(self, new_dir, parts_data, partitions):
...
+            ["default", "-a/b"],
+            ["default", "a-/b"],
+            ["default", "-a-/b"],
+            ["default", "a/Test"],
+            ["default", "a/TEST"],
+            ["default", "a/test1"],
+            ["default", "Test/a"],
+            ["default", "TEST/a"],
+            ["default", "test1/a"],
+            ["default", "te-st/a"],
```
lengau, line 247: Just to cover all of our bases:

```suggestion
            ["default", "te-st/a"],
            pytest.param(["default", "test/ο"], id="uses-omicron"),
```

tests/unit/features/partitions/test_lifecycle_manager.py
```diff
@@ -144,6 +194,7 @@ def test_partitions_default_not_first(self, new_dir, parts_data, partitions):
         "partitions",
         [
             ["default", ""],
+            ["default", "-"],
```
lengau, line 197:
```suggestion
            ["default", "-"],
            pytest.param(["default", "tеst"], id="test-with-cyrillic-e"),
```

lengau (approved):
Looks good! I noticed that we're not actually testing for non-ASCII letters, so I suggested two tricky examples just to cover our bases.

tests/unit/features/partitions/test_lifecycle_manager.py
mr-cal: Nice!

Issue #596, opened 2023-12-01
Handle namespaced partitions in utility functions


mr-cal:
### What needs to get done

The functions [`get_partitioned_path()`](https://github.com/canonical/craft-parts/blob/8fea6453667863903fc6d452c6b9b3d8c46b17c9/craft_parts/utils/path_utils.py#L41), [`get_partition_and_path()`](https://github.com/canonical/craft-parts/blob/8fea6453667863903fc6d452c6b9b3d8c46b17c9/craft_parts/utils/path_utils.py#L66), [`get_stage_dir()`](https://github.com/canonical/craft-parts/blob/8fea6453667863903fc6d452c6b9b3d8c46b17c9/craft_parts/dirs.py#L78), and [`get_prime_dir()`](https://github.com/canonical/craft-parts/blob/8fea6453667863903fc6d452c6b9b3d8c46b17c9/craft_parts/dirs.py#L84),  need to handle namespaced partitions.

The unit tests for these functions should verify that namespaced partitions are properly handled.

### Why it needs to get done

To support components in snapcraft

PR #600, opened 2023-12-07
feat(partitions): handle namespaced partitions in utils

mr-cal:

-----

Handle namespaced partitions in utility functions, `get_partitioned_path()`, `get_partition_and_path()`, `get_stage_dir()`, and `get_prime_dir()`.

#596 
(CRAFT-2303)

### DIFF
--- a/craft_parts/utils/path_utils.py
+++ b/craft_parts/utils/path_utils.py
@@ -17,13 +17,18 @@
 """Utility functions for paths."""
 import re
 from pathlib import PurePath, PurePosixPath
-from typing import NamedTuple, Optional, TypeVar, Union, cast
+from typing import NamedTuple, Optional, Tuple, TypeVar, Union
 
+from craft_parts.errors import FeatureError
 from craft_parts.features import Features
 
 FlexiblePath = TypeVar("FlexiblePath", PurePath, str)
 
+# regex for a path beginning with a partition
 HAS_PARTITION_REGEX = re.compile(r"^\([a-z]+\)(/.*)?$")
+
+# regex for a path beginning with a namespaced partition
+HAS_NAMESPACED_PARTITION_REGEX = re.compile(r"^(\([a-z]+/(?!-)[a-z\-]+(?<!-)\))(/.*)?$")
 
 
 class PartitionPathPair(NamedTuple):
@@ -35,7 +40,10 @@ class PartitionPathPair(NamedTuple):
 
 def _has_partition(path: Union[PurePath, str]) -> bool:
     """Check whether a path has an explicit partition."""
-    return bool(HAS_PARTITION_REGEX.match(str(path)))
+    return bool(
+        HAS_PARTITION_REGEX.match(str(path))
+        or HAS_NAMESPACED_PARTITION_REGEX.match(str(path))
+    )
 
 
 def get_partitioned_path(path: FlexiblePath) -> FlexiblePath:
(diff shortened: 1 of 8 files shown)


craft_parts/utils/path_utils.py
```diff
@@ -80,8 +87,33 @@ def get_partition_and_path(path: FlexiblePath) -> PartitionPathPair:
...
+    :returns: A tuple containing the partition and inner path. If there is no inner
+    path, the second element will be an empty string.
+    """
+    # split regular partitions
+    if re.match(HAS_PARTITION_REGEX, str_path):
+        if "/" in str_path:
+            # split into partition and inner_path
+            partition, inner_path = str_path.split("/", maxsplit=1)
+            # remove extra forward slashes between the partition and inner path
+            return partition, inner_path.lstrip("/")
```
mr-cal, line 110: This was a bug I uncovered, fixed, and tested as part of this PR

The bug is that `get_partitioned_path("(default)//foo")` returned `/foo` instead of `default/foo`

lengau: Oh that's sneaky. Well spotted!

tests/unit/features/partitions/test_parts.py
```diff
@@ -130,39 +130,94 @@ def valid_filesets(self):
...
+            "(a/c-d)//test",
+            "(a/c-d)/test/(otherdir)",
+            "(a/c-d)/test/(otherdir-123)",
+            "(a/c-d)/test/(default)",
+            "(a/c-d)/test/(kernel)",
+            "(a/c-d)/test/(a/b)",
+            "(a/c-d)/test/(a/c-d)",
+            "test/(a/c-d)",
+            "test(a/c-d)/test",
+            "test/(a/c-d)/test",
```
cmatsuoka, line 187: Interesting list of cases, why is `(a/c-d)/...` considered valid? Is it because it considers `(a/c-d)` as a directory name? Or are `(a` and `c-d)` treated as separate directories?

tests/unit/features/partitions/test_parts.py
```diff
@@ -130,39 +130,94 @@ def valid_filesets(self):
...
-        Assumes "default" and "partition" are the only valid partitions.
+        Assumes "default", "a/b", "a/c-d" and "kernel" are the only valid partitions.
         """
         return [
             "()",
             "(foo)",
             "(bar)/test",
-            "(baz)",
+            "(BAZ)",
+            "(foo)/",
```
cmatsuoka, line 201: I see that `foo` is not a valid partition, why is `(foo)/` it invalid when a similar `(kernel)/` in the previous list is valid?

mr-cal: This is a confusing name for a fixture.  

I'll update the docstring. These filepaths are not necessarily violating the naming convention for partitions but the partitions here are unknown (and thus invalid) to a LifecycleManager was created with the partitions `default`, `a/b`, `a/c-d`, and `kernel`.

tests/unit/features/partitions/test_parts.py
```diff
@@ -130,39 +130,94 @@ def valid_filesets(self):
...
+            "test/(a/b)/test",
+            "(a/c-d)",
+            "(a/c-d)/",
+            "(a/c-d)//",
+            "(a/c-d)/test",
+            "(a/c-d)//test",
+            "(a/c-d)/test/(otherdir)",
+            "(a/c-d)/test/(otherdir-123)",
+            "(a/c-d)/test/(default)",
+            "(a/c-d)/test/(kernel)",
```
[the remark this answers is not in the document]
mr-cal: Correct.

And the filetree representation would be:
```
a
└─ c-d
   └─ test
       └─ (kernel)
```

tests/unit/features/partitions/test_parts.py
mr-cal: The latter is correct.  Does [Alex's comment](https://github.com/canonical/craft-parts/pull/600/files/6ea1343da63b06c762256fdda9a925986d103b82#r1419456295) answer your question?

tests/unit/features/partitions/executor/test_migration.py
```diff
@@ -23,6 +23,9 @@ class TestFileMigration(test_migration.TestFileMigration):
     """Tests for file migration with partitions enabled."""
 
 
+@pytest.mark.xfail(
+    reason="fails to create envvars for namespaced partitions, will be fixed by #597"
+)
```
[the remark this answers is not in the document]
mr-cal: Oh nice catch!

I'll leave it unstrict since some tests in these imported test suites pass.

And like you said, they are going to be removed in the next PR.

tests/unit/features/partitions/test_parts.py
cmatsuoka: Yes, thanks for the clarification.

cmatsuoka (approved):
Thanks!

PR #601, opened 2023-12-07
feat: upgrade to pydantic 2.x

linostar:

-----

Upgrade craft-parts to pydantic 2.x because it is a dependency of Rockcraft.


linostar:
@cjdcordeiro @cmatsuoka Ready for review.

cjdcordeiro:
nice! pinging @sergiusens and @tigarmo too

setup.py
```diff
@@ -48,14 +48,12 @@ def is_rtd() -> bool:
 install_requires = [
     "overrides",
     "PyYAML",
-    "pydantic>=1.9.0,<2.0.0",
-    "pydantic-yaml[pyyaml]>=0.11.0,<1.0.0",
+    "pydantic>=2.0.0,<3.0.0",
+    "pydantic-yaml[ruamel]>=1.1.0,<=1.2.0",
```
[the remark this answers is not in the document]
linostar: This pydantic2 bump is going to happen to all other craft-* repos as well, so there is going to be a re-release for all.
Can you point out in what area will/can ruamel's different behavior affect the *craft projects?

setup.py
```diff
@@ -48,14 +48,12 @@ def is_rtd() -> bool:
...
     "pyxdg",
     "requests",
     "requests-unixsocket",
-    # See: https://github.com/msabramo/requests-unixsocket/pull/69
-    # When updating to urllib3 v2, also remove the constraint on types-requests.
-    "urllib3<2",  # keep compatible API
+    "urllib3==1.26.0",  # keep compatible API
```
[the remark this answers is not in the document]
linostar: urllib2+ definitely break the tests.
This change is a side effect of testing locally. I'll revert to it was, although it urllib3<2 will install 1.26.x too.

craft_parts/parts.py
```diff
@@ -64,36 +81,39 @@ class PartSpec(BaseModel):
...
         """Check if overlay attributes specified when feature is disabled."""
         if not Features().enable_overlay:
             raise ValueError("overlays not supported")
         return item
 
-    @root_validator(pre=True)
-    def validate_root(cls, values: Dict[str, Any]) -> Dict[str, Any]:
+    @model_validator(mode="after")
+    @classmethod
+    def validate_root(cls, values: Any) -> Any:  # noqa: ANN401
```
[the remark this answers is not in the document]
linostar: Will adjust that. Thanks for pointing it out!

craft_parts/plugins/rust_plugin.py
```diff
@@ -35,13 +36,20 @@
...
+
+
 # A workaround for mypy false positives
 # see https://github.com/samuelcolvin/pydantic/issues/975#issuecomment-551147305
 # The proper fix requires Python 3.9+ (needs `typing.Annotated`)
 if TYPE_CHECKING:
     UniqueStrList = List[str]
 else:
-    UniqueStrList = conlist(str, unique_items=True)
+    UniqueStrList = Annotated[List[str], AfterValidator(_unique_str_list)]
```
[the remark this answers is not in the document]
linostar: Yeah I think Set will be fit-in replacement for UniqueStrList.

craft_parts/packages/snaps.py
```diff
@@ -134,18 +134,22 @@ def get_store_snap_info(self) -> Optional[Dict[str, Any]]:
...
+                            self.name,
+                            http_err.response.status_code,
+                            retry_count,
+                        )
+                        if http_err.response.status_code == 404:  # noqa: PLR2004
+                            raise errors.SnapUnavailable(
+                                snap_name=self.name, snap_channel=self.channel
+                            ) from http_err
+                    else:
+                        logger.debug("No response information in http error.")
```
tigarmo, line 151: I think we should add the rest of the original debug message here (like self.name and retry_count)

tigarmo: why was this changed? seems unrelated to pydantic

setup.py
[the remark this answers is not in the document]
tigarmo: what's the need to go from `pyyaml` to `ruamel`?

craft_parts/parts.py
```diff
@@ -64,36 +81,39 @@ class PartSpec(BaseModel):
...
-
-        validate_assignment = True
-        extra = "forbid"
-        allow_mutation = False
-        alias_generator = lambda s: s.replace("_", "-")  # noqa: E731
+    model_config = ConfigDict(
+        validate_assignment=True,
+        extra="forbid",
+        frozen=True,
+        strict=True,
```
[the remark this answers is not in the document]
tigarmo: specifically, this would break because `source-tag` is a float and not a string right? I think this is "too breaking" of a change: it would break all snaps, rocks, charms etc that use things that look like numbers.

This is also mostly unrelated to migrating to v2: if possible, we should stick to the minimum required to do the update with as few changes as possible.

craft_parts/parts.py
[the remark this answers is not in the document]
tigarmo: shouldn't this be `mode="before"`, to keep the semantics of `pre=True`?

craft_parts/permissions.py
```diff
@@ -48,11 +48,12 @@ class Permissions(BaseModel):
     mode: Optional[str] = None
 
     # pylint: disable=no-self-argument
-    @root_validator(pre=True)
-    def validate_root(cls, values: Dict[Any, Any]) -> Dict[Any, Any]:
+    @model_validator(mode="after")
```
tigarmo, line 51: same here `mode="before"`

tests/integration/plugins/test_application_plugin.py
```diff
@@ -145,7 +145,7 @@ def test_application_plugin_missing_stuff(new_dir, partitions):
         )
 
     assert raised.value.part_name == "foo"
-    assert raised.value.message == "- field 'app-stuff' is required"
+    assert raised.value.message == "- Field required in field 'app-stuff'"
```
tigarmo, line 148: I think we need to handle this better in `from_validation_error()`, because now it looks awkward "Field required in field"

tests/unit/features/overlay/test_parts.py
```diff
@@ -407,17 +407,27 @@ def test_part_unmarshal_type_error(self):
...
+            raised.value.message == "- Input should be a valid string in field 'plugin'"
+        )
+
+    @pytest.mark.parametrize("fileset", ["stage", "prime"])
+    def test_relative_path_validation_slash(self, fileset):
+        with pytest.raises(errors.PartSpecificationError) as raised:
+            Part("foo", {fileset: ["bar", "/baz", "x"]})
+        assert raised.value.part_name == "foo"
+        assert raised.value.message == (
+            "- Value error, '/baz' must be a relative path (cannot start with "
```
tigarmo, line 420: If possible I think we should lose this `Value error, ` prefix

tests/unit/test_parts.py
```diff
@@ -407,17 +407,27 @@ def test_part_unmarshal_type_error(self, partitions):
...
+            Part("foo", {fileset: ["bar", "/baz", "x"]}, partitions=partitions)
+        assert raised.value.part_name == "foo"
+        assert raised.value.message == (
+            "- Value error, '/baz' must be a relative path (cannot start with "
+            f"'/') in field '{fileset}'"
+        )
 
     @pytest.mark.parametrize("fileset", ["stage", "prime"])
-    def test_relative_path_validation(self, partitions, fileset):
+    def test_relative_path_validation_empty(self, partitions, fileset):
```
tigarmo, line 425: how come this was split into two tests?

craft_parts/packages/snaps.py
linostar: I carried over from Claudio's experimental branch for pydantic2. I suppose he was getting some http_errors without responses. I didn't test if that's still needed but an extra check that causes no harm.

tests/unit/features/overlay/test_parts.py
linostar: Unfortunately, it is something coming from pydantic2.

tests/unit/test_parts.py
linostar: because `field_validator` decorator, the replacement for the deprecated `validator`, doesn't support `each_item` property. So instead of checking each item in the list with `each_item` automatically, my alternative solution was to check them manually with for loop, which leads to only one malformed field raising an error even in the case of multiple malformed fields. For that purpose, I had to make a separate test for each of the two malformations: one for the empty value and another for the slash-starting value.

I think there should be a way to replicate the old behavior of `each_item` with using Annotated, but not sure if it will need to a different outcome. You can check the current "hacky" implementation here:
https://github.com/linostar/craft-parts/blob/6369ff3d193c35cf21c33035c18b378c6a6a396f/craft_parts/parts.py#L97

craft_parts/permissions.py
linostar: > what's the need to go from pyyaml to ruamel?
According to @cmatsuoka, pydantic2 likes ruamel better and some tests will need adjustments if we need to keep using pyyaml. ruamel on the other hand makes the transition to pydantic2 smoother, and since ruamel supports YAML 1.2 format and also preserves the order of dictionaries and the formatting of the original YAML file: that's an additional advantage. Virtually, switching to ruamel should pose no issue, but @mr-cal raised some concerns about it in the comments above.

tigarmo:
ah I see. That's a minor-ish regression, because it means that we won't
report all validation errors in one go. We should preserve the original
behavior if possible

craft_parts/permissions.py
linostar: > specifically, this would break because source-tag is a float and not a string right? I think this is "too breaking" of a change: it would break all snaps, rocks, charms etc that use things that look like numbers.
This is also mostly unrelated to migrating to v2: if possible, we should stick to the minimum required to do the update with as few changes as possible.

source-tag is a string by definition in Parts class, and in our units tests we only test tags like "v2.3". Is it a common practice for people writing *craft.yaml files to use unquoted numeric values there?
I carried over the "strict" field from Claudio's experimental branch, but if you think that some fields should not be strict-typed due to common practice or other reasons then it should be removed.

tigarmo: I don't know if it's a common practice but I would expect many cases out there to use things like `1.0` for the tag. And source-tag is just one of the properties, I imagine that many things might break with the other properties if we enable this strictness

linostar: > shouldn't this be mode="before", to keep the semantics of pre=True?

You would assume `mode="before"` is just a replacement for `pre=True`, and the documention will only help you believe so! But try to replace that one word "after" with "before" and most unit tests will go red:
image

Apparently, you need to to the model validation after the field validation, because otherwise the model will throw an error because it can't find the fields that weren't validated yet!

tests/integration/plugins/test_application_plugin.py
linostar: Indeed. Perhaps `Field required under 'app-stuff'` will sound better?

craft_parts/parts.py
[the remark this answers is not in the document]
linostar: I replied to Tiago's comment below before seeing this comment.
Now I get a better view of this issue, I agree strict=True should be "strictly" reverted.

tests/unit/test_parts.py
tigarmo: ah I see. That's a minor-ish regression, because it means that we won't
report all validation errors in one go. We should preserve the original
behavior if possible

tests/integration/plugins/test_application_plugin.py
tigarmo: well the previous message `field 'app-stuff' is required` looked fine to me!

linostar: Sorry, you're right. I thought the missing field was a field under "app-stuff" and not "app-stuff" itself.

craft_parts/plugins/rust_plugin.py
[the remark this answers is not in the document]
linostar: After reconsideration, `Set` may not be a good fit because order may be important in some fields like `rust_path`.
As for a generic `UniqueList`, there isn't a reference for another type of unique lists used in the project, so I guess it is not necessary for now.

craft_parts/parts.py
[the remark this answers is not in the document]
linostar: I tried using non-class methods in `mode=after` validators as pointed out in the referenced doc, however it seems to be requiring a considerable amount of change in the asserts in our tests for it to work without error. Using class methods works just fine, although the doc implies it shouldn't, as we're using `cast` to get the properties from `values` parameter instead of getting them through `self`.

linostar:
Summary of changes made after the last review:
- changed "Field required in field" message everywhere
- removed `strict` condition from all class configurations
- used `Annotated` to report all validation errors in one go on stage_files and prime_files fields

setup.py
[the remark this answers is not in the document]
lengau: The main thing I can think of is that `ruamel.yaml` uses YAML 1.2 rather than 1.1, and they're not strictly compatible. For example, pyyaml will accept strings like "yes" as booleans, whereas ruamel.yaml keeps them as strings

```
Python 3.11.6 (main, Oct  8 2023, 05:06:43) [GCC 13.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import yaml as pyyaml
>>> pyyaml.safe_load("my-bool: yes")
{'my-bool': True}
>>> from ruamel.yaml import YAML
>>> ruamel_yaml = YAML(typ="safe")
>>> ruamel_yaml.load("my-bool: yes")
{'my-bool': 'yes'}
>>> ruamel_yaml.load("my-bool: true")
{'my-bool': True}
```

Personally I prefer the `ruamel` behaviour, but it would be good to put guardrails on the error messages at minimum.

craft_parts/state_manager/build_state.py
```diff
@@ -16,23 +16,25 @@
...
-from pydantic import validator
+from pydantic import AfterValidator
+from typing_extensions import Annotated
 
 from .step_state import StepState, validate_hex_string
 
+if TYPE_CHECKING:
+    HexString = str
+else:
+    HexString = Annotated[str, AfterValidator(validate_hex_string)]
```
lengau, line 30:
```suggestion
HexString = Annotated[str, AfterValidator(validate_hex_string)]
```

seems to work fine on my machine

craft_parts/plugins/rust_plugin.py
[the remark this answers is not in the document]
lengau: Can you remove the mypy workaround? It seems to be fixed now.

lengau (approved):
I think we can get rid of some of the type checker workarounds now, but I'm otherwise happy with this.

CAVEAT: I don't think we can merge this to main right now. @cmatsuoka how would you like to handle it?

setup.py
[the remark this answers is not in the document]
linostar: I'm not sure if this particular case needs extra guardrails, since it seems that a `yes` is expected to be string `"yes"` anyway in our codebase (see [this example](https://github.com/canonical/craft-parts/blob/0f5ec0ee704ef462e007437bce6c4749b318fab6/tests/unit/plugins/test_plugins.py#L164)).

craft_parts/state_manager/build_state.py
linostar: done

craft_parts/plugins/rust_plugin.py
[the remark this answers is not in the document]
linostar: done

craft_parts/parts.py
```diff
@@ -56,44 +87,44 @@ class PartSpec(BaseModel):
...
+            AfterValidator(_path_be_relative),
+        ]
+    ] = Field(["*"], alias="stage")
+    prime_files: List[
+        Annotated[
+            str,
+            AfterValidator(_path_not_empty),
+            AfterValidator(_path_be_relative),
+        ]
+    ] = Field(["*"], alias="prime")
```
thp-canonical, line 103: As dicts and lists are mutable, wouldn't `Field(default_factory=dict, alias="organize")`, `Field(default_factory=lambda: ["*"], "alias="overlay")`, ... be more correct here?

Getting some new PyRight validation errors when trying to instantiate a `PartSpec()` that just specifies a plugin, like so:

```python3
    foo = PartSpec(plugin="foo")
```

PyRight shows me:

```
error: Arguments missing for parameters "organize", "overlay", "stage", "prime" (reportCallIssue)
```

linostar: Since you already proposed a fix in a new PR, I'll resolve this here.

lengau:
Closing as superseded :-)

PR #607, opened 2023-12-08
fix(partitions): use base prime dir for getting primed stage packages

mr-cal:

-----

This bug came up when testing with snapcraft.

I'm not sure exactly how to best test it.


craft_parts/executor/part_handler.py
```diff
@@ -418,7 +418,7 @@ def _run_prime(
             and is_deb_based()
         ):
             primed_stage_packages = _get_primed_stage_packages(
-                contents.files, prime_dir=self._part.prime_dir
+                contents.files, prime_dir=self._part.base_prime_dir
```
cmatsuoka, line 421: This is something we didn't think about, should primed packages list be global or per partition? My gut feeling is that we should do it per partition, and if global is more appropriate the application could merge the lists.

mr-cal:
I'm closing the PR but keeping the branch since the snapcraft branch `feature/components` currently points to this branch.

Once #620 is resolved, I can delete this branch and snapcraft can point to the craft-parts branch `feature/namespaced-partitions`.

PR #609, opened 2023-12-13
chore(docs): cleanup sphinx warnings

tigarmo:
This commit cleans up the warnings generated when building the Sphinx docs, and makes the Readthedocs build fail if any new warnings are generated. For pydantic.ErrorDict specifically, the only way I could find to silence those warnings was to altogether drop them before they are emitted, unfortunately.

-----


craft_parts/lifecycle_manager.py
```diff
@@ -80,7 +80,7 @@ class LifecycleManager:
         enabled. The first partition must be "default" and all partitions must be
         lowercase alphabetical.
     :param custom_args: Any additional arguments that will be passed directly
-        to :ref:`callbacks<callbacks>`.
+        to callbacks.
```
tigarmo, line 83: for the life of me I could not get a reference to the auto-generated docs to work

cmatsuoka (approved):

PR #617, opened 2023-12-15
test: make tests work in python 3.12+

lengau:

-----

### DIFF
--- a/tests/unit/plugins/test_base.py
+++ b/tests/unit/plugins/test_base.py
@@ -78,12 +78,17 @@ def test_abstract_methods(new_dir):
     project_info = ProjectInfo(application_name="test", cache_dir=new_dir)
     part_info = PartInfo(project_info=project_info, part=part)
 
-    with pytest.raises(TypeError) as raised:
-        FaultyPlugin(properties=None, part_info=part_info)  # type: ignore[reportGeneralTypeIssues]
-    assert str(raised.value) == (
-        "Can't instantiate abstract class FaultyPlugin with abstract methods "
-        "get_build_commands, get_build_environment, get_build_packages, get_build_snaps"
+    expected = (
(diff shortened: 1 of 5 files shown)


syu-w (approved):

tests/unit/plugins/test_base.py
```diff
@@ -78,13 +78,18 @@ class FaultyPlugin(Plugin):
...
 
-    with pytest.raises(TypeError) as raised:
-        FaultyPlugin(properties=None, part_info=part_info)  # type: ignore[reportGeneralTypeIssues]
-    assert str(raised.value) == (
-        "Can't instantiate abstract class FaultyPlugin with abstract methods "
-        "get_build_commands, get_build_environment, get_build_packages, get_build_snaps"
+    expected = (
+        r"^Can't instantiate abstract class FaultyPlugin with(out an implementation "
+        r"for)? abstract methods '?get_build_commands'?, '?get_build_environment'?, "
+        r"'?get_build_packages'?, '?get_build_snaps'?$"
```
syu-w, line 84: Not sure in this case we want to try match all of them

tests/unit/plugins/test_base.py
```diff
@@ -78,13 +78,18 @@ class FaultyPlugin(Plugin):
...
     )
 
+    with pytest.raises(
+        TypeError,
+        match=expected,
+    ):
```
cmatsuoka, line 90: Can we have that in a single line?

cmatsuoka (commented):
Thanks! (but it seems that we have a linter error)

tests/unit/plugins/test_base.py
lengau: Done!

cmatsuoka (approved):
Thanks!

PR #618, opened 2023-12-19
fix(plugins): add proxy support to ant plugin (#579)

cmatsuoka:
Honor the proxy environment variable in ant projects.

-----

### DIFF
--- a/craft_parts/plugins/ant_plugin.py
+++ b/craft_parts/plugins/ant_plugin.py
@@ -17,7 +17,10 @@
 """The Ant plugin."""
 
 import logging
-from typing import Any, Dict, List, Optional, Set, cast
+import os
+import shlex
+from typing import Any, Dict, Iterator, List, Optional, Set, cast
+from urllib.parse import urlsplit
 
 from overrides import override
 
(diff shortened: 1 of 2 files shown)


kubiko (approved):
LGTM
thank you @cmatsuoka!

kubiko:
@cmatsuoka you can re-run the test, it should pass now I have updated my LP account…

cmatsuoka:
It's still failing for some reason, it seems that there's an email address still not in Launchpad? I can also amend the commit if needed.

syu-w (approved):

kubiko:
> @kubiko It's still failing for some reason, it seems that there's an email address still not in Launchpad? I can also amend the commit if needed.

@cmatsuoka that's strange I have added that email to LP account as well.
I have now changed the primary email to another email used by LP, let's see if that helps.

syu-w:
@kubiko No, it still failed. Notice it shows `ondra@ondra.org` in the checking.

kubiko:
@cmatsuoka hmm that's odd.
OK can you try once more?
I have signed again explicitly with that email.

syu-w:
Let's wait for a while, it may needs sometime to update. As for now it still failed.

craft_parts/plugins/ant_plugin.py
```diff
@@ -142,6 +145,15 @@ def get_build_packages(self) -> Set[str]:
     @override
     def get_build_environment(self) -> Dict[str, str]:
         """Return a dictionary with the environment to use in the build step."""
+        # Getting ant to use a proxy requires a little work; the JRE doesn't
+        # help as much as it should.  (java.net.useSystemProxies=true ought
+        # to do the trick, but it relies on desktop configuration rather
+        # than using the standard environment variables.)
```
lengau, line 151: Thanks for this comment! My first question here was going to be whether we can put this upstream in the general Java plugin. (And perhaps at some point we can provide a lot of this logic in the java plugin, such as retrieving the proxy options, but it's probably a good idea to let it simmer here for a bit even if we do want to do that.)

lengau:
@syu-w it looks like it's succeeding now. The output text from the CLA checker is non-obvious, but what it's doing at this point is checking @kubiko 's account (which is not tied to a CLA signing for some reason) and then falling back to the email address attached to his commits and succeeding there.

PR #623, opened 2024-01-16
fix(part handlers): return files list after override

syu-w:
Override steps should return all files like the non-override steps. Otherwise the callbacks will not able to know which files they should process. (like patchelf)

Fix: https://github.com/snapcore/snapcraft/issues/4508

### DIFF
--- a/craft_parts/executor/part_handler.py
+++ b/craft_parts/executor/part_handler.py
@@ -22,7 +22,7 @@ import os.path
 import shutil
 from glob import iglob
 from pathlib import Path
-from typing import Any, Callable, Dict, List, Optional, Sequence, Set, cast
+from typing import Any, Callable, Dict, List, Optional, Sequence, Set, Tuple, cast
 
 from typing_extensions import Protocol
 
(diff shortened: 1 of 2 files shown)


craft_parts/executor/part_handler.py
```diff
@@ -483,7 +483,8 @@ def _run_step(
                 step=step_info.step,
                 work_dir=work_dir,
             )
-            return StepContents()
+            files, dirs = _get_files(work_dir)
```
tigarmo, line 486: I'm not opposed to this behavior "conceptually" but this will include files and directories that were primed by other parts too, right? Is that the current behavior?

tigarmo (commented):
can you check what happens currently, callback-wise, when you have more than one part?

syu-w:
> can you check what happens currently, callback-wise, when you have more than one part?

Yes, it got called multiple times from each prime steps. But it does not seems broken anything for `patchelf`.

craft_parts/executor/part_handler.py
syu-w: You mean `craftctl` also needs to get all files list?

cmatsuoka: It sounds better to have it on the server side of the control API where the built-in action is executed. Maybe the step handler could hold the results to be recovered after the client execution?

PR #624, opened 2024-01-16
test: use `new_dir` fixture

mr-cal:

-----

- `tests/unit/packages/test_base.py::TestOriginStagePackage::test_mark_origin_stage_package` was creating `.tests-xattr-test-dir/` in the cwd
- `tests/unit/overlays/test_chroot.py::TestChroot::test_runner` was creating `foo.txt` in the cwd

### DIFF
--- a/tests/unit/overlays/test_chroot.py
+++ b/tests/unit/overlays/test_chroot.py
@@ -46,6 +46,7 @@ def fake_conn():
     return FakeConn()
 
 
+@pytest.mark.usefixtures("new_dir")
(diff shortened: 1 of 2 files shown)


lengau (approved):
LGTM!

As a side note: I wonder whether it'd be worth adding pyfakefs to craft-parts's test deps? These tests aren't the best example of its benefits, but I'd bet we could speed up our testing overall by using it.

tigarmo (approved):

PR #626, opened 2024-01-22
feat(rust_plugin): use `rustup` snap and more options

liushuyu:

-----

This pull request adds a few more options to the existing Rust plugin. It also switches to the `rustup` Snap instead of the ad-hoc installation method.

### DIFF
--- a/craft_parts/plugins/rust_plugin.py
+++ b/craft_parts/plugins/rust_plugin.py
@@ -53,6 +53,7 @@ class RustPluginProperties(PluginProperties, PluginModel):
     rust_no_default_features: bool = False
     rust_ignore_toolchain_file: bool = False
     rust_cargo_parameters: List[str] = []
+    rust_inherit_ldflags: bool = False
     source: str
     after: Optional[UniqueStrList] = None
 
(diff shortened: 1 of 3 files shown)


liushuyu:
The CI is failing because the Python plugin is bugging out

craft_parts/plugins/rust_plugin.py
```diff
@@ -53,9 +51,29 @@ class RustPluginProperties(PluginProperties, PluginModel):
     rust_channel: Optional[str] = None
     rust_use_global_lto: bool = False
     rust_no_default_features: bool = False
+    rust_ignore_toolchain_file: bool = False
+    rust_verbose_output: bool = False
```
cmatsuoka, line 55: Certain plugins have the `<plugin name>-parameters` option to append to the relevant command line, can we use that for verbose output instead? We could use `rust-cargo-parameters` or similar and append to `config_cmd` below, and this could be used for other simple parameters.

liushuyu: Done

lengau (approved):
Looks great! Thanks!

cmatsuoka (approved):
Thanks!

PR #628, opened 2024-01-23
build(deps): pin overrides!=7.6.0

mr-cal:

-----

`overrides` 7.6.0 introduced a bug: https://github.com/mkorpela/overrides/issues/121

A handful of python plugin tests are failing with:
```
FAILED tests/integration/plugins/test_python.py::test_python_plugin_no_system_interpreter[True] - TypeError: test_python_plugin_no_system_interpreter.<locals>.MyPythonPlugin._get_system_python_interpreter: No super class method found
```

due to how the tests override superclass functions:
```
    class MyPythonPlugin(craft_parts.plugins.plugins.PythonPlugin):
        @override
        def _get_system_python_interpreter(self) -> Optional[str]:
            return None
```

Our options are:
1. Merge this PR
2. Use the workaround described in the bug linked above
3. Wait for a fix in `overrides`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,7 +46,8 @@ def is_rtd() -> bool:
 
 
 install_requires = [
-    "overrides",
+    # see https://github.com/mkorpela/overrides/issues/121
+    "overrides<7.6",
     "PyYAML",
     "pydantic>=1.9.0,<2.0.0",
     "pydantic-yaml[pyyaml]>=0.11.0,<1.0.0",
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,7 @@ def is_rtd() -> bool:
 
 install_requires = [
     # see https://github.com/mkorpela/overrides/issues/121
-    "overrides<7.6",
+    "overrides!=7.6.0",
     "PyYAML",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -46,7 +46,8 @@ def is_rtd() -> bool:
 
 
 install_requires = [
-    "overrides",
+    # see https://github.com/mkorpela/overrides/issues/121
+    "overrides<7.6",
```
lengau, line 50: I'm guessing this is sufficiently breaking that it'll be fixed in 7.6.1, so maybe we want:

```suggestion
    "overrides!=7.6.0",
```

lengau (approved):
Option 4: use `typing-extensions` to bring in the `override` decorator from [PEP 698](https://peps.python.org/pep-0698/).

(It looks like core24 is on track to get Python 3.12 by default so we might want to move in that direction anyway)

My personal opinion:

1. Merge this now as a short-term fix
2. Assess whether the PEP 698 override decorator meets all of our needs.

mr-cal:
Not sure why readthedocs failed, closing and re-opening

cmatsuoka (approved):

PR #637, opened 2024-01-25
feat: add a utility function that determines whether a part has slices

linostar:

-----
A utility function is needed in rockcraft to determine whether parts has slices in them. 
Instead of implementing such a function in rockcraft, it is better to implement it in craft-parts instead.

### DIFF
--- a/craft_parts/parts.py
+++ b/craft_parts/parts.py
@@ -99,12 +99,9 @@ class PartSpec(BaseModel):
             # This check is only relevant in deb systems.
             return values
 
-        def is_slice(name: str) -> bool:
-            return "_" in name
-
         # Detect a mixture of .deb packages and chisel slices.
         stage_packages = values.get("stage-packages", [])
-        has_slices = any(name for name in stage_packages if is_slice(name))
+        has_slices = part_has_slices(values)
         has_packages = any(name for name in stage_packages if not is_slice(name))
 
         if has_slices and has_packages:
@@ -640,6 +637,23 @@ def part_has_overlay(data: Dict[str, Any]) -> bool:
     spec = _get_part_spec(data)
 
     return spec.has_overlay
+
+
+def part_has_slices(data: Dict[str, Any]) -> bool:
+    """Whether the part described by ``data`` has slices in its stage-packages.
+
+    :param data: The part data to query.
+    """
+    stage_packages = data.get("stage-packages", [])
+    return any(name for name in stage_packages if is_slice(name))
+
+
+def is_slice(name: str) -> bool:
+    """Whether the stage-package is a slice or a package.
+
+    :param name: name of the package.
+    """
+    return "_" in name
 
 
 def _get_part_spec(data: Dict[str, Any]) -> PartSpec:


linostar:
@tigarmo Ready for review.

linostar:
Tagging @cmatsuoka @lengau @mr-cal for review.

Failing integreation test is due to network connection issue so it needs a retrigger.

SamirPS (approved):

craft_parts/parts.py
```diff
@@ -642,6 +639,23 @@ def part_has_overlay(data: Dict[str, Any]) -> bool:
     return spec.has_overlay
 
 
+def part_has_slices(data: Dict[str, Any]) -> bool:
+    """Whether the part described by ``data`` has slices in its stage-packages.
```
cmatsuoka, line 643: How and when is this going to be accessed? It feels that it should be a part property instead of a call receiving a list of part names.

linostar: This is needed for Rockcraft's [PR 464](https://github.com/canonical/rockcraft/pull/464).
It will work as a part property as well, and I don't have a particular preference for either. However, there is already a similar utility function called `part_has_overlay`, so I put this one in the same scope.

craft_parts/parts.py
```diff
@@ -642,6 +639,23 @@ def part_has_overlay(data: Dict[str, Any]) -> bool:
     return spec.has_overlay
 
 
+def part_has_slices(data: Dict[str, Any]) -> bool:
```
tigarmo, line 642: I think it's important to add a basic test for `part_has_slices()` and `is_slice()`. They are currently covered by the existing validator test but if refactor the validator the functions might become uncovered

tigarmo (changes requested):

linostar:
I have changed the utility function to a property, and added unit tests for it. This is to accomodate the function to how similar functions are unit tested.
Note the failing unit-tests checks are not related to this PR.

craft_parts/parts.py
```diff
@@ -169,6 +166,15 @@ def has_overlay(self) -> bool:
...
         )
 
+    @property
+    def has_slices(self) -> bool:
+        """Whether the part has slices in its stage-packages.
+
+        :param data: The part data to query.
+        """
+        stage_packages = self.stage_packages or []
+        return any(name for name in stage_packages if is_slice(name))
```
lengau, line 176:
```suggestion
        return any(is_slice(name) for name in stage_packages)
```

lengau (approved):
Code looks good, thanks :-)

lengau:
@linostar could you rebase this on main and resolve the conflicts here?

SamirPS (approved):

craft_parts/parts.py
```diff
@@ -175,6 +172,15 @@ def has_overlay(self) -> bool:
             or self.overlay_files != ["*"]
         )
 
+    @property
+    def has_slices(self) -> bool:
+        """Whether the part has slices in its stage-packages.
+
+        :param data: The part data to query.
+        """
+        stage_packages = self.stage_packages or []
```
cmatsuoka, line 181: I would test for `self.stage_packages` and short-circuit the return value here.

cmatsuoka (commented):
LGTM, maybe we could make it slightly more performant with a short-circuit return in `has_slice`.

linostar:
@cmatsuoka @tigarmo Can we merge this PR if there are no further comments, so that we can move forward with https://github.com/canonical/rockcraft/pull/464?

SamirPS (approved):

linostar:
This is closed in favor of #904.

PR #654, opened 2024-02-08
docs: Link filesets specification from part properties

thp-canonical:
Was looking for how the `prime` and `stage` part properties take glob-like inclusion and exclusion patterns, but the docs didn't point to the fact that these are "filesets".

Add a link to the fileset specification explanation, so that people looking for all the possible fileset-specific ways of specifying files to include and exclude can find it easily.

Also, "at least item" -> "at least 1 item"

### DIFF
--- a/docs/base/part_properties.rst
+++ b/docs/base/part_properties.rst
@@ -186,7 +186,8 @@ prime
 
 **Step:** prime
 
-The files to copy from the staging area to the priming area.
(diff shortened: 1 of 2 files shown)


sergiusens (approved):

lengau:
Unit test failures are unrelated (https://github.com/canonical/craft-parts/issues/661)

lengau (approved):

PR #662, opened 2024-02-13
feat(executor): check for stage file collisions in each partition

### LINKED ISSUE #633: Check for staged file collisions in each partition
### What needs to get done

The [`check_for_stage_collisions()`](https://github.com/canonical/craft-parts/blob/5ad7be9c186f577d813e416330913497f868ee3b/craft_parts/executor/collisions.py#L30) function needs to check for file collisions in each partition.

### Why it needs to get done

Currently, craft-parts is only checking for collisions in the default partition.

mr-cal:

-----

Check for collisions where multiple parts try to stage a file with different contents in each partition.

I refactored the existing unit tests and added new integration tests.  I recommend reviewing per-commit.

Fixes #633 
(CRAFT-2416)

### DIFF
--- a/craft_parts/executor/collisions.py
+++ b/craft_parts/executor/collisions.py
@@ -28,14 +28,52 @@ from craft_parts.parts import Part
 from craft_parts.permissions import Permissions, permissions_are_compatible
 
 
-def check_for_stage_collisions(part_list: List[Part]) -> None:
+def check_for_stage_collisions(
+    part_list: List[Part], partitions: Optional[List[str]]
+) -> None:
     """Verify whether parts have conflicting files to stage.
 
-    :param part_list: The list of parts to be tested.
+    If the partitions feature is enabled, then check if parts have conflicting files to
+        stage for each partition.
+    If the partitions feature is disabled, only check for conflicts in the default
+        stage directory.
+
+    :param part_list: The list of parts to check.
+    :param partitions: An optional list of partition names.
+
+    :raises PartConflictError: If conflicts are found.
+    :raises FeatureError: If partitions are specified but the feature is not enabled or
+        if partitions are not specified and the feature is enabled.
+    """
+    if partitions and not Features().enable_partitions:
+        raise errors.FeatureError(
+            "Partitions specified but partitions feature is not enabled."
+        )
+
+    if partitions is None and Features().enable_partitions:
+        raise errors.FeatureError(
+            "Partitions feature is enabled but no partitions specified."
+        )
+
+    if partitions:
+        for partition in partitions:
+            _check_for_stage_collisions_per_stage_directory(part_list, partition)
+    else:
+        _check_for_stage_collisions_per_stage_directory(part_list, None)
+
+
+def _check_for_stage_collisions_per_stage_directory(
+    part_list: List[Part], partition: Optional[str]
+) -> None:
+    """Verify whether parts have conflicting files for a single stage directory.
+
+    :param part_list: The list of parts to check.
+    :param partition: If the partitions feature is enabled, then the name of the
+        partition containing the stage directory to check.
+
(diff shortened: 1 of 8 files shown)


craft_parts/executor/collisions.py
```diff
@@ -28,22 +28,60 @@
...
+        raise errors.FeatureError(
+            "Partitions specified but partitions feature is not enabled."
+        )
+
+    if partitions is None and Features().enable_partitions:
+        raise errors.FeatureError(
+            "Partitions feature is enabled but no partitions specified."
+        )
+
+    if partitions:
```
mr-cal, line 58: This can be written as

```python
 for partition in partitions or [None]:
  _check_for_stage_collisions...
```

but requires silencing pyright.

tests/unit/features/partitions/executor/test_collisions.py
```diff
@@ -0,0 +1,43 @@
...
+from tests.unit.executor import test_collisions
+
+
+class TestCollisions(test_collisions.TestCollisions):
+    """Check collision scenarios with partitions enabled."""
+
+
+class TestCollisionsPartitionError:
+    def test_partitions_enabled_but_not_defined(self, partitions, tmpdir):
+        """Raise an error if partitions are defined but not enabled."""
```
lengau, line 30:
```suggestion
        """Raise an error if partitions are enabled but not defined."""
```

tests/integration/features/partitions/executor/test_collisions.py
```diff
@@ -0,0 +1,167 @@
...
+        )
+
+        with pytest.raises(PartFilesConflict) as raised:
+            with lcm.action_executor() as aex:
+                aex.execute(lcm.plan(Step.STAGE))
+
+        assert str(raised.value) == (
+            "Failed to stage: parts list the same file with different contents or permissions.\n"
+            "Parts 'part2' and 'part1' list the following files, but with different contents or permissions:\n"
+            "    file"
```
lengau, line 166: Can we have this provide the partition name (when enabled) too? It could speed up debugging for more complex projects to have both a set of parts and a destination partition to look into, especially if the user scripts the partition assignments.

lengau (changes requested):
The scrollbar was daunting at first, but when I actually read through it, it was a very elegant implementation combined with some great, thorough testing.

Changes requested are a typo fix and a quality of life request

tests/integration/features/partitions/executor/test_collisions.py
mr-cal: Great idea!

lengau (approved):
Thanks!

craft_parts/executor/collisions.py
```diff
@@ -28,22 +28,60 @@
...
+        for partition in partitions:
+            _check_for_stage_collisions_per_stage_directory(part_list, partition)
+    else:
+        _check_for_stage_collisions_per_stage_directory(part_list, None)
+
+
+def _check_for_stage_collisions_per_stage_directory(
+    part_list: List[Part], partition: Optional[str]
```
cmatsuoka, line 66: Passing the path directly as an argument to the function could make the internal logic easier to follow, and avoid passing `None` to be used as a dictionary key.

mr-cal: I can not figure out how to implement your suggestion!  

I can't pass a single directory to this function; it needs more than just a single path.  It needs `part_install_dir[partition]` for all parts and it also needs the partition name for calls like `filesets.migratable_filesets(stage_fileset, srcdir, partition)`.

cmatsuoka: You're right, you're iterating over all parts there. Thanks for the clarification.

cmatsuoka (approved):

PR #663, opened 2024-02-15

### LINKED ISSUE #650: Include all partitions by default for `stage` and `prime` keywords
### What needs to get done

#### Exclusion example
```yaml
parts:
  my-part:
    organize:
      a: (foo)/a
      b: (bar)/b
    stage:
      - -(foo)/a/c
```

#### Inclusion example
```yaml
parts:
  my-part:
    organize:
      a: (foo)/a
      b: (bar)/b
    stage:
      - (foo)/a/c
```

### Why it needs to get done

To align with the spec.

```yaml
parts:
  foo:
    plugin: dump
    source: .
    organize: 
      (default)/file: (mypart)/file
    stage:
     - (default)/*
     - (mypart)/*
    prime:
     - (default)/*
     - (mypart)/*
```

mr-cal:

-----

When partitions are enabled, all content will be staged and primed in all partitions by "default".

I put "default" in quotes because I'm not setting the default in the [PartSpec pydantic model](https://github.com/canonical/craft-parts/blob/c11570597d4a397aba1744fa08adcd344682246e/craft_parts/parts.py#L59).  I can't set the default value outright, but I can modify the data with a `@validator` function.  I would probably validate the usage of partitions inside the `PartSpec` model as well.  

As a final confusing note, I think @lengau and I both implemented the same validation in different ways: [here](https://github.com/canonical/craft-parts/blob/c11570597d4a397aba1744fa08adcd344682246e/craft_parts/lifecycle_manager.py#L158) and [here](https://github.com/canonical/craft-parts/blob/c11570597d4a397aba1744fa08adcd344682246e/craft_parts/lifecycle_manager.py#L162).

Fixes #650

### DIFF
--- a/craft_parts/lifecycle_manager.py
+++ b/craft_parts/lifecycle_manager.py
@@ -35,6 +35,7 @@ from craft_parts.parts import Part, part_by_name, validate_partition_usage
 from craft_parts.state_manager import states
 from craft_parts.steps import Step
 from craft_parts.utils.partition_utils import validate_partition_names
+from craft_parts.utils.path_utils import get_partition_and_path
 
 
 class LifecycleManager:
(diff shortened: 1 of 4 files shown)


mr-cal:
Closing in favor of #664

PR #664, opened 2024-02-17

mr-cal:

-----

### chore: redesign partition validators

This commit combines two redundant validators and changes where partition usage in parts is validated.  

Originally, I added a [validator](https://github.com/canonical/craft-parts/pull/493/files#diff-68d2086e111409ee4298829260c916b031d71f8b27081f83b135ce59d5ff8a66R152) when the `LifecycleManager` was initialized.  Then Alex added [another](https://github.com/canonical/craft-parts/pull/517/files#diff-68d2086e111409ee4298829260c916b031d71f8b27081f83b135ce59d5ff8a66R159) (and better) validator also during `LifecycleManager` initialization based on this [prototype](https://github.com/canonical/craft-parts/commit/075eac2b04b70f97388faeee74a5679fa2c8c92c#diff-68d2086e111409ee4298829260c916b031d71f8b27081f83b135ce59d5ff8a66R393).

But things have changed.  Now the `partitions` are provided to a part, so validation can occur when the part is initialized.  This paves the way for the next commit.

### fix: stage and prime content in all partitions by default

This is similar to #663, but the processing occurs when the part is initialized and after the filter keywords are validated.  It also uses a more complex data structure that later gets flattened out.


craft_parts/parts.py
```diff
@@ -384,42 +429,110 @@ def check_partition_usage(self, partitions: List[str]) -> List[str]:
...
+            return fileset_mapping
+
+        # sort fileset entries into partitions
+        for entry in fileset:
+            # split file into an optional prefix (a hyphen character) and the file
+            split_entry = (entry[0], entry[1:]) if entry[0] == "-" else ("", entry)
+
+            partition, inner_path = get_partition_and_path(split_entry[1])
+            fileset_mapping[partition].append(
+                f"{split_entry[0]}({partition})/{inner_path}"
```
cmatsuoka, line 527: Can't this be simply `entry`?

craft_parts/parts.py
```diff
@@ -365,17 +372,55 @@ def has_overlay(self) -> bool:
...
+    @property
+    def stage_files(self) -> List[str]:
+        """Get stage files for a part after setting defaults for all partitions.
+
+        TODOs:
+          - self.stage_files conflicts with self.spec.stage_files
+          - should this return a flattened data structure or a dictionary?
+            - the latter would simplify the codebase but require refactoring
+        """
+        return list(itertools.chain.from_iterable(self._stage_files_mapping.values()))
```
cmatsuoka, line 384: Why do you need to regroup split entries instead of using the original listing?

cmatsuoka:
Nice work! Added some questions about reusing existing data, but overall it looks good.

mr-cal:
Closing in favor of a different solution

PR #667, opened 2024-02-20
fix: make overlay and partition features mutually exclusive

### LINKED ISSUE #632: Make partitions and overlay features mutually exclusive
### What needs to get done

Require partitions and overlays to be mutually exclusive features.

### Why it needs to get done

The first implementation of partitions (https://github.com/canonical/craft-parts/pull/517) allowed both features to be enabled.  It added some basic integration tests, but they were not comprehensive.  The same integration tests pass after partitions were [moved to a top-level directory](https://github.com/canonical/craft-parts/pull/627).

However, @lengau, @cmatsuoka, and @mr-cal are concerned that enabling both these features does not have a well-defined behavior.  There is no immediate need for an application to have both features enabled, so we prefer to make them mutually exclusive until the behavior is fully defined and agreed upon.

mr-cal:

-----

Make `overlay` and `partitions` mutually exclusive features.

This is a "fix" because they are orthogonal and were never intended to be used together.

If an application ever needs both features, we will have to start with a spec.

Fixes #632 
(CRAFT-2415)


craft_parts/features.py
```diff
@@ -36,6 +37,24 @@ class Features(metaclass=Singleton):
...
+        self, *, enable_overlay: bool = False, enable_partitions: bool = False
+    ) -> None:
+        """Initialize craft-parts features.
+
+        :param enable_overlay: Enables the overlay step.
+        :param enable_partitions: Enables the usage of partitions.
+
+        :raises FeatureError: If mutually exclusive features are enabled.
+        """
+        if enable_overlay and enable_partitions:
```
lengau, line 50: Since this is a dataclass, this might be better to do in `__post_init__`. As a side benefit we won't have to touch `__init__` when we add other features.

lengau (changes requested):

craft_parts/features.py
mr-cal: Ack, great idea!

lengau (approved):

cmatsuoka (approved):
Thanks!

PR #668, opened 2024-02-21

### LINKED ISSUE #650: Include all partitions by default for `stage` and `prime` keywords

mr-cal:

-----

@cmatsuoka and I have found a much better solution.  Instead of changing the default values of stage and prime keywords, we changed the behavior of the fileset that consumes the stage and prime keywords.

Depends on #669
Fixes #650
(CRAFT-2458)


mr-cal:
Moving to draft, @cmatsuoka and I decided to make the following scenarios errors:
```yaml
stage:
  - (default)
  - (foo)

stage:
  - (default)/
  - (foo)/
```

lengau (approved):

craft_parts/parts.py
```diff
@@ -464,6 +476,13 @@ def _check_partitions_in_filepaths(
                             f"    misused partition {partition!r} in {filepath!r}"
                         )
 
+            if require_inner_path:
+                _, inner_path = get_partition_and_path(filepath)
+                if not inner_path:
+                    error_list.append(
+                        f"    no path specified after partition in {filepath!r}"
```
sergiusens, line 483: this might be a todo for later, but I would have made the error list about the data and left the formatting to the exception that renders it as a string

sergiusens: please create a task for this

sergiusens (approved):
looks pretty good!

craft_parts/parts.py
mr-cal: Created #670

mr-cal:
This will be ready to land after #669 gets another review.

tests/unit/features/partitions/executor/test_filesets.py
```diff
@@ -156,18 +154,39 @@ def test_partition_filesets(entries, includes, excludes):
...
+        ("mypart", ["(mypart)/*"], ["*"], []),
+        ("mypart", ["(mypart)/foo"], ["foo"], []),
+        # non-default partition, exclusive
+        ("default", ["-(mypart)/*"], ["*"], []),
+        ("default", ["-(mypart)/foo"], ["*"], []),
+        ("mypart", ["-(mypart)/*"], ["*"], ["*"]),
+        ("mypart", ["-(mypart)/foo"], ["*"], ["foo"]),
+        # no partitions
+        ("default", [], ["*"], []),
+        ("mypart", [], ["*"], []),
```
cmatsuoka, line 189: nice set of test cases!

cmatsuoka (approved):
Looks good, thanks!

PR #669, opened 2024-02-21
refactor: redesign partition validators

### LINKED ISSUE #650: Include all partitions by default for `stage` and `prime` keywords

mr-cal:

-----

There were two separate instances of validating partitions in two places. Both were run when intializing the LifecycleManager.

Since then, partitions are now passed to the part during initialization. This allows us to validate earlier when the part is initialized.

I have combined the best aspects of both validators.

This is a prerequisite to resolve #650 where I will need to add extra validation when the part is created.


lengau (approved):
🚀

mr-cal:
Merged indirectly via #668  (I meant for this to get merged first but it's OK)

PR #676, opened 2024-03-01
docs(plugins): add dump plugin docs

syu-w:


lengau:
Unit tests should be fixed in https://github.com/canonical/craft-parts/pull/678

docs/explanation/dump_plugin.rst
```diff
@@ -0,0 +1,42 @@
...
+which branch, tag, and/or commit to clone if it is a repository.
+
+
+If you are not using the source files directly in the final payload, but want to
+run some custom commands to generate them. Then you should use the ``nil``
+plugin instead of this dump plugin to avoid copying any unnecessary files.
+
+
+Combining the dump plugin with the :ref:`part properties <part_properties>`
+allows to extend the behavior to create new files, modify existing files,
```
lengau, line 25:
```suggestion
allows extending the behavior to create new files, modify existing files,
```

docs/base/dump_plugin.rst
```diff
@@ -0,0 +1,43 @@
...
+
+
+Keywords
+--------
+
+This plugin uses the common :ref:`plugin <part-properties-plugin>` keywords as
+well as those for :ref:`sources <part-properties-sources>`.
+
+You will need to specify at least the :ref:`source <source>` keyword.
+:ref:`source-type <source_type>` keyword is optional but recommended for
```
tigarmo, line 22: I think instead of "for specific" it should be "to specify"

docs/base/dump_plugin.rst
```diff
@@ -0,0 +1,43 @@
...
+Dependencies
+------------
+
+This plugin has no dependencies.
+
+
+How it works
+------------
+
+During the build step, the plugin performs the following actions:
```
tigarmo, line 35: I think this list is misleading: as far as I know every bullet point [except the last](https://github.com/canonical/craft-parts/blob/main/craft_parts/plugins/dump_plugin.py#L82) is actually handled by the source type, and not the plugin. That is, almost everything here is shared by all plugins, including the `nil` plugin

docs/base/dump_plugin.rst
```diff
@@ -0,0 +1,43 @@
...
+
+* Check the ``source-type`` keyword to determine the type of the source if
+  specified, otherwise, it will try to guess the type based on the ``source``.
+* Download the file or clone the repository if the ``source`` is a remote
+  location.
+* Copy the file or directory if the ``source`` is a local location.
+* Unpack the file if it is an archive specified by the ``source-type`` keyword.
+* Copy all contents and preserve the directory structure to the build directory.
```
tigarmo, line 43:
```suggestion
* Copy all contents and preserve the directory structure to the part's install directory.
```

docs/explanation/dump_plugin.rst
```diff
@@ -0,0 +1,42 @@
+.. _dump_plugin_explanation:
+
+Dump Plugin
+===========
+
+The dump plugin is a simple plugin that by default is equivalent to run the
```
tigarmo, line 6:
```suggestion
The dump plugin is a simple plugin that by default is equivalent to running the
```

docs/explanation/dump_plugin.rst
```diff
@@ -0,0 +1,42 @@
...
+    cp --archive --link --no-dereference . "${CRAFT_PART_INSTALL}"
+
+
+The :ref:`source <source>` property is the key to this plugin. With some
+additional properties, they define the location where to get the files from, or
+which branch, tag, and/or commit to clone if it is a repository.
+
+
+If you are not using the source files directly in the final payload, but want to
+run some custom commands to generate them. Then you should use the ``nil``
```
tigarmo, line 20:
```suggestion
run some custom commands to generate them, then you should use the ``nil``
```

docs/explanation/dump_plugin.rst
```diff
@@ -0,0 +1,42 @@
...
+Combining the dump plugin with the :ref:`part properties <part_properties>`
+allows extending the behavior to create new files, modify existing files,
+and/or filter files for the final payload.
+
+
+For example, it could be used with the :ref:`override-build <override_build>`,
+to convert file formats. 
+
+
+.. ifconfig:: project in ["Snapcraft"]
```
tigarmo, line 33: I think we should move away from this kind of project-specific customization in craft-parts. Instead, I think these contents should be in `base/explanation/dump_plugin.rst` so that Snapcraft itself can customize it

docs/how-to/include_files.rst
```diff
@@ -0,0 +1,171 @@
+.. _how_to_include_files_and_resources:
+
+******************************************
+Including local files and remote resources
+******************************************
+
+Craft-parts provides built-in :ref:`dump plugin <craft_parts_dump_plugin>` for
```
tigarmo, line 7:
```suggestion
Craft-parts provides the built-in :ref:`dump plugin <craft_parts_dump_plugin>` for
```

docs/how-to/include_files.rst
```diff
@@ -0,0 +1,171 @@
...
+all kinds of projects that need to include local files and remote resources.
+This plugin is used to download, unpack, and copy files and directories from the
```
tigarmo, line 9: Again, I think this emphasis on the dump plugin doing the downloading, unpacking, etc of the files is misleading because it suggests that the other plugins *don't* do that

docs/how-to/include_files.rst
```diff
@@ -0,0 +1,171 @@
...
+.. code-block:: text
+
+  .
+  ├── misc
+  │   ├── fonts
+  │   │   └── good.otf
+  │   └── services
+  │       ├── README
+  │       └── hello.service
+  └── snap
```
tigarmo, line 40: I don't think we should mention snapcraft here

docs/how-to/include_files.rst
```diff
@@ -0,0 +1,171 @@
...
+      └── snapcraft.yaml
+
+
+The following dump part can be used to include the ``hello.service`` and
+``good.otf`` files from the ``misc`` directory, then move them to the correct
+locations, keeping only ``/usr`` in the final payload:
+
+.. code-block:: yaml
+
+  parts:
```
tigarmo, line 50: :+1:

docs/base/dump_plugin.rst
syu-w: They are shared, but that doesn't affect how to explain to the user what the plugin does. You can't say that this plugin copies files and then doesn't explain to the user how those files got there.

docs/how-to/include_files.rst
syu-w: I've updated some of the wording and it should be clearer

docs/base/dump_plugin.rst
lengau: I'm fine leaving it as-is, but perhaps briefer version would be:

```
During the build step, the plugin copies the entire directory tree from the source to the part's install directory.
```

It is probably out of scope to explain the source type here, and we reference the [part source](https://canonical-craft-parts.readthedocs-hosted.com/en/latest/explanation/parts.html#source) to send more details of how the source directory gets filled.

docs/explanation/dump_plugin.rst
```diff
@@ -0,0 +1,35 @@
...
+run some custom commands to generate them, then you should use the ``nil``
+plugin instead to avoid copying any unnecessary files.
+
+
+Combining the dump plugin with the :ref:`part properties <part_properties>`
+allows extending the behavior to create new files, modify existing files,
+and/or filter files for the final payload.
+
+
+For example, it could be used with the :ref:`override-build <override_build>`,
```
lengau, line 29:
```suggestion
and/or filter files for the final payload. For example, it could be used with the :ref:`override-build <override_build>`,
```

I think this makes more sense as a single paragraph.

docs/explanation/dump_plugin.rst
```diff
@@ -0,0 +1,35 @@
...
+to convert file formats. 
+
+
+The plugin could also be used with the :ref:`organize <organize>` to reorganize
```
lengau, line 33:
```suggestion
The plugin can also be used with the :ref:`organize <organize>` to reorganize
```

"could" here implies that the dump plugin is hypothetical, rather than the situation being a hypothetical on the real dump plugin.

docs/how-to/include_files.rst
```diff
@@ -0,0 +1,178 @@
+.. _how_to_include_files_and_resources:
+
+******************************************
+Including local files and remote resources
+******************************************
+
+Craft-parts provides the built-in :ref:`dump plugin <craft_parts_dump_plugin>`
+for all kinds of projects that need to include local files and remote resources 
+as is.
+This plugin using the :ref:`source <source>` property in the part to download,
```
lengau, line 10:
```suggestion
This plugin uses the :ref:`source <source>` property in the part to download,
```

docs/how-to/include_files.rst
```diff
@@ -0,0 +1,178 @@
...
+
+* Include static files in the final payload, such as scripts, configurations,
+  services, documentation, media files, or other resources that are not
+  generated by the build process.
+
+* Include pre-built third-party packages, libraries, binaries, or other
+  artifacts that are not available in the system's package manager or in the
+  project's source code.
+
+The supported source types please refer to :ref:`source_type`.
```
lengau, line 32:
```suggestion
For supported source types please refer to :ref:`source_type`.
```

docs/how-to/include_files.rst
```diff
@@ -0,0 +1,178 @@
...
+      └── share
+          └── fonts
+              └── good.otf
+
+
+Example: To include a remote third-party deb package from a URL
+---------------------------------------------------------------
+
+The following dump part downloads a ``busybox-static`` deb package from a
+remote URL and includes it in the final payload:
```
lengau, line 87: It might be worth noting that this doesn't bring in any dependencies.

lengau (approved):
I love the examples in the how-to! Those would have been incredibly helpful for me when learning parts.

Attached comments are minor and I'm happy either way. Most of them are just grammatical.

docs/how-to/include_files.rst
syu-w: I thought about it, but I'm not sure if I should use a more complex example.

tigarmo (approved):

PR #680, opened 2024-03-06
refactor(docs) move more docs to the shared base

tigarmo:
The real intent of this PR is the second commit (the "bookend" labels in the Python plugin reference), but in order to use that in Snapcraft we need to add other documents to the shared base, because they are referenced by the plugin docs.


docs/base/explanation/parts.rst
```diff
@@ -0,0 +1,186 @@
...
+Not all of these steps may be needed for every use case, and tools that use
+the Craft Parts framework can skip those that are not appropriate for their
+purposes.
+
+Tools like `Snapcraft`_ and `Charmcraft`_ that use the concepts of parts to
+describe a build process typically accept specifications of parts in YAML format. This allows each part to be described in a convenient,
+mostly-declarative format. Libraries that use parts may use the underlying
+data structures to describe them.
+
+Describing a part
```
lengau, line 34: Thinking about charmcraft docs, these sections would be nice as their own snippets if we were using sphinx.

We're not using sphinx for charmcraft so I'm not requesting anything right now, but it's a possible future improvement to separate them into snippets that can each be bookended with app-specific stuff.

lengau (approved):
Great step forward, thanks!

docs/base/explanation/parts.rst
lengau: Never mind - I see how you're doing it in Rockcraft and that makes a lot of sense :-)

cmatsuoka (approved):
Thanks!

PR #685, opened 2024-03-12
docs: add maven plugin documentation

cmatsuoka:

-----


syu-w (approved):

docs/base/maven_plugin.rst
```diff
@@ -0,0 +1,66 @@
...
+
+
+Dependencies
+------------
+
+The plugin expects Maven to be available on the system as the ``mvn`` executable, unless
+a part named ``maven-deps`` is defined. In this case, the plugin will assume that this
+part will stage the ``mvn`` executable to be used in the build step.
+
+A Java Runtime must be staged if not part of the execution environment.
```
lengau, line 49: Staging packages doesn't have the intended results in Charmcraft. Idk whether it's worth mentioning here, but maybe an alternative text like "This plugin does not make a Java runtime available. This may be handled by installing a Java runtime in the target environment" would cover it well enough? The snapcraft example below shows how to stage a Java runtime.

lengau (approved):

docs/base/maven_plugin.rst
tigarmo: one solution here is to wrap this Dependencies section [like on the python plugin docs](https://github.com/canonical/craft-parts/blob/main/docs/base/python_plugin.rst?plain=1#L62) so that specific tools can give specific guidance on how to provision java

docs/base/maven_plugin.rst
```diff
@@ -0,0 +1,66 @@
...
+
+
+Example
+-------
+
+This is an example of a Snapcraft part using the Maven plugin. Note that the Maven and
```
tigarmo, line 55: I think this Snapcraft-specific example should be in Snapcraft's docs, and then Rockcraft should have an example too (I think this is what we should do for all plugins)

tigarmo (changes requested):

docs/base/maven_plugin.rst
cmatsuoka: Kept a generic plugin operation description here, overriding the section in Snapcraft docs.

docs/base/maven_plugin.rst
cmatsuoka: Moved the example to Snapcraft docs.

mattculler (approved):

tigarmo (approved):

PR #687, opened 2024-03-15
ci: run spelling, woke and linkcheck on docs

cmatsuoka:

-----


.github/workflows/tests.yaml
```diff
@@ -14,14 +14,15 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
+          submodules: true
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
           python-version: "3.10"
```
lengau, line 21: There's a way in the python installer action to cache pip downloads - might be useful here.

I *think* it might even be able to cache the built wheel so we're not rebuilding `python-apt` each time.

.github/workflows/tests.yaml
```diff
@@ -44,6 +45,11 @@ jobs:
           sudo snap install --classic node
           sudo snap install --classic pyright
           make test-pyright
+      - name: Run linkcheck, woke, spelling
+        run: |
+          sudo snap install woke
```
lengau, line 50: If you move the snap installs to just before doing the apt install and add a `--no-wait` and then do `snap watch --last=install` before the first time we use a snap package it'll make the install time negligible.

lengau (approved):
Looks good - I've suggested a couple of small optimisations, but they'll have only a minimal impact

.github/workflows/tests.yaml
cmatsuoka: We can apply those changes uniformly to all projects to avoid having an outlier.

tigarmo (approved):
thanks!

.github/workflows/tests.yaml
cmatsuoka: This can be applied uniformly to all projects if the gain is significant.

PR #688, opened 2024-03-18
feat(docs)!: rework bundling of shared docs

tigarmo:
This PR re-organizes the docs/ folder to share *most* of the rst files. Moving forward, the expectation is that new content will be written inside of docs/common/craft-parts/ if there is any expectation at all that downstream applications will want to include it in their own docs. The idea is to keep to the /common/craft-parts/ folder so that both Craft Parts and applications can refer to the docs using the same path,  without overlaps/conflicts.

Not everything is shared - some docs are kept in Craft Parts as they are not meant to be consumed by users of downstream applications. Some examples:

- Class references to Craft-parts internals;
- Docs about using ``craft-parts`` directly from the command line;
- "Meta documentation" about documenting the library itself;
- Front-page and tables of contents that structure Craft Parts' own docs.

-----

This PR is split over multiple commits to simplify review. The first commit has most of the changes but they are all shuffling files around.


docs/conf.py
```diff
@@ -81,10 +81,17 @@
     "**venv",
     "base",
     "sphinx-resources",
+    # These RST files are explicitly excluded here because they are included by
+    # other files - without this exclusion, Sphinx will complain about duplicate
+    # labels.
+    "common/craft-parts/explanation/overlay_parameters.rst",
```
lengau, line 87: Can we exclude these with `common/**`?

lengau (approved):
Looks good! One idea for potential future proofing, but I'm not sure whether it's reasonable.

docs/conf.py
tigarmo: We could, but I think that has more drawbacks than advantages. An advantage of having to spell out every file is that sphinx will then complain if you add a file in `common/` but forget to reference it in a TOC. This actually made me realize that a bunch of how-tos and explanations had never been linked in [rockcraft's docs](https://github.com/canonical/rockcraft/pull/515/files#diff-c92bad01f08eb61a41f190d07ad34d0829592613d9336e342083237faabdcc91R11), for example.

syu-w (approved):

docs/conf.py
lengau: Yep, agreed. Thanks for the explanation

PR #689, opened 2024-03-18
fix(organize): use the part's install directories

mr-cal:

-----

Fixes an issue discovered via https://github.com/canonical/snapcraft/pull/4654 where `organize_files()` would organize in the `cwd` instead of the project's `base_dir`.

The problem was that `organize_files()` was redefining a part's install directories incorrectly by creating relative filepaths.  

To fix this, `organize_files` now accepts the part's install directories instead of redefining them.


sergiusens (approved):
thank you very much!

PR #721, opened 2024-04-26
fix(part_handler): fetch correct target stage packages

dilyn-corner:

-----

This is a partial attempt at fixing https://github.com/canonical/snapcraft/issues/4542 specifically [Alex's comment](https://github.com/canonical/snapcraft/issues/4542#issuecomment-1934943478).

Note that this change does not take into consideration cases where the stage package is only available for `architecture: all`.

### DIFF
--- a/craft_parts/executor/part_handler.py
+++ b/craft_parts/executor/part_handler.py
@@ -898,7 +898,7 @@ class PartHandler:
             fetched_packages = packages.Repository.fetch_stage_packages(
                 cache_dir=step_info.cache_dir,
                 package_names=stage_packages,
-                arch=step_info.host_arch,
+                arch=step_info.target_arch,
                 base=step_info.base,
                 stage_packages_path=self._part.part_packages_dir,
             )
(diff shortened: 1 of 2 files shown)


lengau (approved):

lengau:
Adding @mr-cal specifically since he probably has context on what else we need to do to fix this before we merge. (See: https://github.com/canonical/craft-parts/pull/258)

lengau:
We'll need to ensure we test for regressions against LP #[1983009](https://bugs.launchpad.net/snapcraft/+bug/1983009) before merging

mr-cal (commented):
As discussed in a call with @dilyn-corner a few weeks ago and in a few chats, I have two concerns here:

1. We need a design that does not change this behavior for existing applications.  For example, we don't want the behavior for core22 snaps to change.
2. We retracted this change when core22 was getting released. @cmatsuoka was working with someone in Canonical who maintains the archives on an issue where `dpkg --add-architecture` would be misinformed by the archives that packages were available for an architecture when they were not.  I don't believe that issue was solved.  We will need to thoroughly test staging packages inside the same architecture families (stage `armhf` in an `arm64` build environment) and across families (e.g. stage a `riscv64` package in an `amd64` build environment).

mr-cal:
Related: https://github.com/canonical/craft-archives/issues/104

cmatsuoka:
The main archive metadata was (still is?) misleading. Some architectures are moved to a separate server after the metadata is already generated, so the metadata lists content that's not actually there.

PR #727, opened 2024-05-15
fix(sequencer): rerun steps only if explicitly requested

cmatsuoka:
Instead of implicitly cleaning and re-executing steps when a part name
is specified, only do it when explicitly requested by the user.
Applications must set the `rerun` argument accordingly when planning
the lifecycle execution.

Fixes canonical/snapcraft#4806

-----


mr-cal (changes requested):
Looks good!

Can you add/extend an integration test in `tests/integration/sequencer/test_sequencer.py`?

cmatsuoka:
Rebased on feature/2.0

lengau (approved):

kubiko:
@cmatsuoka @mr-cal @tigarmo 
Is still suppose to fix when `snapcraft build <part>` does clean step? Because it does not seem to work

PR #737, opened 2024-05-20
feat: use debian architecure names in api

cmatsuoka:
Normalize architecture names across APIs to use Debian names instead
of Python machine names. Debian names are used in build-on/build-for
entries, and by using them in the public API we also avoid the need
to handle unexpected architecture name variations that sometimes appear
when using Python machine names.

-----


tests/integration/lifecycle/test_lifecycle.py
```diff
@@ -369,7 +369,7 @@ def setup(self, new_dir, partitions):
             parts,
             application_name="test_update",
             cache_dir=new_dir,
-            arch="aarch64",
+            arch="amd64",
```
lengau, line 372: This is a behavioural change on the tests (arm64 -> amd64). I'm not sure it matters much, but I figured I'd call it out for examination.

lengau (approved):

tests/integration/lifecycle/test_lifecycle.py
cmatsuoka: Ooo ops this is a typo. Thanks for spotting it.

mr-cal (approved):

PR #740, opened 2024-05-22
sources: add 7z source handler

liushuyu:

-----

Allow parts to specify 7z sources. Based on the implementation from Snapcraft.

Supersedes #232. The changes from #232 include:

- Use better maintained upstream 7zip version instead of defunct `p7zip` package
- Switch linter annotations from Pylint to Ruff
- Use `-o` path option instead of using `cwd=` option in `subprocess.run`


.github/workflows/tests.yaml
```diff
@@ -142,7 +142,7 @@ jobs:
       - name: Install additional test dependencies
         run: |
 
-          sudo apt install -y ninja-build cmake scons qt5-qmake \
+          sudo apt install -y ninja-build cmake scons qt5-qmake p7zip-full \
```
lengau, line 145:
```suggestion
          sudo apt install -y ninja-build cmake scons qt5-qmake p7zip \
```

Let's ensure this works with just a basic 7-zip

lengau (approved):

.github/workflows/tests.yaml
liushuyu: Done

PR #742, opened 2024-05-23
chore: move python version compatibility to 3.10

### LINKED ISSUE #729: Bump minimum Python version to 3.10
### What needs to get done

Bump minimum Python version to 3.10

### Why it needs to get done

Because.

cmatsuoka:

-----

### DIFF
--- a/craft_parts/plugins/rust_plugin.py
+++ b/craft_parts/plugins/rust_plugin.py
@@ -21,7 +21,7 @@ import os
 import re
 import subprocess
 from textwrap import dedent
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, cast
+from typing import TYPE_CHECKING, Any, cast
 
 from overrides import override
 from pydantic import conlist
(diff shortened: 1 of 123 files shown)


cmatsuoka:
Rebased on features/2.0

mr-cal (approved):
Looks good, a few non-blocking thoughts:

1. Can you use `from collections import namedtuple` instead of `from typing import NamedTuple`?
2. There's a leftover `Optional` here: https://github.com/canonical/craft-parts/blob/78bbc6a98dd9ea495487fd580ef7a7eefbaa274b/craft_parts/overlays/layers.py#L45
3. There's a leftover `Union` here: https://github.com/canonical/craft-parts/blob/78bbc6a98dd9ea495487fd580ef7a7eefbaa274b/craft_parts/packages/apt_cache.py#L395

lengau:
Re: @mr-cal 's #1: I don't like the location of `typing.NamedTuple`, but it's better to use when we're using strong typing. Ideally we'd replace most or all of these with dataclasses in a separate PR.

craft_parts/plugins/rust_plugin.py
```diff
@@ -37,7 +37,7 @@
 # see https://github.com/samuelcolvin/pydantic/issues/975#issuecomment-551147305
 # The proper fix requires Python 3.9+ (needs `typing.Annotated`)
 if TYPE_CHECKING:
```
lengau, line 39: GitHub's not letting me apply the suggestion to the full thing, but:

```suggestion
UniqueStrList Annotated[liststr], conlist(str, unique_items=True)]
```

lengau (approved):
LGTM! The one thing I caught doesn't really need changing because it goes away with pydantic 2 anyway.

cmatsuoka:
Removed remaining `Optional` and `Union` instances, typing adjustments to be handled separately.

PR #753, opened 2024-06-13
feat(deb): Don't run apt-get update as non-root user

thp-canonical:

-----

`apt-get update` will fail as non-root user. It is only needed when some build-dependencies are not installed yet. If all build dependencies are installed, craft-parts can be run without root permissions. This was previously only possible by setting the environment variable `CRAFT_PARTS_PACKAGE_REFRESH=0`.


tigarmo:
@thp-canonical is this superseded by #752 ?

thp-canonical:
Yes, indeed :)

PR #763, opened 2024-06-25
partitions feature documentation

mattculler:

-----

Write documentation for application developers who want to use partitions.

(CRAFT-3030)


docs/how-to/partitions.rst
```diff
@@ -0,0 +1,123 @@
...
+
+  organize:
+    (kernel)/usr/local/bin/hello: bin/hello
+
+  Cannot organize files from 'kernel' partition.
+  Files can only be organized from the 'default' partition
+
+When the ``stage`` and ``prime`` keywords are not provided for a part, craft-parts' default behavior is to stage and prime all files for the part in all partitions.
+
+(If a stage or prime filter `is` applied to a partition, the default behavior will not be affected for the other partitions.)
```
mr-cal, line 81: needs double ticks:

```suggestion
(If a stage or prime filter ``is`` applied to a partition, the default behavior will not be affected for the other partitions.)
```

Will you be so kind as to sneak this into your PR?

```diff
diff --git a/.github/workflows/tests.yaml b/.github/workflows/tests.yaml
index 1bff626..31c3fa2 100644
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -42,6 +42,9 @@ jobs:
       - name: Run pydocstyle
         run: |
           make test-pydocstyle
+      - name: Run lint-docs
+        run: |
+          tox run -e lint-docs
       - name: Run pyright
         run: |
           sudo snap install --classic node
```

craft-parts hasn't been rebased onto starbase so this is a bandaid to ensure we don't start snowballing documentation problems from a lack of linting.

It looks like there are 3 small errors in existing documentation that should be very quick to fix.

Ideally you would break all lines at 80 characters, but a ton of docs here are not conforming to that so we can deal with that later.

mattculler: Thanks @mr-cal!  In your suggestion, the single backticks were intentional - I want that "is" to be italicized, not monospaced.

And yep no problem, I'll add the doc linting band-aid and fix the other linting issues.

mr-cal: I think `*` is the right way to do italics or "emphasis" for RST

mattculler: Thanks, fixed

docs/how-to/partitions.rst
```diff
@@ -0,0 +1,123 @@
...
+``default`` must always be the first listed partition.
+
+Partition names and namespace names must consist of *only* lower-case alphabetic characters, unless a partition exists under a namespace, in which case it may also contain hyphen characters, though the first and last characters must still be alphabetic.
+
+.. _app_changes:
+
+Required application changes
+============================
+
+To add partition support to an application, two basic changes are needed to the ``craft-application``.
```
mr-cal, line 19: This feels like a "How to enable partitions through craft-application` rather than a `craft-parts` how to.  I think your how-to from a craft-application perspective is more useful but perhaps this belongs in craft-application?

Ashort reference in craft-parts could cover the specifics like naming conventions, env vars, etc.

docs/how-to/partitions.rst
```diff
@@ -0,0 +1,123 @@
...
+Defined partitions may be referenced in the ``organize``, ``stage``, and ``prime`` sections of your configuration yaml files::
+
+  organize:
+    <source-path>: (<partition>)/<path>
+  stage:
+    - (<partition>)/<path>
+  prime:
+    - (<partition>)/<path>
+
+Paths in the configuration yaml not beginning with a partition label will implicitly use the default partition.
```
mr-cal, line 69: I would move this to reference as it is essentially an API reference.

docs/how-to/partitions.rst
```diff
@@ -0,0 +1,123 @@
...
+  $CRAFT_PRIME                   -> prime
+  $CRAFT_DEFAULT_PRIME           -> prime
+  $CRAFT_KERNEL_PRIME            -> partitions/kernel/prime
+  $CRAFT_COMPONENT_BAR_BAZ_PRIME -> partitions/component/bar-baz/prime
+
+(Note that the hyphen in the partition ``component/bar-baz`` is converted to an underscore in the corresponding variable name.)
+
+You might use these variables in a lifecycle override section of a configuration yaml.  For instance::
+
+  ...
```
mr-cal, line 104: Syntax hightlight with:

```suggestion
.. code-block:: yaml
```

mr-cal (commented):
The documentation itself looks great! I have some opinions on the location and separation of how-to and reference.

docs/how-to/partitions.rst
```diff
@@ -0,0 +1,136 @@
+******************************************
+Add partition support to an application
+******************************************
+
+Partitions basics
+=================
+
+In the below examples, we work with three partitions: ``default``, ``kernel``, and ``bar-baz``.  The ``default`` and ``kernel`` partitions do not have a namespace.  The ``bar-baz`` partition is part of the ``component`` namespace.
```
cmatsuoka, line 8: Perhaps it would be easier to understand if you establish the concept of partitions first, mentioning that they're statically created by the application, and covering namespaced partitions after that (maybe in a separate  section) as a mechanism to do dynamic partitioning.

cmatsuoka, line 8: Maybe calling the partitions `default`, `foo` and `bar` and namespacing `bar` to have `bar-baz` inside could make example names more consistent and easier to follow?

docs/how-to/partitions.rst
```diff
@@ -0,0 +1,136 @@
...
+
+To add partition support to an application, two basic changes are needed to the ``craft-application``.
+
+#. Enable the feature
+
+   When creating the :class:`AppMetadata <craft_application.AppMetadata>`, specify that the application will use the partitions feature:
+
+   .. code-block:: python
+
+     from craft_application import AppMetadata, AppFeatures
```
cmatsuoka, line 27: As @mr-cal mentioned above, craft-parts is not aware of the existence of craft-application. If in craft-parts, this documentation should use only the craft-parts API. Craft-application usage should be part of the craft-application documentation.

docs/how-to/partitions.rst
```diff
@@ -0,0 +1,136 @@
...
+         partitions=["default", "kernel", "component/bar-baz"],
+       )
+
+Using the partitions
+====================
+
+Partitions cannot be used until `after you have configured your application <#app-changes>`_.
+
+In configuration yaml lifecycle
+-------------------------------
```
cmatsuoka, line 62: This section title is a bit confusing, maybe it could be something like "In the parts definition"? Note that craft-parts doesn't have a "project" concept.

docs/how-to/partitions.rst
```diff
@@ -0,0 +1,136 @@
...
+
+Defined partitions may be referenced in the ``organize``, ``stage``, and ``prime`` sections of your configuration yaml files:
```
cmatsuoka, line 64: Maybe explicitly mention that the partition names are enclosed in parentheses?

docs/how-to/partitions.rst
mattculler: Good suggestion!

docs/how-to/partitions.rst
mattculler: Yeah that makes sense, I'll split it up.

docs/how-to/partitions.rst
mattculler: Will update section title in the `craft-application` partitions docs PR.

mr-cal (approved):

docs/common/craft-parts/reference/part_properties.rst
```diff
@@ -90,6 +90,22 @@ to the ``bin`` directory in the staging area and renamed to ``hello``:
    organize:
      hello.py: bin/hello
 
+If partitions are in use by the application, they may be referenced by prepending the partition name, surrounded by parentheses, to the destination path.  Only the destination path may specify a partition; source paths always reference the ``default`` partition.  Omitting the partition name in the destination path causes the file to be copied to the ``default`` partition.
```
lengau, line 93: Stuff that mentions partitions should have a way to ignore them when including the file (e.g. a header so we can have an `end_before`).

docs/common/craft-parts/reference/partition_specific_output_directory_variables.rst
```diff
@@ -0,0 +1,18 @@
+Partition-specific output directory environment variables
+---------------------------------------------------------
+
+If partitions have been enabled, partition-specific environment variables will be created for use during step processing and execution of user-defined scriptlets.
```
lengau, line 4:
```suggestion
If partitions are enabled, partition-specific environment variables will be created for use during step processing and execution of user-defined scriptlets.
```

docs/common/craft-parts/reference/part_properties.rst
mattculler: @lengau I don't understand what you mean.  I'm going to mark this resolved for now, I expect to make another pass on the docs as I implement partitions and we can address this then.

cmatsuoka (approved):

docs/common/craft-parts/reference/part_properties.rst
lengau: The stuff that's in the `common` directory gets `include`d in other projects. Rockcraft for example includes these files but doesn't have partitions enabled, so we should have a way to exclude the info about partitions from the Rockcraft docs.

mattculler: Ahh ok gotcha, thank you!  I'll do this on the next round.

PR #775, opened 2024-07-10
chore!: restructure plugin models

lengau:

-----

This does several related things:

1. Removes the vestigial `PluginModel` from `craft_parts.plugins.base`
2. Combines `PluginPropertiesModel` and `PluginProperties` (which didn't need separation anymore)
3. Makes the `unmarshal` method of `PluginProperties` usable by all plugins (and deletes the unnecessary overrides)

### DIFF
--- a/craft_parts/plugins/nil_plugin.py
+++ b/craft_parts/plugins/nil_plugin.py
@@ -20,12 +20,11 @@ Using this, parts can be defined purely by utilizing properties that are
 automatically included, e.g. stage-packages.
 """
 
-from typing import Any, Literal
+from typing import Literal
 
 from overrides import override
-from typing_extensions import Self
 
-from .base import Plugin, extract_plugin_properties
+from .base import Plugin
 from .properties import PluginProperties
 
 
@@ -33,23 +32,6 @@ class NilPluginProperties(PluginProperties, frozen=True):
     """The part properties used by the nil plugin."""
 
     plugin: Literal["nil"] = "nil"
-
-    @classmethod
-    def unmarshal(cls, data: dict[str, Any]) -> Self:
-        """Populate class attributes from the part specification.
-
-        :param data: A dictionary containing part properties.
-
-        :return: The populated plugin properties data object.
-        """
-        required = ["source"] if "source" in data else []
-        return cls.model_validate(
-            extract_plugin_properties(
-                data,
-                plugin_name="nil",
-                required=cls._required_fields,
-            )
-        )
 
 
 class NilPlugin(Plugin):
--- a/craft_parts/plugins/nil_plugin.py
+++ b/craft_parts/plugins/nil_plugin.py
@@ -20,12 +20,12 @@ Using this, parts can be defined purely by utilizing properties that are
 automatically included, e.g. stage-packages.
 """
 
-from collections.abc import Collection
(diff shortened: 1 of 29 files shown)


craft_parts/plugins/properties.py
```diff
@@ -42,19 +39,38 @@ class PluginProperties(PluginPropertiesModel):
...
     """
 
+    model_config = pydantic.ConfigDict(
+        alias_generator=lambda s: s.replace("_", "-"),
+        extra="forbid",
+        frozen=True,
+        validate_assignment=True,
+    )
+
+    plugin: Any = ""
```
sergiusens, line 49: why Any?

craft_parts/plugins/properties.py
```diff
@@ -65,4 +81,7 @@ def get_pull_properties(cls) -> list[str]:
     def get_build_properties(cls) -> list[str]:
         """Obtain the list of properties affecting the build stage."""
         properties = cls.schema(by_alias=True).get("properties")
-        return list(properties.keys()) if properties else []
+        if properties:
+            del properties["plugin"]
```
sergiusens, line 85: how safe is this?

craft_parts/plugins/properties.py
lengau: It was left over from what @tigarmo suggests when I was playing around with different ways to make creating child nicer and cleaner with pyright. In the end I went with double-declaring `frozen=True` though, so I'll switch this back to `str`

craft_parts/plugins/properties.py
lengau: It should be safe (the `plugin` key not existing in the properties dict would mean we have a much bigger problem), but I rewrite it with a list comprehension anyway since that's faster and safer

craft_parts/plugins/nil_plugin.py
```diff
@@ -20,27 +20,18 @@
...
+from typing import Literal
 
 from overrides import override
 
 from .base import Plugin
 from .properties import PluginProperties
 
 
-class NilPluginProperties(PluginProperties):
+class NilPluginProperties(PluginProperties, frozen=True):
```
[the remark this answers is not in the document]
lengau: Somewhat counter-intuitively, no. The name `_required_fields` came from it being the `required` argument in `extract_plugin_properties`, but in reality it's more like `_extra_fields` or `_fields_that_dont_start_with_the_plugins_name`.

Looking at this though, `extract_plugin_properties` is only actually used here, so I'm thinking about dumping it altogether and moving the logic here anyway.

craft_parts/plugins/properties.py
lengau: (And if you meant because of the `del`, yes that's safe too, as `schema` creates a new dict each time.)

craft_parts/plugins/nil_plugin.py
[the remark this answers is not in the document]
lengau: Ok, I've moved the logic from `extract_plugin_properties` into `PluginProperties.unmarshal()` and used the model's properties to determine the fields to use.

tigarmo (approved):

mr-cal (approved):
Nice refactor!

PR #778, opened 2024-07-12
chore: pydantic 2 related cleanup

lengau:

-----

Requires #775

### DIFF
--- a/craft_parts/constraints.py
+++ b/craft_parts/constraints.py
@@ -15,10 +15,21 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """Pydantic constraints for various items."""
 
+import collections
 import re
-from typing import Annotated
+from typing import Annotated, TypeVar
 
-from pydantic import BeforeValidator, Field
+from pydantic import AfterValidator, BeforeValidator, Field
+
+T = TypeVar("T")
+
+
+def _validate_list_is_unique(value: list) -> list:
+    value_set = set(value)
+    if len(value_set) == len(value):
+        return value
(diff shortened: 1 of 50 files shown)


craft_parts/constraints.py
```diff
@@ -15,10 +15,21 @@
...
-from typing import Annotated
+from typing import Annotated, TypeVar
 
-from pydantic import BeforeValidator, Field
+from pydantic import AfterValidator, BeforeValidator, Field
+
+T = TypeVar("T")
+
+
+def _validate_list_is_unique(value: list) -> list:
```
dariuszd21, line 27:
```suggestion
def _validate_list_is_unique(value: list[T]) -> list[T]:
```

craft_parts/constraints.py
```diff
@@ -43,3 +54,5 @@ def _validate_relative_path_str(path: str) -> str:
     # The field here is used to provide information in the JSON schema and IDEs.
     Field(description="relative path", min_length=1, pattern=re.compile(r"^[^\/].*")),
 ]
+
+UniqueList = Annotated[list[T], AfterValidator(_validate_list_is_unique)]
```
dariuszd21, line 58: It's worth considering if something like `craft_parts/types.py` wouldn't be a better place for this.

dariuszd21 (approved):

craft_parts/constraints.py
lengau: I didn't want to call this file `types` because I wanted to make it obvious that the only types that should belong in here are constrained types for pydantic.

mr-cal (approved):

PR #787, opened 2024-07-18
chore: simplify extract_plugin_properties

lengau:
Ruff 0.5 complains about the for loop

-----

### DIFF
--- a/craft_parts/plugins/base.py
+++ b/craft_parts/plugins/base.py
@@ -18,7 +18,7 @@
 
 import abc
 from copy import deepcopy
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Type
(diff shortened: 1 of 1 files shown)


mr-cal (approved):

craft_parts/plugins/base.py
```diff
@@ -132,7 +132,7 @@ class PluginModel(PluginPropertiesModel):
 
 
 def extract_plugin_properties(
-    data: Dict[str, Any], *, plugin_name: str, required: Optional[List[str]] = None
+    data: Dict[str, Any], *, plugin_name: str, required: Collection[str] = ()
```
mattculler, line 135: Nice!  This is so much better than the old tricks we had to use when we wanted to use a mutable default value.

mattculler (approved):

PR #801, opened 2024-08-06
fix: partitions break when using special characters

mattculler:

-----

Partitions with numbers and hyphens can currently be added, but once you get into lifecycle steps the process will fail strangely - craft-parts thinks paths like `(partition-name1)` are normal paths.

(CRAFT-3031)

### DIFF
--- a/tests/unit/features/partitions/test_parts.py
+++ b/tests/unit/features/partitions/test_parts.py
@@ -118,14 +118,13 @@ class TestPartPartitionUsage:
     @pytest.fixture()
     def partition_list(self):
         """Return a list of partitions, 'default' and 'kernel'."""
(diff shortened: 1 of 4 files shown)


tests/unit/features/partitions/test_parts.py
```diff
@@ -220,8 +224,8 @@ def invalid_fileset(self):
 
         These filepaths are not necessarily violating the naming convention for
         partitions, but the partitions here are unknown and thus invalid to a
```
mr-cal, line 226:
```suggestion
        partitions, but the partitions here are unknown and thus invalid.
```

mr-cal (approved):

tigarmo (approved):

PR #819, opened 2024-08-19
feat(plugins): create base Python plugin

lengau:

-----

This creates a base Python plugin to allow multiple Python plugins with different build systems.

### DIFF
--- a/craft_parts/plugins/base.py
+++ b/craft_parts/plugins/base.py
@@ -19,13 +19,11 @@
 from __future__ import annotations
 
 import abc
-import shutil
 import textwrap
 from copy import deepcopy
 from typing import TYPE_CHECKING
 
 from overrides import override
-from pydantic import errors
 
 from craft_parts.actions import ActionProperties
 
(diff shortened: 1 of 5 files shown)


mr-cal (approved):
Nice!

craft_parts/plugins/base.py
```diff
@@ -122,3 +127,176 @@ def _get_java_post_build_commands(self) -> list[str]:
...
+    @override
+    def get_build_environment(self) -> dict[str, str]:
+        """Return a dictionary with the environment to use in the build step.
+
+        Child classes that need to override this should extend the dictionary returned
+        by this class.
+        """
+        python3_path = shutil.which("python3")
+        if python3_path is None:
+            raise errors.PluginEnvironmentValidationError(
```
tigarmo, line 160: is this check new? isn't it going to conflict with `PARTS_PYTHON_INTERPRETER`?

lengau: Hmm good point. This was pulled out of my initial poetry plugin. I'd like to check for an installed Python at some point, but that's out of scope here so I'm just going to remove this for now.

tigarmo (approved):

PR #826, opened 2024-08-21
feat(partitions): custom error for nonexistent partition

mattculler:

-----

Raise a custom error when a nonexistent partition is asked for, so `corncraft` can catch it more cleanly.

Depending on the stringification is fragile, but if you don't do some sort of deeper check then you could accidentally handle other `ValueError`s.

(CRAFT-3031)

### DIFF
--- a/craft_parts/dirs.py
+++ b/craft_parts/dirs.py
@@ -82,7 +82,7 @@ class ProjectDirs:
         if self._partitions:
             if not partition:
                 raise PartitionUsageError(
-                    error_list=f"Partitions are enabled, you must specify which partition's {dir_name} you want.",
+                    error_list=[f"Partitions are enabled, you must specify which partition's {dir_name!r} you want."],
                     partitions=self._partitions,
                 )
             if partition not in self._partitions:
--- a/craft_parts/dirs.py
+++ b/craft_parts/dirs.py
@@ -20,6 +20,7 @@ from collections.abc import Sequence
 from pathlib import Path
 from types import MappingProxyType
 
+from craft_parts.errors import PartitionNotFound
 from craft_parts.utils import partition_utils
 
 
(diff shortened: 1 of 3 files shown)


craft_parts/dirs.py
```diff
@@ -74,14 +75,25 @@ def __init__(
...
         )
 
+    def _validate_requested_partition(
+        self, dir_name: str, partition: str | None = None
+    ) -> None:
+        """Ensure the requested partition is valid."""
+        if self._partitions:
+            if not partition:
+                raise PartitionUsageError(
+                    error_list=f"Partitions are enabled, you must specify which partition's {dir_name} you want.",
```
lengau, line 85:
```suggestion
                    error_list=[f"Partitions are enabled, you must specify which partition's {dir_name!r} you want."],
```

Thinking it makes more sense to say "...which partition's 'usr' you want" than "...which partitions usr you want."

But also, we probably want that as an iterable containing that one string rather than an iterable containing each character 🙂

craft_parts/errors.py
```diff
@@ -688,3 +693,19 @@ def __init__(self, warning_list: Iterable[str]) -> None:
...
+
+class PartitionNotFound(PartitionUsageError):
+    """A partition has been specified that does not exist.
+
+    :param partition_name: The name of the partition that does not exist.
+    :param partitions: Iterable of the names of valid partitions.
+    """
+
+    def __init__(self, partition_name: str, partitions: Iterable[str]) -> None:
+        self.partition_name = partition_name
```
mr-cal, line 706: Do you need this line?

mr-cal (approved):

craft_parts/dirs.py
mattculler: Oof, good catch!

craft_parts/errors.py
mattculler: Yep I do, it's for callers.  Added a comment clarifying.

lengau (approved):

tigarmo (approved):

PR #830, opened 2024-08-26
feat(docs): add poetry plugin reference

lengau:

-----


docs/common/craft-parts/reference/plugins/poetry_plugin.rst
```diff
@@ -0,0 +1,105 @@
...
+well as those for :ref:`sources <part-properties-sources>`.
+
+Additionally, this plugin provides the plugin-specific keywords defined in the
+following sections.
+
+poetry-with:
+~~~~~~~~~~~~
+**Type:** list of strings
+
+Extra dependency groups to use other than the defaults.
```
medubelko, line 23: On line 30 we provide Poetry's docs about env vars, but here we assume the user is familiar with dependency groups. What do you think about linking to https://python-poetry.org/docs/managing-dependencies#dependency-groups here?

docs/common/craft-parts/reference/plugins/poetry_plugin.rst
```diff
@@ -0,0 +1,105 @@
...
+
+Environment variables
+---------------------
+
+This plugin also sets environment variables in the build environment. User-set
+environment variables will override these values. Users may also set
+`environment variables to configure poetry`_ using the
+:ref:`build-environment <build_environment>` key.
```
medubelko, line 31: > key

I've resisted broaching this because it'll take some time to align all the technical terms, but is there an internally-consistent rule for when to use "key" and "keyword"? It appears like free variation to me.

docs/common/craft-parts/reference/plugins/poetry_plugin.rst
```diff
@@ -0,0 +1,105 @@
...
+
+PARTS_PYTHON_INTERPRETER
+~~~~~~~~~~~~~~~~~~~~~~~~
+**Default value:** python3
+
+The interpreter binary to search for in ``PATH`` or an absolute path to the
```
medubelko, line 37:
```suggestion
Either the interpreter binary to search for in ``PATH``, or an absolute path to the
```

This should help emphasize that these are separate possibilities.

docs/common/craft-parts/reference/plugins/poetry_plugin.rst
```diff
@@ -0,0 +1,105 @@
...
+interpreter (e.g. ``${CRAFT_STAGE}/bin/python``).
+
+PARTS_PYTHON_VENV_ARGS
+~~~~~~~~~~~~~~~~~~~~~~
+**Default value:** (empty string)
+
+Additional arguments for venv.
```
medubelko, line 44: I feel like we could clarify which call to `venv` we mean here. I assume the one Poetry creates?

docs/common/craft-parts/reference/plugins/poetry_plugin.rst
```diff
@@ -0,0 +1,105 @@
...
+
+.. _poetry-details-begin:
+
+Dependencies
+------------
+
+Python
+~~~~~~
+
+By default this plugin uses Python from the base when it is available and
```
medubelko, line 54: > Python from the base

Can we make this "the system-level Python installation"? I feel "base" is too vague.

docs/common/craft-parts/reference/plugins/poetry_plugin.rst
```diff
@@ -0,0 +1,105 @@
...
+**Type:** list of strings
+
+Extra dependency groups to use other than the defaults.
+
+Environment variables
+---------------------
+
+This plugin also sets environment variables in the build environment. User-set
+environment variables will override these values. Users may also set
+`environment variables to configure poetry`_ using the
```
medubelko, line 30:
```suggestion
`environment variables to configure Poetry`_ using the
```

For here and other instances of `poetry`. We have to follow the naming convention laid out by outside projects.

docs/common/craft-parts/reference/plugins/poetry_plugin.rst
```diff
@@ -0,0 +1,105 @@
...
+      after: [poetry-deps]
+
+.. _poetry-details-end:
+
+How it works
+------------
+
+During the build step, the plugin performs the following actions:
+
+1. It creates a virtual environment directly into the ``${CRAFT_PART_INSTALL}``
```
medubelko, line 94:
```suggestion
1. It creates a virtual environment in the ``${CRAFT_PART_INSTALL}``
```

medubelko (changes requested):

docs/common/craft-parts/reference/plugins/poetry_plugin.rst
lengau: I don't see a reason not to link it here.

docs/common/craft-parts/reference/plugins/poetry_plugin.rst
lengau: Agreed except about the comma (these are two objects, not two sentences). Adjusted

docs/common/craft-parts/reference/plugins/poetry_plugin.rst
lengau: I've adjusted this - it should make sense alongside the "How it works" section now.

docs/common/craft-parts/reference/plugins/poetry_plugin.rst
```diff
@@ -0,0 +1,106 @@
...
+Use of ``python3-<python-package>`` in stage-packages will force the inclusion
+of the Python interpreter.
+
+Poetry
+~~~~~~
+
+By default, this plugin gets poetry from the ``python3-poetry`` package on the build
+system. If that is not desired (for example, if a newer version  of poetry is
+required), a ``poetry-deps`` part can install poetry in the build system. Any parts
+that use the poetry plugin must run ``after`` the ``poetry-deps`` part:
```
medubelko, line 68:
```suggestion
By default, this plugin gets Poetry from the ``python3-poetry`` package on the build
system. If that is not desired (for example, if a newer version  of Poetry is
required), a ``poetry-deps`` part can install poetry in the build system. Any parts
that use the Poetry plugin must run ``after`` the ``poetry-deps`` part:
```

Following up with https://github.com/canonical/craft-parts/pull/830#discussion_r1731946482.

medubelko (approved):

docs/common/craft-parts/reference/plugins/poetry_plugin.rst
```diff
@@ -0,0 +1,106 @@
+.. _craft_parts_poetry_plugin:
+
+Poetry plugin
+=============
+
+**Version**: 2.1
```
tigarmo, line 6: I wonder if this is relevant? Since this document is part of craft-parts, if a given application has this reference page then its version of craft-parts must have the poetry plugin right?

Might sound pedantic but I'm thinking about the point of view of a, say, Snapcraft user; there isn't really anything that they can do with this `2.1` since they typically don't know which version of craft-parts their Snapcraft is using

tigarmo (approved):

docs/common/craft-parts/reference/plugins/poetry_plugin.rst
lengau: Agreed and removed. I had it there because of [the plugin documentation howto](https://canonical-craft-parts.readthedocs-hosted.com/en/latest/how-to/document-plugin.html). Do you think we should remove it from there?

tigarmo: yeap

lengau: https://github.com/canonical/craft-parts/pull/831

PR #831, opened 2024-08-28
docs: remove version from plugin documentation howto

lengau:
Put a note that it may be useful to include in app-specific plugins

-----

### DIFF
--- a/docs/how-to/document-plugin.rst
+++ b/docs/how-to/document-plugin.rst
@@ -9,16 +9,12 @@ headings and main text.
 
 <name>
     The name of the plugin; e.g. Python
-<version>
-    The version of the library or tool that provides the plugin; e.g. Snapcraft 8.0.
 
 Replace the instructions with suitable descriptions.
 
 
 <Name> plugin
 -------------
-
-**Version**: <version>
 
 A general description of the plugin that includes as many of these items as
 necessary:
@@ -30,6 +26,9 @@ necessary:
 
 It is also useful to indicate if the plugin replaces an existing plugin, or
 should be used instead of another similar plugin.
+
+If documenting an application-specific plugin, it may be appropriate to include
+the version of the application in which this plugin first appears here.
 
 
 Keywords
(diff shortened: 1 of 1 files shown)


docs/how-to/document-plugin.rst
```diff
@@ -31,6 +27,9 @@ necessary:
 It is also useful to indicate if the plugin replaces an existing plugin, or
 should be used instead of another similar plugin.
 
+If documenting an application-specific plugin, it may be appropriate to include
```
medubelko, line 30: How about:

```suggestion
If you're documenting an application-specific plugin, we encourage you to include
```

lengau: It's not necessarily encouraged though.

medubelko: Oh, I thought this meant that non-application-specific plugins should never have versions, but some application-specific ones should. What are we advising for other plugins?

lengau: For generic craft-parts plugins, they should not have a version. For application-specific ones, it depends on the app. If the app uses versioned documentation, it's not necessary. If the app's documentation is unversioned, it might make sense.

medubelko: What do you think about including all that nuance? I'm not sure I could grok that from this template.

tigarmo (approved):

docs/how-to/document-plugin.rst
lengau: That seems more like something that's up to the app development team.

docs/how-to/document-plugin.rst
```diff
@@ -31,6 +27,9 @@ necessary:
...
+the version of the application in which this plugin first appears here.
+
```
lengau, line 32:
```suggestion
```

docs/how-to/document-plugin.rst
lengau: I've removed the paragraph as a whole.

medubelko (approved):

Issue #844, opened 2024-09-12
npm plugin fails on launchpad


ogra1:
### Bug Description

since several weeks builds with the npm plugin fail in launchpad when an override-build is used and the `craftctl default` step is executed while working fine locally ...

seemingly it fails on this line where it is calling the "npm --version" command:

https://github.com/canonical/craft-parts/blob/main/craft_parts/plugins/npm_plugin.py#L301

since this happens even before the line after the `craftctl default` is executed there must be some internal thing going wrong. please note that this code used to build reliable about two months ago and a simple re-build now reliably fails.

The source as a fork of the customers code is here:

https://github.com/ogra1/everest-demo/tree/main

### To Reproduce

building locally in an lxd container with --destructive mode as well as building by just calling `snapcraft` and leaving the container handling to snapcraft works well ... as soon as the same code from github is attached to an autobuild on launchpad it fails, probably due to a PATH variable setting missing while extracting the node tarball or some such.

### part yaml

```shell
parts:
  nodered:
    plugin: npm
    source: node-red
    npm-include-node: true
    npm-node-version: 22.1.0
    override-build: |
      craftctl default
      npm install
      mkdir -p $CRAFT_PART_INSTALL/lib
      cp -av node_modules $CRAFT_PART_INSTALL/lib/
      ln -s ../lib/node_modules/node-red/red.js $CRAFT_PART_INSTALL/bin/node-red
    prime:
      - -share
```

### Relevant log output
Pull step:

```shell
Pulling nodered
[12/Sep/2024:13:17:02 +0000] "CONNECT nodejs.org:443 HTTP/1.0" 200 16353 "-" "-"
:: + '[' '!' -f /build/everest-demo/parts/nodered/cache/node-v22.1.0-linux-x64.tar.gz ']'
:: + mkdir -p /build/everest-demo/parts/nodered/cache
:: + curl --retry 5 -s https://nodejs.org/dist/v22.1.0/SHASUMS256.txt -o /build/everest-demo/parts/nodered/cache/SHASUMS256.txt
[12/Sep/2024:13:17:02 +0000] "CONNECT nodejs.org:443 HTTP/1.1" 200 10232 "-" "curl/7.81.0"
:: + curl --retry 5 -s https://nodejs.org/dist/v22.1.0/node-v22.1.0-linux-x64.tar.gz -o /build/everest-demo/parts/nodered/cache/node-v22.1.0-linux-x64.tar.gz
[...]

PR #845, opened 2024-09-12
fix(plugins): ensure stateless npm plugin

liushuyu:

-----

This pull request works around an issue that regressed NPM-based projects on the Launchpad auto-build system.

The issue was that the Launchpad auto-build system executed pull and build steps separately, thereby losing the shared state between them.

This is a temporary fix for the 1.33 branch.

### DIFF
--- a/craft_parts/plugins/npm_plugin.py
+++ b/craft_parts/plugins/npm_plugin.py
@@ -280,6 +280,12 @@ class NpmPlugin(Plugin):
     @override
     def get_pull_commands(self) -> List[str]:
         """Return a list of commands to run during the pull step."""
+        return []
+
+    @override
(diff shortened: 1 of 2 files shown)


liushuyu:
I think I can't fix the linter errors in this hotfix due to too many files affected.

tigarmo:
I confirmed that with this fix the repro case linked in #844 builds locally

tigarmo (approved):
thanks for the quick fix ✨

mr-cal (approved):

cmatsuoka (approved):
Thanks for the quick fix. This should work for now, and we can work on a better solution later (maybe adding infrastructure for plugin-defined state files).

PR #864, opened 2024-10-03
feat: support workspaces for go

sergiusens:

-----

### DIFF
--- a/craft_parts/plugins/go_plugin.py
+++ b/craft_parts/plugins/go_plugin.py
@@ -123,20 +123,19 @@ class GoPlugin(Plugin):
 
         # Matches go-use plugin expectation.
         workspace_dir = self._part_info._project_info.dirs.parts_dir  # noqa: SLF001
-        workspace = workspace_dir / "work.go"
+        workspace = workspace_dir / "go.work"
 
-        init_cmds: list[str] = []
         if workspace.exists():
-            init_cmds.append(f"go work use {self._part_info.part_build_dir}")
+            init_cmd = f"go work use {self._part_info.part_build_dir}"
         else:
-            init_cmds.append("go mod download all")
+            init_cmd = "go mod download all"
 
         tags = f"-tags={','.join(options.go_buildtags)}" if options.go_buildtags else ""
 
         generate_cmds: list[str] = [f"go generate {cmd}" for cmd in options.go_generate]
 
         return [
-            *init_cmds,
+            init_cmd,
             *generate_cmds,
             f'go install -p "{self._part_info.parallel_build_count}" {tags} ./...',
         ]
--- a/craft_parts/plugins/go_plugin.py
+++ b/craft_parts/plugins/go_plugin.py
@@ -37,7 +37,6 @@ class GoPluginProperties(PluginProperties, frozen=True):
 
     go_buildtags: list[str] = []
     go_generate: list[str] = []
-    go_workspace_use: list[str] = []
 
     # part properties required by the plugin
     source: str  # pyright: ignore[reportGeneralTypeIssues]
--- a/craft_parts/plugins/go_plugin.py
+++ b/craft_parts/plugins/go_plugin.py
@@ -94,10 +94,6 @@ class GoPlugin(Plugin):
       (list of strings)
       Parameters to pass to `go generate` before building. Each item on the list
       will be a separate `go generate` call. Default is not to call `go generate`.
-    - ``go-workspace-use``
-      (list of strings)
-      Parameters for `go work use` statements to setup a workspace to consume
-      go modules from.
     """
 
     properties_class = GoPluginProperties
@@ -125,13 +121,13 @@ class GoPlugin(Plugin):
         """Return a list of commands to run during the build step."""
         options = cast(GoPluginProperties, self._options)
 
+        # Matches go-use plugin expectation.
+        workspace_dir = self._part_info._project_info.dirs.parts_dir
+        workspace = workspace_dir / "work.go"
+
         init_cmds: list[str] = []
-        if options.go_workspace_use:
-            # go.work should not be part of the source to be built in this
-            # scenario
-            init_cmds.append("go work init")
-            init_cmds.append("go work use .")
-            init_cmds.extend([f"go work use {use}" for use in options.go_workspace_use])
+        if workspace:
+            init_cmds.append(f"go work use {self._part_info.part_build_dir}")
         else:
             init_cmds.append("go mod download all")
 
--- a/craft_parts/plugins/go_plugin.py
+++ b/craft_parts/plugins/go_plugin.py
@@ -122,7 +122,7 @@ class GoPlugin(Plugin):
         options = cast(GoPluginProperties, self._options)
 
         # Matches go-use plugin expectation.
-        workspace_dir = self._part_info._project_info.dirs.parts_dir
+        workspace_dir = self._part_info._project_info.dirs.parts_dir  # noqa: SLF001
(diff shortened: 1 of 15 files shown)


craft_parts/plugins/go_plugin.py
```diff
@@ -120,12 +125,21 @@ def get_build_commands(self) -> list[str]:
         """Return a list of commands to run during the build step."""
         options = cast(GoPluginProperties, self._options)
 
+        init_cmds: list[str] = []
+        if options.go_workspace_use:
+            # Only initialize if go.work is not there.
+            init_cmds.append("[ -f go.work ] || go work init")
```
bowenfan96, line 131: I think it's non-trivial to support pre-existing go.work files, because:

- go will error out if you add the same module twice ("appears multiple times in workspace"). `go workspace use` will not check and silently do the right thing (i.e. not add the duplicate entry), afaict.
- their local go.work paths would likely not exist anyway (unless checked into the sourcecraft source tree -- which is not what we want them to do anyway)

So, we might need to explicitly error out here, at least for a start, because, without more intelligent parsing, we won't know if we're double adding a module.

bowenfan96 (approved):
Thanks very much Sergio! This is just as we discussed and looks great to me with my limited knowledge on craft plugins. I raised a minor point about not supporting existing go.work files inline.

craft_parts/plugins/go_plugin.py
sergiusens: I was in conflict here as I do not want to potentially break existing projects (we are a rolling release after all :-D), but I guess they only break if they opt-in to the feature, for which it won't be a _no change break_, I'll make that amendment.

docs/common/craft-parts/reference/plugins/go_plugin.rst
```diff
@@ -33,6 +33,18 @@ Parameters to pass to `go generate`_ before building. Each item on the list
...
 ``go generate``.
 
+go-workspace-use
+~~~~~~~~~~~~~~~~
+**Type**: list of strings
+**Default** []
+
+Parameters to setup a `go workspace`_ with modules to use for the build. The
+current source is added automatically when any path is defined. Paths must be on
+the local filesystem (e.g.; within ``$CRAFT/STAGE``).
```
tigarmo, line 43:
```suggestion
the local filesystem (for example, within ``$CRAFT_STAGE``).
```

My understanding is that we should avoid the latin abbreviations

docs/common/craft-parts/reference/plugins/go_plugin.rst
```diff
@@ -48,7 +60,7 @@ provision it by itself, to allow flexibility in the choice of compiler version.
 
 Common means of providing ``go`` are:
 
-* The ``golang`` Ubuntu package, declared as a ``build-package``.
+***** The ``golang`` Ubuntu package, declared as a ``build-package``.
```
tigarmo, line 63: what's this change?

tests/integration/plugins/test_go.py
```diff
@@ -114,3 +114,53 @@ def test_go_generate(new_dir, partitions):
...
+    actions = lf.plan(Step.PRIME)
+
+    with lf.action_executor() as ctx:
+        ctx.execute(actions)
+
+    binary = Path(lf.project_info.prime_dir, "bin", "workspace")
+    assert binary.is_file()
+
+    output = subprocess.check_output([str(binary), "--say=hello"], text=True)
+    # This is the expected output that "gen/generator.go" sets in "main.go"
```
tigarmo, line 165: is this comment a copy/paste?

tigarmo (changes requested):

sergiusens:
There is a _how do we clean_ item to deal with, but might need to come later

tigarmo:
image
😿 😭

craft_parts/plugins/go_use_plugin.py
```diff
@@ -0,0 +1,83 @@
...
+
+    @override
+    def get_build_environment(self) -> dict[str, str]:
+        """Return a dictionary with the environment to use in the build step."""
+        return {}
+
+    @override
+    def get_build_commands(self) -> list[str]:
+        """Return a list of commands to run during the build step."""
+        # Set the go workspace directory to live at the root of all parts.
```
tigarmo, line 75: this feels very dirty, on a scale of 1-10 how "production"-ready are we considering this PR?

craft_parts/plugins/go_plugin.py
```diff
@@ -37,6 +37,7 @@ class GoPluginProperties(PluginProperties, frozen=True):
 
     go_buildtags: list[str] = []
     go_generate: list[str] = []
+    go_workspace_use: list[str] = []
```
tigarmo, line 40: is this still used?

tigarmo (commented):
@sergiusens can you talk about the motivation for moving on from the previous implementation to this one?

craft_parts/plugins/go_use_plugin.py
sergiusens: this is what we agreed to with @cmatsuoka

craft_parts/plugins/go_plugin.py
sergiusens: nope

craft_parts/plugins/go_use_plugin.py
```diff
@@ -0,0 +1,83 @@
...
+    """The part properties used by the Go Use plugin."""
+
+    plugin: Literal["go-use"] = "go-use"
+
+    # part properties required by the plugin
+    source: str  # pyright: ignore[reportGeneralTypeIssues]
+
+
+class GoUsePlugin(Plugin):
+    """A plugin to to setup the source into a go workspace.
```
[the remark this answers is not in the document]
sergiusens: fixed

cmatsuoka (approved):
This seems to do what was discussed. Ordinarily I would call this an experimental plugin but in the use context it would be rather inconvenient to have to plaster `--use-experimental-plugins` everywhere. It should be fine if we properly address global state handling early next cycle.

craft_parts/plugins/go_use_plugin.py
```diff
@@ -0,0 +1,104 @@
...
+    @override
+    def get_build_environment(self) -> dict[str, str]:
+        """Return a dictionary with the environment to use in the build step."""
+        return {}
+
+    @override
+    def get_build_commands(self) -> list[str]:
+        """Return a list of commands to run during the build step."""
+        # Set the go workspace directory to live at the root of all parts.
+        workspace_dir = self._part_info._project_info.dirs.parts_dir  # noqa: SLF001
```
tigarmo, line 79:
```suggestion
        workspace_dir = self._part_info.project_info.dirs.parts_dir  # noqa: SLF001
```

https://github.com/canonical/craft-parts/blob/4b67d42c2741d4b558288c5c2036c1d8b0180866/craft_parts/infos.py#L389

docs/common/craft-parts/reference/plugins/go_use_plugin.rst
```diff
@@ -0,0 +1,71 @@
+.. _craft_parts_go_use_plugin:
+
+Go Use plugin
+=============
+
+The Go plugin allows for setting up a `go workspace`_ for `Go`_ modules. It is
```
tigarmo, line 6:
```suggestion
The Go Use plugin allows for setting up a `go workspace`_ for `Go`_ modules. It is
```

craft_parts/plugins/go_use_plugin.py
sergiusens: ty

PR #875, opened 2024-10-09
feat: run apt upgrade in overlay

linostar:

-----
Run `apt-get upgrade` in overlay _once_. This is for both Pro and non-Pro rocks.
The object is to make the overlay base a Pro overlay base by upgrading all packages there.

### DIFF
--- a/craft_parts/executor/executor.py
+++ b/craft_parts/executor/executor.py
@@ -99,6 +99,8 @@ class Executor:
                     self._project_info.overlay_mount_dir, self._project_info
                 )
                 ctx.refresh_packages_list()
+                logger.info("Upgrading packages in the overlay base")
+                ctx.upgrade_packages()
 
         callbacks.run_prologue(self._project_info)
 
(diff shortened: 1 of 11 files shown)


linostar:
@canonical/starcraft-reviewers

craft_parts/executor/executor.py
```diff
@@ -99,6 +99,8 @@ def prologue(self) -> None:
                     self._project_info.overlay_mount_dir, self._project_info
                 )
                 ctx.refresh_packages_list()
+                logger.info("Upgrading packages in the overlay base")
+                ctx.upgrade_packages()
```
tigarmo, line 103: @linostar have you tested this with Rockcraft? I'm don't remember the details but I'm almost sure that the contents of this `PackageCacheMount` layer do not end up in the final lifecycle payload. Typically we update the packages list and then mount [a layer on top of this one](https://github.com/canonical/craft-parts/blob/main/craft_parts/executor/part_handler.py#L240) to install overlay-packages and execute the overlay-script.

sergiusens:
I am closing this as we decided to go down another path.

PR #887, opened 2024-10-26
feat: app-specific project dirs

mattculler:

-----

Also considered calling these "app dirs" rather than "user dirs".

(CRAFT-3120)


mattculler:
After discussion with the team, this is not the right implementation for what I'm trying to do.  Will use ProjectInfo instead.

PR #902, opened 2024-11-15
build: migrate to pyproject.toml

lengau:

-----

This removes setup.py and setup.cfg in favour of the newer pyproject.toml file.

It is a necessary portion of our move to starflow workflows, but standalone and per-project.


pyproject.toml
```diff
@@ -1,3 +1,177 @@
...
+
+[build-system]
+requires = [
+    "setuptools>=69.0",
+    "setuptools_scm[toml]>=7.1"
+]
+build-backend = "setuptools.build_meta"
+
+[tool.setuptools.package-dir]
+"craft_parts_docs" = "docs/common"
```
sergiusens, line 152: when adding this it seems it disables automatic package discovery, which is the root cause to the import errors for craft parts itself

```suggestion
[tool.setuptools.package-dir]
"craft_parts" = "craft_parts"
"craft_parts_docs" = "docs/common"
```

pyproject.toml
```diff
@@ -1,3 +1,178 @@
...
+[tool.setuptools.package-dir]
+"craft_parts" = "craft_parts"
+"craft_parts_docs" = "docs/common"
+
+
+[tool.setuptools.dynamic]
+readme = {file = "README.rst"}
+
+[tool.setuptools_scm]
+write_to = "craft_parts/_version.py"
```
tigarmo, line 160: Going to need to update [this](https://github.com/canonical/craft-parts/blob/main/craft_parts/__init__.py#L19)
and remove .bumpversion.cfg

tigarmo (changes requested):

sergiusens (approved):
considering Tiago's request I consider this good to go

lengau:
Rebased on main, also added bumpversion change.

pyproject.toml
lengau: Done! https://github.com/canonical/craft-parts/pull/902/commits/070b8341765435aeb0ceca461b390f4a9e172291

lengau:
Linters and tests are succeeding locally for me - will move out of draft if they succeed on CI.

mr-cal (approved):

tigarmo (approved):
thanks
I think it's a good idea to remove the bumpversion.cfg file, as now it's broken and somebody might end up using it by accident

lengau:
@tigarmo I agree - I'd just missed it in my work, so thanks for catching it!

PR #904, opened 2024-11-20
feat: add chisel as a build snap when stage-packages contain slices

linostar:

-----

This PR aims to stop chisel from being unnecessarily installed in the Rockcraft snap.

It does that by having chisel as a build snap only when the following conditions are met:
- there is at least one `part` that has slices in its `stage-packages`
- there is no `part` that has chisel in its `build-snaps`

### DIFF
--- a/craft_parts/parts.py
+++ b/craft_parts/parts.py
@@ -178,11 +178,15 @@ class PartSpec(BaseModel):
     @property
     def has_slices(self) -> bool:
         """Return whether the part contains chisel slices."""
-        return any(p for p in self.stage_packages if "_" in p)
+        if not self.stage_packages:
+            return False
+        return any("_" in p for p in self.stage_packages)
 
     @property
     def has_chisel_as_build_snap(self) -> bool:
         """Return whether the part has chisel as build snap."""
+        if not self.build_snaps:
+            return False
         return any(
             p for p in self.build_snaps if p == "chisel" or p.startswith("chisel/")
         )
--- a/craft_parts/parts.py
+++ b/craft_parts/parts.py
@@ -175,6 +175,18 @@ class PartSpec(BaseModel):
             or self.overlay_files != ["*"]
         )
 
+    @property
+    def has_slices(self) -> bool:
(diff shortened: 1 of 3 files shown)


craft_parts/parts.py
```diff
@@ -175,6 +175,18 @@ def has_overlay(self) -> bool:
...
 
+    @property
+    def has_slices(self) -> bool:
+        """Return whether the part contains chisel slices."""
+        return any(p for p in self.stage_packages if "_" in p)
+
+    @property
+    def has_chisel_as_build_snap(self) -> bool:
+        """Return whether the part has chisel as build snap."""
+        return any(
```
SamirPS, line 186: can't we just do 

` p for p in self.build_snaps if p.startswith("chisel")`

linostar: No, because if there was a snap called "chiselhelper" for instance, it will match with that, which we don't want.

SamirPS: yeah true

linostar:
@canonical/starcraft-reviewers This is ready for review.

tests/unit/test_parts.py
```diff
@@ -335,6 +335,31 @@ def test_part_has_overlay(self, partitions):
...
+        p = Part("foo", tc_spec, partitions=partitions)
+        assert p.spec.has_slices == tc_result
+
+    @pytest.mark.parametrize(
+        ("tc_spec", "tc_result"),
+        [
+            ({}, False),
+            ({"build-snaps": ["hello", "chisel/latest/candidate"]}, True),
+            ({"build-snaps": ["chisel"]}, True),
+            ({"build-snaps": ["hello"]}, False),
```
mr-cal, line 356: You had a good point in your comment, I think you should test it.

```suggestion
            ({"build-snaps": ["hello"]}, False),
            ({"build-snaps": ["chiselhelper"]}, False),
```

mr-cal (approved):

tests/unit/test_parts.py
linostar: Added.

PR #910, opened 2024-11-22
feat: add field to indicate user-caused errors

bepri:

-----

Needed by https://github.com/canonical/craft-application/pull/559

### DIFF
--- a/tests/unit/test_errors.py
+++ b/tests/unit/test_errors.py
@@ -27,7 +27,7 @@ def test_parts_error_brief():
     assert str(err) == "A brief description."
     assert (
         repr(err)
-        == "PartsError(brief='A brief description.', details=None, resolution=None, doc_slug=None)"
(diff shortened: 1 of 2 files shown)


tests/unit/test_errors.py
```diff
@@ -27,7 +27,7 @@ def test_parts_error_brief():
     assert str(err) == "A brief description."
     assert (
         repr(err)
-        == "PartsError(brief='A brief description.', details=None, resolution=None, doc_slug=None)"
+        == "PartsError(brief='A brief description.', details=None, resolution=None, doc_slug=None, user_error=False)"
```
dariuszd21, line 30: You should also an assertion for `user_error` field as it is for others

dariuszd21 (approved):
Thank you!

Do we want it to be set to `True` for the specific errors by default ?

Consider also parametrizing some tests around e.g. `PluginBuildError` to see if value is correctly propagated

PR #925, opened 2024-12-04
feat: use "craft.git" for git where possible

### LINKED ISSUE #900: use `craft.git` for git operation instead of relying on build-package
### What needs to get done

use `craftgit`, where git client is run as a subprocess with the fallback to the `git` to ensure backward compatibility. 

### Why it needs to get done

our snap package will bundle with `craftgit` which is git client packaged specifically for the craft libraries use.
we should use it in the places where we really on the system git and fallback to the git if it's not a snap or the binary is not available.

tigarmo:
If "craft.git" is available on the PATH, use it. Otherwise, fallback to "git".

Fixes #900

-----


craft_parts/utils/git.py
```diff
@@ -0,0 +1,32 @@
...
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Helpers to use git."""
+import shutil
+from functools import cache
+
+
+@cache
+def get_git_command() -> str:
```
tigarmo, line 23: since we'll soon support custom source handlers, I added this as a free function so that git-based handlers can use it too

craft_parts/sources/git_source.py
```diff
@@ -80,13 +82,19 @@ class GitSource(SourceHandler):
     """
 
     source_model = GitSourceModel
-    command: ClassVar[str] = "git"
```
lengau, line 83: I don't expect that anything would break with this, but it's technically an API change.

Would it be less of a change to make the method below a property instead?

craft_parts/utils/git.py
```diff
@@ -0,0 +1,32 @@
...
+    """Get the git command to use.
+
+    This function will prefer the "craft.git" binary if available on the system,
+    returning its full path. Otherwise, it will always return "git", without
+    checking for availability.
```
lengau, line 28: If I understand correctly, this means that `craft.git` only has to be on the `PATH` for the context where craft-parts runs, and `git` only has to be on the `PATH` for the context where the commands run?

I'm not sure whether we manipulate `PATH` like we do with other environment variables for the execution context, but I very much like this.

lengau (approved):

craft_parts/sources/git_source.py
tigarmo: I did that initially but alas, classmethod properties are removed in Python 3.13 and some classmethods access this

craft_parts/utils/git.py
tigarmo: can you clarify what you mean? I'm not sure I understand

craft_parts/sources/git_source.py
```diff
@@ -80,13 +82,19 @@ class GitSource(SourceHandler):
...
+
+    @staticmethod
+    def command() -> str:
```
sergiusens, line 87: if you are changing the signature for this, can we not use the new method this wraps directly instead?

craft_parts/sources/git_source.py
sergiusens: this would have been ideal :-)

craft_parts/sources/git_source.py
tigarmo: sure
done

sergiusens (approved):

PR #927, opened 2024-12-05
feat: add an API for registering source types

### LINKED ISSUE #915: add API for registering source types
### What needs to get done

Add an API that allows to register new source types.

### Why it needs to get done

We want to be able to register additional source types from other `*crafts`, similarly to how the plugins are handled

lengau:
This provides a pair of functions (sources.register, sources.unregister) for registering and unregistering source types. This differs slightly from plugin registration:

- All current sources are considered mandatory, meaning they cannot be replaced or unregistered by the application.
- Mandatory sources can be changed at a major release.

-----

Fixes #915, CRAFT-3734

### DIFF
--- a/craft_parts/sources/sources.py
+++ b/craft_parts/sources/sources.py
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright 2015-2021 Canonical Ltd.
+# Copyright 2015-2021,2024 Canonical Ltd.
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
@@ -27,7 +27,7 @@ code for that part, and how to unpack it if necessary.
     directory tree or a tarball or a revision control repository
     ('git:...').
 
-  - source-type: git, bzr, hg, svn, tar, deb, rpm, or zip
+  - source-type: git, tar, deb, rpm, or zip
 
     In some cases the source string is not enough to identify the version
     control system or compression algorithm. The source-type key can tell
@@ -83,10 +83,12 @@ import re
 from pathlib import Path
 from typing import TYPE_CHECKING
 
+import pydantic_core
+
 from craft_parts.dirs import ProjectDirs
 
 from . import errors
-from .base import SourceHandler
+from .base import BaseSourceModel, SourceHandler
 from .deb_source import DebSource
 from .file_source import FileSource
 from .git_source import GitSource
@@ -102,8 +104,7 @@ if TYPE_CHECKING:
 
 SourceHandlerType = type[SourceHandler]
 
-
-_source_handler: dict[str, SourceHandlerType] = {
+_MANDATORY_SOURCES: dict[str, SourceHandlerType] = {
     "local": LocalSource,
     "tar": TarSource,
     "git": GitSource,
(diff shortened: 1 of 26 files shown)


tests/unit/sources/test_sources.py
```diff
@@ -45,24 +93,55 @@ def test_get_source_handler_class_with_invalid_type():
...
         (".tgz", "tar"),
         (".tar", "tar"),
         ("git:", "git"),
         ("git@", "git"),
         ("git+ssh:", "git"),
+        ("git+https:", "git"),
         (".git", "git"),
-        ("lp:", "bzr"),
-        ("bzr:", "bzr"),
-        ("svn:", "subversion"),
```
lengau, line 58: These were leftovers in `get_source_type_from_uri` - craft-parts doesn't have these source handlers. While this is technically a backwards-incompatible change for `get_source_type_from_uri`, all practical use cases are that this would have errored when trying to find a source class anyway.

docs/how-to/source-handler.rst
lengau, line 1: @medubelko if you'd like I can move this to a separate PR. We didn't have a documentation task for this and I decided it was worth documenting so I just included it here.

docs/how-to/source-handler.rst
```diff
@@ -0,0 +1,86 @@
...
+Once this is defined, a :py:class:`~craft_parts.sources.base.SourceHandler` is
+needed to define the ``PULL`` behaviour of the source type.
+
+.. literalinclude:: source-handler/rsync_source.py
+    :language: python
+    :pyobject: RsyncSource
+
+.. note::
+    Craft Parts does not install any required tools for custom source handlers.
+    The given handler will fail on a machine that does not have
```
medubelko, line 53:
```suggestion
    The handler in this example will fail on a machine that does not have
```

medubelko (approved):
The docs look quite good! I have one minor suggestion but it's not blocking.

tigarmo (approved):
thanks, very nice. Please add a changelog entry but otherwise LGTM

docs/how-to/source-handler.rst
```diff
@@ -0,0 +1,86 @@
...
+needed to define the ``PULL`` behaviour of the source type.
+
+.. literalinclude:: source-handler/rsync_source.py
+    :language: python
+    :pyobject: RsyncSource
+
+.. note::
+    Craft Parts does not install any required tools for custom source handlers.
+    The handler in this example will fail on a machine that does not have
+    `rsync <https://rsync.samba.org/>`_ installed before the part is pulled.
```
dariuszd21, line 54: Is it by design ?
It's a bit of limitation, that you cannot specify additional package to handle the source type

dariuszd21 (approved):
It looks good! Thank you!

The only thing that's important to mention is that new source types cannot define theirs additional  packages and/or fields as they have to follow `PartSpec` definition (but that was always the limitation of `craft-parts`)

craft_parts/sources/sources.py
```diff
@@ -115,6 +116,39 @@
...
+def register(source: SourceHandlerType, /) -> None:
+    """Register source handlers.
+
+    :param source: a SourceHandler class to register.
+    :raises: ValueError if the source handler overrides a built-in source type.
+    """
+    source_name = _get_type_name_from_model(source.source_model)
+    if source_name in _MANDATORY_SOURCES:
+        raise ValueError(f"Built-in source types cannot be overridden: {source_name!r}")
+    _SOURCES[source_name] = source
```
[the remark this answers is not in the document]
lengau: We don't for plugins, so I'm not sure we should here.

docs/how-to/source-handler.rst
lengau: More about not wanting to change too many unrelated parts. Currently sources don't define the packages they need like plugins do, so we'd have to change the design to include that. Since we don't currently have a use case for additional dependencies, the extra work to do that was something I decided against.

When we come up with a use case for it, we can modify sources to include pull packages and snaps like we do with plugins.

mr-cal (approved):

PR #928, opened 2024-12-06
doc: create new how-to on creating a plugin

bepri:

-----

(CRAFT-3762)


docs/how-to/create_plugin.rst
tigarmo, line 1: https://canonical-craft-parts--928.com.readthedocs.build/en/928/how-to/create_plugin.html

docs/how-to/create_plugin.rst
```diff
@@ -0,0 +1,100 @@
+.. _how_to_create_plugin:
+
+How to create a plugin
+======================
+
+A new tool or programming language may be easier to use craft-parts with the help of a new plugin. This document contains instructions on how to recreate the simple :py:class:`~craft-parts.plugins.dump_plugin.DumpPlugin` plugin.
```
tigarmo, line 6: this first sentence reads awkwardly to me

docs/how-to/create_plugin.rst
tigarmo, line 1: @bepri please break those big lines into something more manageable

docs/how-to/create_plugin.rst
```diff
@@ -0,0 +1,100 @@
...
+Creating the plugin class
+~~~~~~~~~~~~~~~~~~~~~~~~~
+The plugin class itself contains the functionality of the plugin. Start by declaring the properties class:
+
+.. code:: py
+
+    class NewDumpPlugin(Plugin):
+        """Copy the content from the part source."""
+        
+        properties_class = DumpPluginProperties
```
tigarmo, line 49:
```suggestion
        properties_class = NewDumpPluginProperties
```

docs/how-to/create_plugin.rst
bepri: Jeez, not sure how I missed this in my first proofread. Thanks!

docs/how-to/create_plugin.rst
```diff
@@ -0,0 +1,131 @@
+.. _how_to_create_plugin:
+
+How to create a plugin
+======================
+
+Plugins help to bring new tools and programming languages into craft-parts.
```
medubelko, line 6: > craft-parts

I think we're using "Craft Parts" for the name, now.

docs/how-to/create_plugin.rst
```diff
@@ -0,0 +1,131 @@
...
+First, we'll have to pull in our imports. Since this is such a simple plugin,
+the list of necessary imports is quite small:
+
+.. code:: py
+
+    from typing import Literal
+    from overrides import override
+
+    from .base import Plugin
+    from .properties import PluginProperties
```
medubelko, line 38: I'm not sure how useful a section on imports is. Python coders know there should be imports, and scripts are open-ended. I don't see much in the way of a decision the reader needs to make here. Is there anything specific to plugins that the reader ought to do as far as imports? If not, we could probably cut this.

docs/how-to/create_plugin.rst
```diff
@@ -0,0 +1,131 @@
...
+
+Defining a properties class
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+All plugins make use of a "properties" class to describe their unique parameters.
+The dump class is very basic, so we only need to define the required parameters:
```
medubelko, line 43: As a reader, I'd like to know:

- Are there _non_ unique arguments?
- Where should I go if I want non-basic arguments?

Are these relevant questions we could answer?

docs/how-to/create_plugin.rst
```diff
@@ -0,0 +1,131 @@
...
+
+.. code:: py
+
+    class NewDumpPluginProperties(PluginProperties, frozen=True):
+        """The part properties used by the new dump plugin."""
+
+        # The human-friendly name of the plugin
+        plugin: Literal["new-dump"] = "new-dump"
+        # The "source" parameter is special - we'll talk about it more soon.
```
medubelko, line 52: I'd put this in text rather than the code. There's a high chance the reader will miss it.

docs/how-to/create_plugin.rst
```diff
@@ -0,0 +1,131 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+All plugins make use of a "properties" class to describe their unique parameters.
+The dump class is very basic, so we only need to define the required parameters:
+
+.. code:: py
+
+    class NewDumpPluginProperties(PluginProperties, frozen=True):
+        """The part properties used by the new dump plugin."""
+
+        # The human-friendly name of the plugin
```
medubelko, line 50: More of a technical question: is `new-dump` human-friendly? Kebab case is still a form of code, in my opinion. We could instead say it's friendlier or a convenient identifier.

docs/how-to/create_plugin.rst
```diff
@@ -0,0 +1,131 @@
...
+
+.. code:: py
+
+    class NewDumpPlugin(Plugin):
+        """Copy the content from the part source."""
+        
+        properties_class = NewDumpPluginProperties
+    
+Next, we'll have to start defining the special methods that make this plugin
+"work". Since this is such a simple plugin, many of these methods are empty:
```
medubelko, line 68: It's best not to use air quotes. It can sound insincere.

docs/how-to/create_plugin.rst
```diff
@@ -0,0 +1,131 @@
...
+        return set()
+
+    @override
+    def get_build_environment(self) -> dict[str, str]:
+        """Return a dictionary with the environment to use in the build
+        step."""
+        return {}
+
+The last method that we have to define, however, is where the actual build
+commands are defined. These are the exact commands executed by a subprocess
```
medubelko, line 97: I think it would be a good idea to link the words "build commands" to the page about the parts lifecycle.

docs/how-to/create_plugin.rst
```diff
@@ -0,0 +1,131 @@
...
+
+    from typing import Literal
+    from overrides import override
+
+    from .base import Plugin
+    from .properties import PluginProperties
+
+Defining a properties class
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+All plugins make use of a "properties" class to describe their unique parameters.
```
medubelko, line 42: > unique parameters

I understand the class name is well-entrenched, but for simplicity we might be better off using "unique arguments", since "parameter" is a heavily-overloaded term across computing and is a structure in Python classes, which we discuss in this doc.

docs/how-to/create_plugin.rst
```diff
@@ -0,0 +1,131 @@
...
+        """Return a list of commands to run during the build step."""
+        install_dir = self._part_info.part_install_dir
+        return [f'cp --archive --link --no-dereference . "{install_dir}"']
+
+Notice we don't require a user-provided source directory/file - this is because
+we let the ``source`` parameter do the heavy-lifting. The source parameter is a
+mandatory parameter that specifies the working files of a part. It has a lot of
+specialised logic for different file types ranging from tar files to git
+repositories, so we can leverage that to get much stronger functionality out of
+our otherwise simple plugin.
```
medubelko, line 113: Can we move everything after "heavy-lifting" into the earlier section about properties/arguments?

docs/how-to/create_plugin.rst
```diff
@@ -0,0 +1,131 @@
...
+
```
medubelko, line 114: Can we provide a few more details about other bits the developer might want to take action on, like the first steps with `get_build_snaps`? My concern is that with all this placeholder code there's not much for our audience to go on.

docs/how-to/create_plugin.rst
```diff
@@ -0,0 +1,131 @@
...
+Final steps
+~~~~~~~~~~~
+Now that we have our very own plugin, the last step is to make the lifecycle
+manager aware of this plugin. Since we created our plugin directly in
+``craft-parts``, all that's needed is to add it to a dictionary in
+:file:`craft_parts/plugins/plugins.py`:
```
medubelko, line 120: Aside from the "we", this is good!

docs/how-to/create_plugin.rst
```diff
@@ -0,0 +1,131 @@
...
+.. code:: py
+
+    from .new_dump_plugin import NewDumpPlugin
+
+    # ...
+
+    _BUILTIN_PLUGINS: dict[str, PluginType] = {
+        # ...
+        "new-dump": NewDumpPlugin,
+    }
```
medubelko, line 131: I know the other how-tos in Craft Parts are still nascent, but have you considered placing some links to further tasks or references in a _Next steps_ section?

medubelko (changes requested):
Thanks for putting this to the page so quickly @bepri.

I've left some very high-level comments and suggestions about goals, structure, and tone in how-tos. This draft a great first step and I think with some adjustment we can align it with the aims of Diataxis.

docs/how-to/create_plugin.rst
bepri: The need to import `Literal` and `override` should be pretty easy to guess if we just started using it in code without mention of it, but first-time developers could get confused on `Plugin` and `PluginProperties` since they aren't from any standard library, but rather declared locally. I could move each of the imports into their relevant code blocks, though. What do you think?

docs/how-to/create_plugin.rst
bepri: - I think I'm conflating "non-unique" and "required" in a weird way here. I'll reword it. What I'm trying to say is that there are mandatory properties (such as the name of the plugin), but you can also add your own special properties (such as compiler flags or dependency files). I'll adjust.
- Answering this might go beyond scope of this how-to since we aren't using any "special properties" (as defined above), but I can add a couple of examples there or slightly modify the dump plugin to have some sort of "special property" if you think that would be useful to demonstrate. Let me know!

docs/how-to/create_plugin.rst
bepri: I left the doc comments in those functions to hopefully cover that problem. I'm having some difficulty balancing "only explaining what we need" and "covering all the bases" without creating too complex of an example. I can give a small definitions list blurb of the functions, perhaps? Taking suggestions here!

docs/how-to/create_plugin.rst
medubelko: Groovy. Let's tell the reader to import `Plugin` and `PluginProperties`. You can drop a quick line about those being local and necessary for the rest of the class.

I don't think it's necessary to itemise the code blocks, though.

docs/how-to/create_plugin.rst
medubelko: The added description clarifies thing. I would call what we're describing the plugin's main interface.

I think we should dial back the amount of description in your latest commit, and compose this section like this:

> Next, define the plugin's interface. All plugins make use of a `properties` class to describe their interfaces.
There are two recommended properties that all plugins have in common, and you can add more to account for additional behaviour.
> 
> [sample code]
> 1. Add the two recommended properties:
>     -  ``plugin`` as a Pydantic literal. This property represents the exposed name of your plugin. It will be invoked in user-facing tools and configurations, such as keys that select your plugin in a Snapcraft recipe.
>     - ``source`` as a string. This property is the path to the working files of a part. It has specialised logic for different file types ranging from tarballs to Git repositories, so you can leverage it to get much broader functionality out of an otherwise simple plugin.
> 2. Add any custom properties your plugin needs. For example, a plugin based on Java might have a unique property to specify the Java version, or a Python-based plugin might have one to specify a requirements file.
> 
> The  ``new-dump`` plugin will be quite simple, so it doesn't need custom properties. 
```

File: docs/how-to/create_plugin.rst (hunk 10 above)
Comment (medubelko) [reply docs/how-to/create_plugin.rst]: Ignore this, I misunderstood the description.

File: docs/how-to/create_plugin.rst (hunk 12 above)
Comment (medubelko) [reply docs/how-to/create_plugin.rst:120]: There are a few things we could do:

1. Chunk this section into steps. Each new paragraph could be its own step, so I count three. This advice applies to all these sections, BTW.
2. After "many of these methods are empty", we could _direct_ the reader to "Fill in the behaviour of these methods", for example.
3. Expanding on # 2, we could then bring back your earlier Java example, and suggest that for that hypothetical plugin,  `get_build_packages` would need to be extended to shepherd in the installation of Java runtimes and dependencies for the plugin.

If it helps, here's a [recent example](https://canonical-snapcraft.readthedocs-hosted.com/en/latest/howto/craft-for-platforms/example-gtk4-app/) of how I described writing code (well, configurations) as steps.

**Review by medubelko — CHANGES_REQUESTED** [review]
@bepri I left follow-up points in the comment chains.

The major overriding change I think we should try for is to turn as much of this as possible into concrete steps. Steps are the backbone of how-tos. This can seem daunting for code, but it's possible with the right patterns. I linked this [recent page from Snapcraft](https://canonical-snapcraft.readthedocs-hosted.com/en/latest/howto/craft-for-platforms/example-gtk4-app/) in another comment. It should offer a reasonably effective baseline for how to accomplish code-as-steps. Let me know if you'd like more guidance on this front.

**Review by medubelko — COMMENTED** [review]
@bepri I like the updated structure!

There's some RST-related errata that we ought to adjust, unrelated to content. Would you like me to commit them directly, or leave more comments?

**Comment by bepri** [comment]
Feel free to add them yourself, just let me know when you're done so I can learn from what you decide to fix!

File: docs/how-to/create_plugin.rst [hunk 15]
```diff
@@ -0,0 +1,157 @@
+.. _how_to_create_plugin:
+
+How to create a plugin
+======================
+
+Plugins help to bring new tools and programming languages into Craft Parts.
+This document contains instructions on how to recreate the simple
+:py:class:`~craft-parts.plugins.dump_plugin.DumpPlugin` plugin.
```
mr-cal, line 8: Is this supposed to link to something? It just renders as monospaced text for me.

mr-cal (approved):
Looks excellent to me, nice work!

docs/how-to/create_plugin.rst
bepri: No, that's just how I saw other documentation format a python class name. I can make it link to the dump plugin's documentation if you think that's better, though!

tigarmo: > I can make it link to the dump plugin's documentation if you think that's better, though!

yes please

medubelko (approved):

PR #930, opened 2024-12-09
feat: add uv plugin

bepri:

-----

### DIFF
--- a/craft_parts/plugins/uv_plugin.py
+++ b/craft_parts/plugins/uv_plugin.py
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright 2024 Canonical Ltd.
+# Copyright 2020-2024 Canonical Ltd.
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
@@ -16,85 +16,137 @@
 
 """The uv plugin."""
 
+import pathlib
+import shlex
+import subprocess
+from typing import Literal
+
+import pydantic
+from overrides import override
+
+from craft_parts.plugins import validator
+from craft_parts import errors
+
 from .base import BasePythonPlugin
 from .properties import PluginProperties
-from typing import Literal, Annotated
-from pathlib import Path
-from overrides import override
-from .validator import PluginEnvironmentValidator
-import subprocess
-from craft_parts.errors import PluginEnvironmentValidationError
-import pydantic
-import shlex
+
 
 class UvPluginProperties(PluginProperties, frozen=True):
+    """The part properties used by the uv plugin."""
+
     plugin: Literal["uv"] = "uv"
-    uv_extra: set[str] = pydantic.Field(
+
+    uv_extras: set[str] = pydantic.Field(
         default_factory=set,
         title="Optional dependency groups",
         description="Optional dependency groups to include when installing.",
     )
 
+    # part properties required by the plugin
     source: str  # pyright: ignore[reportGeneralTypeIssues]
 
-class UvPluginEnvironmentValidator(PluginEnvironmentValidator):
+
+class UvPluginEnvironmentValidator(validator.PluginEnvironmentValidator):
+    """Check the execution environment for the uv plugin.
+
+    :param part_name: The part whose build environment is being validated.
+    :param env: A string containing the build step environment setup.
+    """
+
     _options: UvPluginProperties
 
     @override
-    def validate_environment(self, *, part_dependencies: list[str] | None = None) -> None:
+    def validate_environment(
+        self, *, part_dependencies: list[str] | None = None
+    ) -> None:
+        """Ensure the environment has the dependencies to build uv applications.
+
+        :param part_dependencies: A list of the parts this part depends on.
+        """
         version = self.validate_dependency(
             dependency="uv",
             plugin_name=self._options.plugin,
             part_dependencies=part_dependencies,
-            argument="--version",
+            argument="version",
         )
         if not version.startswith("uv") and (
-            part_dependencies is None or "go-deps" not in part_dependencies
+            part_dependencies is None or "uv-deps" not in part_dependencies
         ):
-            raise PluginEnvironmentValidationError(
+            raise errors.PluginEnvironmentValidationError(
                 part_name=self._part_name,
                 reason=f"invalid uv version {version!r}",
             )
 
+
 class UvPlugin(BasePythonPlugin):
+    """A plugin to build python parts."""
+
     properties_class = UvPluginProperties
     validator_class = UvPluginEnvironmentValidator
     _options: UvPluginProperties
     
-    def _system_has_uv(self) -> bool:
-        try:
-            uv_version = subprocess.check_output(["uv", "--version"], text=True)
-        except (subprocess.CalledProcessError, FileNotFoundError):
-            return False
-        return "uv" in uv_version
+    def _get_uv(self) -> str:
+        return f'"${{HOME}}/.cargo/bin/uv'
     
-    def _get_uv_export_commands(self, requirements_path: Path) -> list[str]:
+    @override
+    def _get_pip(self) -> str:
+        return f"{self._get_uv()} pip"
+
+    def _get_uv_export_commands(self, requirements_path: pathlib.Path) -> list[str]:
+        """Get the commands for exporting from uv.
+
+        Application-specific classes may override this if they need to export from
+        uv differently.
+
+        :param requirements_path: The path of the requirements.txt file to write to.
+        :returns: A list of strings forming the export script.
+        """
         export_command = [
             "uv",
             "export",
-            "--format=requirements-txt",
+            "--no-dev",
+            "--format=requirements.txt",
             f"--output={requirements_path}",
-            # Requires uv.lock exists and is up to date
-            "--locked",
         ]
-
-        if self._options.uv_extra:
-            export_command.append(f"--with={','.join(sorted(self._options.uv_extra))}")
+        if self._options.uv_extras:
+            for extra in self._options.uv_extras:
+                export_command.append(
+                    f"--extra={extra}",
+                )
 
         return [shlex.join(export_command)]
     
-    def _get_pip_install_commands(self, requirements_path: Path) -> list[str]:
+    def _get_create_venv_commands(self) -> list[str]:
+        return [f'{self._get_uv()} venv --relocatable "${{CRAFT_PART_INSTALL}}']
+
+    def _get_pip_install_commands(self, requirements_path: pathlib.Path) -> list[str]:
+        """Get the commands for installing with pip.
+
+        Application-specific classes my override this if they need to install
+        differently.
+
+        :param requirements_path: The path of the requirements.txt file to write to.
+        :returns: A list of strings forming the install script.
+        """
         pip = self._get_pip()
         return [
-            f"{pip} install --requirement={requirements_path}",
-            f"{pip} install --no-deps .",
-            f"{pip} check",
+            f'{pip} install --python "${{CRAFT_PART_INSTALL}}/bin/python -r {requirements_path.resolve()}',
+            f'{pip} install --python "${{CRAFT_PART_INSTALL}} "${{CRAFT_PART_NAME}} @ ${{CRAFT_PART_BUILD}}'
         ]
 
+    @override
     def _get_package_install_commands(self) -> list[str]:
+        """Return a list of commands to run during the build step."""
         requirements_path = self._part_info.part_build_dir / "requirements.txt"
+
         return [
             *self._get_uv_export_commands(requirements_path),
             *self._get_pip_install_commands(requirements_path),
         ]
-
+    
+    @override
+    def get_build_environment(self) -> dict[str, str]:
+        build_environment = super().get_build_environment()
+        build_environment["UV_FROZEN"] = "1"
+        build_environment["UV_PYTHON_DOWNLOADS"] = "never"
+        return build_environment
--- a/craft_parts/plugins/uv_plugin.py
+++ b/craft_parts/plugins/uv_plugin.py
@@ -1,0 +1,100 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2024 Canonical Ltd.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License version 3 as published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""The uv plugin."""
+
+from .base import BasePythonPlugin
+from .properties import PluginProperties
+from typing import Literal, Annotated
+from pathlib import Path
+from overrides import override
+from .validator import PluginEnvironmentValidator
+import subprocess
+from craft_parts.errors import PluginEnvironmentValidationError
+import pydantic
+import shlex
+
+class UvPluginProperties(PluginProperties, frozen=True):
+    plugin: Literal["uv"] = "uv"
+    uv_extra: set[str] = pydantic.Field(
+        default_factory=set,
+        title="Optional dependency groups",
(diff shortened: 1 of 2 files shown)


craft_parts/plugins/uv_plugin.py
```diff
@@ -0,0 +1,155 @@
...
+
+class UvPlugin(BasePythonPlugin):
+    """A plugin to build python parts."""
+
+    properties_class = UvPluginProperties
+    validator_class = UvPluginEnvironmentValidator
+    _options: UvPluginProperties
+    
+    def _get_uv(self) -> str:
+        return f'"${{HOME}}/.cargo/bin/uv'
```
lengau, line 89: This needs to be more general - e.g. checking for a correct `uv` in `$PATH` (if the user has the snap installed or has it installed systemwide)

craft_parts/plugins/uv_plugin.py
```diff
@@ -0,0 +1,143 @@
...
+        # Guaranteed not-none by the validator class, which already asserted uv
+        # was installed
+        return cast(str, uv)
+
+    @override
+    def _get_pip(self) -> str:
+        return f"{self._get_uv()} pip"
+
+    def _get_create_venv_commands(self) -> list[str]:
+        return [f'{self._get_uv()} venv --relocatable "${{CRAFT_PART_INSTALL}}']
```
lengau, line 104:
```suggestion
        return [f'{self._get_uv()} venv --relocatable {self._get_venv_directory()}']
```

craft_parts/plugins/uv_plugin.py
```diff
@@ -0,0 +1,143 @@
...
+            requirements_path.resolve(),
+            *[f' --extra "{extra}"' for extra in self._options.uv_extras],
+        ]
+
+        install_commands = [shlex.join(sync_command)]
+
+        install_commands.extend([
+            f'{self._get_pip()} install --python "${{CRAFT_PART_INSTALL}}" "{package} @ ${{CRAFT_PART_BUILD}}"'
+            for package in self._options.uv_install_packages
+        ])
```
lengau, line 128: Now that we're using `uv sync --no-editable` I'm not sure we need this.

craft_parts/plugins/uv_plugin.py
```diff
@@ -0,0 +1,143 @@
...
+
+        sync_command = [
+            self._get_uv(),
+            "sync",
+            "--no-dev",
+            "--no-editable",
+            "--python",
+            '"${{CRAFT_PART_INSTALL}}/bin/python"',
+            "-r",
+            requirements_path.resolve(),
```
lengau, line 119:
```suggestion
```

docs/common/craft-parts/reference/plugins/uv_plugin.rst
```diff
@@ -0,0 +1,100 @@
...
+=========
+
+The uv plugin can be used for Python projects that use the uv build tool.
+
+Keywords
+--------
+
+A sequence of subsections specifying the keywords provided by the plugin and
+describe how the plugin uses them using the same format for keywords used in
+the :ref:`part_properties` reference.
```
medubelko, line 13: This looks like it's leftover text from the template.

docs/common/craft-parts/reference/plugins/uv_plugin.rst
```diff
@@ -0,0 +1,100 @@
...
+**Type:** list of strings
+
+Extra dependency groups to build with. Each element of the list is passed
+exactly as ``--extra GROUP``.
+
+Environment variables
+---------------------
+
+Along with the variables defined by the :ref:`Python plugin <craft_parts_python_plugin-environment_variables>`,
+this plugin defines additional variables defined below.
```
medubelko, line 30:
```suggestion
Along with the environment variables defined by the :ref:`Python plugin
<craft_parts_python_plugin-environment_variables>`, this plugin responds to its
own special variables.
```

docs/common/craft-parts/reference/plugins/uv_plugin.rst
```diff
@@ -0,0 +1,100 @@
...
+
+.. note::
+
+    This section describes how this plugin uses ``uv``-specific variables. For
+    more information, as well as a complete list of environment variables
+    for ``uv``, see the `uv documentation <https://docs.astral.sh/uv/configuration/environment/>`__.
```
medubelko, line 36: Small formatting convention that I'm trying to enforce:

- For software names, even if they share the name with their CLI command and start with a lowercase letter, inline code isn't needed.
- When you _are_ talking about the software's command, then dressing it with inline code is fine.

A convention that started before me was that we use two spaces for indentation in RST.

```suggestion
  This section describes how this plugin uses uv-specific environment variables. For
  more information, as well as a complete list of environment variables
  for uv, see the `uv documentation <https://docs.astral.sh/uv/configuration/environment/>`_.
```

docs/common/craft-parts/reference/plugins/uv_plugin.rst
```diff
@@ -0,0 +1,100 @@
...
+
+Whether or not to update the :file:`uv.lock` file. If true, :file:`uv.lock`
+must exist and will be used as the single source of truth for dependency
+versions, with no attempt made to update them before installation.
+
+UV_PROJECT_ENVIRONMENT
+~~~~~~~~~~~~~~~~~~~~~~
+**Default value:** See below
+
+A filepath to the Python virtual environment to build with. By default, this
```
medubelko, line 50: > filepath

I prefer "path" for simplicity and abstracting away complexities around symlinks, API endpoints, and so forth, but YMMV.

docs/common/craft-parts/reference/plugins/uv_plugin.rst
```diff
@@ -0,0 +1,100 @@
...
+variable populates itself with the directory that this plugin will create the
+virtual environment.
```
medubelko, line 52:
```suggestion
variable populates itself with the directory that this plugin will create in the
virtual environment.
```

docs/common/craft-parts/reference/plugins/uv_plugin.rst
```diff
@@ -0,0 +1,100 @@
...
+
+UV_PYTHON_DOWNLOADS
+~~~~~~~~~~~~~~~~~~~
+**Default value:** never
+
+Whether or not to automatically download Python if the requested version is
+missing.
```
medubelko, line 59: Since the default is a non-standard enum `never`, should we describe how to turn it on? Is the complement `always`? This is the first time I'm running into this in our projects, so if this is an existing pattern, ignore me.

docs/common/craft-parts/reference/plugins/uv_plugin.rst
```diff
@@ -0,0 +1,100 @@
...
+~~~~~~~~~
+**Default value:** ``PARTS_PYTHON_INTERPRETER``
+
+The version of Python that ``uv`` should use. See 
+:ref:`Python plugin environment variables <craft_parts_python_plugin-environment_variables>`
+for more information.
+
+A sequence of subsections specifying the environment variables that the plugin
+defines in the build environment, their default values, and what they are
+typically used for.
```
medubelko, line 71: This looks like more template text.

docs/common/craft-parts/reference/plugins/uv_plugin.rst
```diff
@@ -0,0 +1,100 @@
...
+
+UV_PYTHON_PREFERENCE
+~~~~~~~~~~~~~~~~~~~~
+**Default value:** only-system
+
+Whether ``uv`` should prefer system or ``uv``-managed Python versions.
```
medubelko, line 77:
```suggestion
Whether uv should prefer system or uv-managed Python versions.
```

medubelko, line 77: Same as an earlier comment; should we describe what the other possible values are?

docs/common/craft-parts/reference/plugins/uv_plugin.rst
```diff
@@ -0,0 +1,100 @@
...
+
+
+Dependencies
+------------
+
+``uv`` must already be installed on the build system in order to use this
+plugin. For installation instructions, see `uv documentation <https://docs.astral.sh/uv/getting-started/installation/>`_.
```
medubelko, line 84:
```suggestion
uv must already be installed on the build system in order to use this plugin.
For installation instructions, see the `uv documentation
<https://docs.astral.sh/uv/getting-started/installation/>`_.
```

docs/common/craft-parts/reference/plugins/uv_plugin.rst
```diff
@@ -0,0 +1,100 @@
...
+
+How it works
+------------
+
+During the build step, the plugin performs the following actions:
+
+* It creates a virtual environment directly into the ``${CRAFT_PART_INSTALL}``
```
medubelko, line 91:
```suggestion
* It creates a virtual environment in the ``${CRAFT_PART_INSTALL}``
```

This was a typo in the original document. :(

medubelko (changes requested):

docs/common/craft-parts/reference/plugins/uv_plugin.rst
bepri: This is correct - this variable is the path to the virtual environment itself that the plugin will make, not an existing virtual environment that it modifies. Is the phrasing bad?

docs/common/craft-parts/reference/plugins/uv_plugin.rst
bepri: This one is much less standard and actually required some digging - the uv documentation page I link didn't say either! I'll add the other values.

craft_parts/plugins/uv_plugin.py
```diff
@@ -0,0 +1,130 @@
...
+from .base import BasePythonPlugin
+from .properties import PluginProperties
+
+
+class UvPluginProperties(PluginProperties, frozen=True):
+    """The part properties used by the uv plugin."""
+
+    plugin: Literal["uv"] = "uv"
+
+    uv_extras: set[str] = pydantic.Field(
```
lengau, line 38: We also have dependency groups and potentially future options. I'm still unsure about this vs a `uv_extra_args` option.

docs/common/craft-parts/reference/plugins/uv_plugin.rst
lengau: How about

```suggestion
variable populates itself with the directory in which this plugin will create
the virtual environment.
```

docs/common/craft-parts/reference/plugins/uv_plugin.rst
lengau: It's a string, so maybe put it in quotes to indicate that?

craft_parts/plugins/uv_plugin.py
```diff
@@ -0,0 +1,130 @@
...
+    def get_build_environment(self) -> dict[str, str]:
+        """Return a dictionary with the environment to use in the build step."""
+        build_environment = super().get_build_environment()
+        build_environment["VIRTUAL_ENV"] = str(self._get_venv_directory().resolve())
+        build_environment["UV_PROJECT_ENVIRONMENT"] = build_environment["VIRTUAL_ENV"]
+        build_environment["UV_FROZEN"] = "true"
+        build_environment["UV_PYTHON_DOWNLOADS"] = "never"
+        build_environment["UV_PYTHON"] = '"${PARTS_PYTHON_INTERPRETER}"'
+        build_environment["UV_PYTHON_PREFERENCE"] = "only-system"
+        return build_environment
```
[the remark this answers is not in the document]
bepri: Agreed - I actually never knew you could `or` two dicts together! Thanks.

craft_parts/plugins/uv_plugin.py
bepri: I think I lean towards adding `uv_groups` rather than a generic `uv_extra_args` for both - the former is a lot easier to document (both for us and a user).

.github/workflows/tests.yaml
```diff
@@ -113,6 +113,10 @@ jobs:
           echo "::group::Wait for snap to complete"
           snap watch --last=install
           echo "::endgroup::"
+      - name: Install uv
+        uses: astral-sh/setup-uv@v4
+        with:
+          version: "0.5.7"
```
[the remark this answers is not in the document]
bepri: It doesn't execute it, but it does need it to validate that commands are built properly:
https://github.com/canonical/craft-parts/actions/runs/12301146783/job/34331041648#step:8:453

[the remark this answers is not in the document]
tigarmo: I believe the point is that those tests should mock out even the existence of the `uv` binary (this is how most unittests in craft-parts operate)

craft_parts/plugins/uv_plugin.py
```diff
@@ -0,0 +1,131 @@
...
+class UvPlugin(BasePythonPlugin):
+    """A plugin to build python parts."""
+
+    properties_class = UvPluginProperties
+    validator_class = UvPluginEnvironmentValidator
+    _options: UvPluginProperties
+
+    def _get_uv(self) -> str:
+        uv = shutil.which("uv")
+        # Guaranteed not-none by the validator class, which already asserted uv
```
tigarmo, line 89: This is actually not guaranteed: in the (unlikely) event that the project has a part called `uv-deps` that does *not* actually provide `uv`, this would fail.
Therefore I think it's sensible to check for `None` here and fail

tests/integration/plugins/test_uv.py
```diff
@@ -0,0 +1,272 @@
...
+    plugins.unregister_all()
+
+
+def teardown_module():
+    plugins.unregister_all()
+
+
+@pytest.fixture
+def uv_parts_simple() -> Callable[[], dict[str, Any]]:
+    def _inner() -> dict[str, Any]:
```
tigarmo, line 41: how come this is an inner function?

tests/integration/plugins/test_uv.py
```diff
@@ -0,0 +1,272 @@
...
+        application_name="test_uv",
+        cache_dir=new_dir,
+        partitions=partitions,
+    )
+    actions = lf.plan(Step.PRIME)
+
+    with lf.action_executor() as ctx:
+        ctx.execute(actions)
+
+    primed_script = Path(lf.project_info.prime_dir, "bin", "mytestuv")
```
tigarmo, line 69: I think you should execute this primed_script and check that it prints `"it works with uv too!"`

tests/integration/plugins/test_uv.py
```diff
@@ -0,0 +1,272 @@
...
+                  craftctl default
+            """
+        )
+
+        return cast(dict[str, Any], yaml.safe_load(parts_yaml))
+
+    return _inner
+
+
+def test_find_payload_python_bad_version(new_dir, partitions, uv_parts_complex):
```
tigarmo, line 209: I'm undecided on whether *every* python plugin subclass needs to repeat these tests
since it's already here I guess might as well leave it

tigarmo (changes requested):
nice!

.github/workflows/tests.yaml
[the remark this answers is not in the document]
lengau: ^

craft_parts/plugins/uv_plugin.py
lengau: I'm fine with that. Let's do it!

craft_parts/plugins/uv_plugin.py
lengau: Maybe we should even eliminate this and just use `uv` in the commands (getting from the part's path)? That's what most plugins do and we just fail there with a command not found.

tests/integration/plugins/test_uv.py
lengau: Good call - that will ensure that the relocated venv is correctly usable.

tests/integration/plugins/test_uv.py
lengau: Probably not, since it's really 3 tests for the `BasePythonPlugin`.

lengau (changes requested):

medubelko (approved):

tests/integration/plugins/test_uv.py
bepri: I've created an issue to fix this next pulse: https://github.com/canonical/craft-parts/issues/935

tests/integration/plugins/test_uv.py
bepri: It originally had a parameter that got removed by other PR feedback (which itself only existed because I made that fixture by just copying the `_complex` variant and changing a few things mindlessly). fixed in most recent push

tigarmo (approved):
🚢 🚢

lengau (approved):
🚀 🦀 🌈

mr-cal (approved):
:skier: :mountain: :medal_sports:

PR #931, opened 2024-12-11
feat: npm plugin file listing

mattculler:

Implement origin package -> file mapping for the npm plugin, to be used in xattr tagging.

-----
(CRAFT-3742)

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -39,9 +39,6 @@ jobs:
       - name: Run mypy
         run: |
           make test-mypy
-      - name: Run pydocstyle
-        run: |
-          make test-pydocstyle
       - name: Run lint-docs
         run: |
           tox run -e lint-docs
--- a/Makefile
+++ b/Makefile
@@ -67,7 +67,7 @@ install: clean ## Install python package.
 	python setup.py install
 
 .PHONY: lint
-lint: test-black test-codespell test-ruff test-mypy test-pydocstyle test-pyright ## Run all linting tests
+lint: test-black test-codespell test-ruff test-mypy test-pyright ## Run all linting tests
 
 .PHONY: release
 release: dist ## Release with twine.
@@ -97,10 +97,6 @@ test-integrations: ## Run integration tests.
 .PHONY: test-mypy
 test-mypy:
 	mypy $(SOURCES)
-
-.PHONY: test-pydocstyle
-test-pydocstyle:
-	pydocstyle craft_parts
 
 .PHONY: test-pyright
 test-pyright:
--- a/craft_parts/plugins/base.py
+++ b/craft_parts/plugins/base.py
@@ -22,6 +22,7 @@ import abc
 import pathlib
 import textwrap
 from copy import deepcopy
+from dataclasses import dataclass
 from typing import TYPE_CHECKING
 
 from overrides import override
@@ -36,7 +37,15 @@ if TYPE_CHECKING:
     from craft_parts import infos
 
 
-PackageFileList = dict[tuple[str, str], set[pathlib.Path]]
+@dataclass(frozen=True)
+class Package:
+    """A dataclass that uniquely identifies a package."""
+
+    name: str
+    version: str
+
+
+PackageFileList = dict[Package, set[pathlib.Path]]
 
 
 class Plugin(abc.ABC):
--- a/craft_parts/plugins/base.py
+++ b/craft_parts/plugins/base.py
@@ -36,6 +36,9 @@ if TYPE_CHECKING:
     from craft_parts import infos
 
 
+PackageFileList = dict[tuple[str, str], set[pathlib.Path]]
+
+
 class Plugin(abc.ABC):
     """The base class for plugins.
 
@@ -90,6 +93,10 @@ class Plugin(abc.ABC):
         :param action_properties: The properties to store.
         """
         self._action_properties = deepcopy(action_properties)
+
+    def get_file_list(self) -> PackageFileList:
+        """Get a mapping of (package name, package version) -> installed file."""
+        return {}
 
 
 class BasePythonPlugin(Plugin):
--- a/craft_parts/plugins/base.py
+++ b/craft_parts/plugins/base.py
@@ -37,7 +37,7 @@ if TYPE_CHECKING:
     from craft_parts import infos
 
 
-@dataclass(frozen=True)
+@dataclass(frozen=True, slots=True)
 class Package:
     """A dataclass that uniquely identifies a package."""
 
@@ -45,7 +45,7 @@ class Package:
     version: str
 
 
-PackageFileList = dict[Package, set[pathlib.Path]]
+PackageFiles = dict[Package, set[pathlib.Path]]
 
 
 class Plugin(abc.ABC):
@@ -103,7 +103,7 @@ class Plugin(abc.ABC):
         """
         self._action_properties = deepcopy(action_properties)
 
-    def get_file_list(self) -> PackageFileList:
+    def get_files(self) -> PackageFiles:
         """Get a mapping of (package name, package version) -> installed file."""
         return {}
 
--- a/craft_parts/plugins/npm_plugin.py
+++ b/craft_parts/plugins/npm_plugin.py
@@ -34,7 +34,7 @@ from typing_extensions import Self
 from craft_parts.errors import InvalidArchitecture
 
 from . import validator
-from .base import PackageFileList, Plugin
+from .base import Package, PackageFileList, Plugin
 from .properties import PluginProperties
 
 logger = logging.getLogger(__name__)
@@ -365,7 +365,7 @@ def _append_package_dir(
         for file_or_dir in dirnames + filenames:
             pkg_contents.add(walk_iteration_root / file_or_dir)
 
-    key = (pkg_name, pkg_version)
+    key = Package(pkg_name, pkg_version)
     if key in file_list:
         # It appears we have two installs of the same package and version
         # at different points in the tree.  This is fine.  If we ever care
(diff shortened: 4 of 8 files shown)


.github/workflows/tests.yaml
```diff
@@ -39,9 +39,6 @@ jobs:
       - name: Run mypy
         run: |
           make test-mypy
-      - name: Run pydocstyle
```
tigarmo, line 42: how come?

Makefile
```diff
@@ -67,7 +67,7 @@ install: clean ## Install python package.
 	python setup.py install
 
 .PHONY: lint
-lint: test-black test-codespell test-ruff test-mypy test-pydocstyle test-pyright ## Run all linting tests
```
tigarmo, line 70: same question here

craft_parts/plugins/base.py
```diff
@@ -36,6 +36,9 @@
     from craft_parts import infos
 
 
+PackageFileList = dict[tuple[str, str], set[pathlib.Path]]
```
tigarmo, line 39: I'm wondering whether we're restricting ourselves too much here with a tuple with exactly two values. For example, what happens if we decide to add more metadata (say, origin url)? 

Maybe we should use a dataclass instead?

craft_parts/plugins/npm_plugin.py
```diff
@@ -320,3 +323,99 @@ def get_build_commands(self) -> list[str]:
...
+                if pkg_file.resolve() in links:
+                    link = links[pkg_file]
+                    to_add[pkg_tuple].update(link)
+
+        # Insert those links into our package files data structure
+        for pkg_tuple, pkg_files in to_add.items():
+            file_list[pkg_tuple].update(pkg_files)
+
+    @override
+    def get_file_list(self) -> PackageFileList:
```
tigarmo, line 411: this is just a suggestion, but style-wise I would put this public method *before* the private ones. In my opinion it makes the code easier to read "sequentially"

craft_parts/plugins/npm_plugin.py
```diff
@@ -320,3 +323,99 @@ def get_build_commands(self) -> list[str]:
...
+                self._append_package_dir(pkg_dir, file_list, scope_name)
+
+    def _append_symlinks(self, link_dir: Path, file_list: PackageFileList) -> None:
+        """Append symlinks in link_dir under the appropriate package in file_list."""
+        # Map target paths to links.  This seems backwards but using the
+        # targets as keys is more efficient in the loop below.  It's unlikely
+        # but there could be multiple symlinks here that point at the same
+        # target file, hence the defaultdict(set).
+        links = defaultdict(set)
+        for symlink in list(link_dir.iterdir()):
```
tigarmo, line 392: why the list()?

tests/integration/plugins/test_npm.py
```diff
@@ -166,3 +183,69 @@ def test_npm_plugin_include_node(new_dir, partitions):
...
+    )
+    actions = lifecycle.plan(Step.BUILD)
+
+    with lifecycle.action_executor() as ctx:
+        ctx.execute(actions)
+
+    part_name = list(parts["parts"].keys())[0]
+    actual_file_list = lifecycle._executor._handler[part_name]._plugin.get_file_list()
+
+    # This example bundles in node, which brings a ton of other dependencies -
```
tigarmo, line 205: This is fine but I think this feature would really benefit from a smaller unit test that creates a fake, smaller file tree and then matches the contents of get_file_list() exactly

tests/integration/plugins/test_npm.py
```diff
@@ -166,3 +183,69 @@ def test_npm_plugin_include_node(new_dir, partitions):
...
+    # this is perfect for checking all sorts of weird behavior.
+
+    for (pkg_name, _), pkg_files in actual_file_list.items():
```
tigarmo, line 208: the way this loop is structured is too fragile: if the file list for some reason doesn't have `npm-hello` or `npm` it will still succeed.

tests/integration/plugins/test_npm.py
```diff
@@ -166,3 +183,69 @@ def test_npm_plugin_include_node(new_dir, partitions):
...
+    assert ar300 in actual_file_list
+    assert len(actual_file_list[ar300]) == 4, ar300
+
+    # Added "index.d.ts" file, absent from previous versions
+    ar500 = ("ansi-regex", "5.0.0")
+    assert ar500 in actual_file_list
+    assert len(actual_file_list[ar500]) == 5, ar500
+
+    # Between 5.0.0 and 5.0.1 they seem to have stopped packaging the readme;
+    # back to 4 files per install.
```
tigarmo, line 243: but the assertion says `8`?

tests/integration/plugins/test_npm.py
```diff
@@ -166,3 +183,69 @@ def test_npm_plugin_include_node(new_dir, partitions):
...
+        cache_dir=new_dir,
+        partitions=partitions,
+    )
+    actions = lifecycle.plan(Step.BUILD)
+
+    with lifecycle.action_executor() as ctx:
+        ctx.execute(actions)
+
+    part_name = list(parts["parts"].keys())[0]
+    actual_file_list = lifecycle._executor._handler[part_name]._plugin.get_file_list()
```
tigarmo, line 203: I think you should check at least some of the files (the values), in addition to the package names and versions

tigarmo (changes requested):

.github/workflows/tests.yaml
mattculler: Ruff does the pydocstyle tests.  The existing config wanted me to provide docstrings for public methods that were overrides, which we no longer require, so I downstreamed the relevant starbase config and removed the discrete pydocstyle usage.

craft_parts/plugins/base.py
mattculler: Yeah that's a good point and shouldn't be a big change, I'll do that.

craft_parts/plugins/npm_plugin.py
mattculler: I agree.  In fact these private methods don't need to be methods at all, I should be able to remove them from the class altogether.

craft_parts/plugins/npm_plugin.py
mattculler: I think it was an artifact of a previous implementation, removed.

tests/integration/plugins/test_npm.py
mattculler: Comment clarified.

tests/integration/plugins/test_npm.py
mattculler: Will do.

craft_parts/plugins/npm_plugin.py
mattculler: https://github.com/canonical/craft-parts/pull/931/commits/81141908fcd9641c6dda0b3d9d50badd358c7ec4

tests/integration/plugins/test_npm.py
mattculler: https://github.com/canonical/craft-parts/pull/931/commits/579fd4c638648e50eecd4814e51e2974871182c7

tests/integration/plugins/test_npm.py
mattculler: https://github.com/canonical/craft-parts/pull/931/commits/205557378f9aa07cbaa47c025c1b778dff3b2b0b

tests/integration/plugins/test_npm.py
mattculler: https://github.com/canonical/craft-parts/pull/931/commits/d385065bf4b95b769d2be393af8bbf4e3ce78e8d

craft_parts/plugins/npm_plugin.py
mattculler: https://github.com/canonical/craft-parts/pull/931/commits/c03240001586b77ac3b0d5b7ae4e071c3354467b

Makefile
mattculler: See above

craft_parts/plugins/base.py
mattculler: https://github.com/canonical/craft-parts/pull/931/commits/1997a6f7d3344d07c6beb7993f95ce064486894d

tests/integration/plugins/test_npm.py
mattculler: Ok will do

craft_parts/plugins/base.py
```diff
@@ -36,6 +37,17 @@
     from craft_parts import infos
 
 
+@dataclass(frozen=True)
```
lengau, line 40: Given that we'll be creating a lot of these, it might be worthwhile to have:

```suggestion
@dataclass(frozen=True, slots=True)
```

craft_parts/plugins/base.py
```diff
@@ -36,6 +37,17 @@
...
+class Package:
+    """A dataclass that uniquely identifies a package."""
+
+    name: str
+    version: str
+
+
+PackageFileList = dict[Package, set[pathlib.Path]]
```
lengau, line 48: Perhaps just `PackageFiles` since it's not a list?

craft_parts/plugins/base.py
```diff
@@ -91,6 +103,10 @@ def set_action_properties(self, action_properties: ActionProperties) -> None:
         """
         self._action_properties = deepcopy(action_properties)
 
+    def get_file_list(self) -> PackageFileList:
```
lengau, line 106: Same here.

```suggestion
    def get_files(self) -> PackageFiles:
```

tigarmo (approved):
thanks! Approved pending addressing @lengau 's comments

tests/integration/plugins/test_npm.py
mattculler: https://github.com/canonical/craft-parts/pull/931/commits/898e4ed53674c7caa5246b6d767ccbac1c066a51

craft_parts/plugins/base.py
mattculler: https://github.com/canonical/craft-parts/pull/931/commits/517e0ede0221c490f6df90106bdff9f6eff9c7bb

craft_parts/plugins/base.py
mattculler: https://github.com/canonical/craft-parts/pull/931/commits/517e0ede0221c490f6df90106bdff9f6eff9c7bb

craft_parts/plugins/base.py
mattculler: https://github.com/canonical/craft-parts/pull/931/commits/517e0ede0221c490f6df90106bdff9f6eff9c7bb

lengau (approved):
LGTM, let's go! 🚀

PR #946, opened 2024-12-18
docs: Fix poetry-with description

carlcsaposs-canonical:
--with is required for optional and non-optional dependencies

see https://github.com/python-poetry/poetry-plugin-export/pull/212 and https://github.com/python-poetry/poetry-plugin-export/issues/299

### DIFF
--- a/craft_parts/plugins/poetry_plugin.py
+++ b/craft_parts/plugins/poetry_plugin.py
@@ -37,8 +37,8 @@ class PoetryPluginProperties(PluginProperties, frozen=True):
 
     poetry_with: set[str] = pydantic.Field(
         default_factory=set,
-        title="Optional dependency groups",
-        description="optional dependency groups to include when installing.",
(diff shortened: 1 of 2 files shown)


craft_parts/plugins/poetry_plugin.py
```diff
@@ -37,8 +37,8 @@ class PoetryPluginProperties(PluginProperties, frozen=True):
 
     poetry_with: set[str] = pydantic.Field(
         default_factory=set,
-        title="Optional dependency groups",
-        description="optional dependency groups to include when installing.",
+        title="Dependency groups to include",
+        description="optional and non-optional dependency groups to include. By default, only the main dependencies are included.",
```
lengau, line 41:
```suggestion
        description="dependency groups to include. By default, only the main dependencies are included.",
```

lengau (approved):
Thanks!

mr-cal (approved):

PR #961, opened 2025-01-09
feat(plugins): cargo-package plugin

lengau:
This adds a cargo-package plugin that allows packaging a cargo crate to a local directory-based registry.

Using the plugin has a side-effect of making all rust builds offline.

-----
CRAFT-3765


docs/reference/changelog.rst
```diff
@@ -2,6 +2,13 @@
 Changelog
 *********
 
+2.3.0 (2025-XX-YY)
+------------------
+
+New features:
+
+- Add a ``cargo-package`` plugin for offline rust crate building.
```
medubelko, line 10:
```suggestion
- Add a ``cargo-package`` plugin for offline Rust crate building.
```

medubelko (approved):

craft_parts/plugins/cargo_package_plugin.py
```diff
@@ -0,0 +1,162 @@
...
+"""
+
+
+class CargoPackagePluginProperties(PluginProperties, frozen=True):
+    """The part properties used by the cargo-package plugin."""
+
+    plugin: Literal["cargo-package"] = "cargo-package"
+
+    # part properties required by the plugin
+    cargo_package_features: UniqueList[str] = []
```
dariuszd21, line 55: Maybe for the frozen class is not that important, but you may want to update it.

```suggestion
    cargo_package_features: UniqueList[str] = Field(default_factory=list)
```

craft_parts/plugins/cargo_package_plugin.py
```diff
@@ -0,0 +1,162 @@
...
+            return
+
+        self.validate_dependency(
+            dependency=options.cargo_package_cargo_command,
+            plugin_name=options.plugin,
+            part_dependencies=part_dependencies,
+        )
+
+
+class CargoPackagePlugin(Plugin):
```
dariuszd21, line 84: On a separate note, outside of scope of this item.
Would it be difficult to reuse this plugin for creating crates that may be build against and published to the given remote repository ?

dariuszd21 (approved):
It looks nice and clean.
I left few comments, but they're not blocking ones

craft_parts/plugins/cargo_package_plugin.py
```diff
@@ -0,0 +1,162 @@
...
+        )
+
+        return [
+            self._get_package_command(),
+            textwrap.dedent(
+                f"""\
+                    pushd target/package
+                    for package in *.crate; do
+                        tar xf $package
+                        rm $package
```
mr-cal, line 153: I don't expect spaces in package names but may as well:

```suggestion
                        tar xf "$package"
                        rm "$package"
```

craft_parts/plugins/cargo_package_plugin.py
```diff
@@ -0,0 +1,162 @@
...
+                    done
+                    for package in */; do
+                        echo '{{"files":{{}}}}' > $package/.cargo-checksum.json
```
mr-cal, line 156:
```suggestion
                        echo '{{"files":{{}}}}' > "$package/.cargo-checksum.json"
```

craft_parts/plugins/cargo_package_plugin.py
```diff
@@ -0,0 +1,162 @@
...
+        return set()
+
+    @override
+    def get_build_packages(self) -> set[str]:
+        """Return a set of required packages to install in the build environment."""
+        return set()
+
+    @property
+    def _registry_dir(self) -> pathlib.Path:
+        return self._part_info.project_info.dirs.stage_dir / "cargo_registry"
```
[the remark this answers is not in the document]
mr-cal: Is this what ST143 is about?

mr-cal (approved):
:+1:

craft_parts/plugins/cargo_package_plugin.py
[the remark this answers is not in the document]
lengau: Yeah, this is what we need ST144 to solve.

craft_parts/plugins/cargo_package_plugin.py
```diff
@@ -0,0 +1,162 @@
...
+
+    @override
+    def get_build_environment(self) -> dict[str, str]:
+        """Return a dictionary with the environment to use in the build step."""
+        return {}
+
+    def _get_package_command(self) -> str:
+        options = cast(CargoPackagePluginProperties, self._options)
+        cargo = options.cargo_package_cargo_command
+        features = ",".join(options.cargo_package_features)
```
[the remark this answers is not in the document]
lengau: From the cargo docs:

> Feature names may include characters from the [Unicode XID standard](https://unicode.org/reports/tr31/) (which includes most letters), and additionally allows starting with _ or digits 0 through 9, and after the first character may also contain -, +, or ..

> Note: [crates.io](https://crates.io/) imposes additional constraints on feature name syntax that they must only be [ASCII alphanumeric](https://doc.rust-lang.org/std/primitive.char.html#method.is_ascii_alphanumeric) characters or _, -, or +.

I think we're fine to presume that the features won't contain whitespace.

craft_parts/plugins/cargo_package_plugin.py
```diff
@@ -0,0 +1,162 @@
...
+        if features:
+            features = f"--features {features}"
+        return f"{cargo} package --no-verify {features}"
```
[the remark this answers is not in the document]
lengau: I'd rather have `--no-verify` be the default in order to prevent unnecessary builds. We could include an opt-in option for it, but I think that would be a separate feature in the plugin if we decide we need it. In general though this would only be used during debugging, so I'm not sure of its value.

craft_parts/plugins/cargo_package_plugin.py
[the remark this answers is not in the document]
lengau: Yes, and this part has to depend on that part. Basically, right now you have to recreate the full dependency tree.

lengau: No, but it would probably be easier to just write a separate plugin for that since that would just be a matter of running `cargo package` and `cargo publish` without unpacking the crates or doing any of the other hacky work to make a registry directory work.

craft_parts/plugins/cargo_package_plugin.py
```diff
@@ -0,0 +1,163 @@
...
+
+    @override
+    def get_build_commands(self) -> list[str]:
+        """Return a list of commands to run during the build step."""
+        # We do not want this implementation detail exposed in the run script
+        # This sets the default registry for all parts to use.
+        cargo_config = pathlib.Path("~/.cargo/config.toml").expanduser().resolve()
+        if not cargo_config.exists():
```
sergiusens, line 128: how does this play with destructive mode?

lengau: It will destroy your user's cargo config in destructive mode.

lengau: With ST143 though we'll instead set `CARGO_HOME` and use that.

lengau: This now uses a project-specific CARGO_HOME under the work directory.

lengau:
Re-requesting @mr-cal as maintainer :-)

craft_parts/plugins/cargo_package_plugin.py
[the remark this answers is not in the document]
sergiusens: until ST144 solves it, it probably should live in `parts/cargo_registry` instead, if not we introduce a backwards incompatible change

mr-cal (approved):
:man_cook: :kiss:

craft_parts/executor/environment.py
```diff
@@ -160,6 +160,8 @@ def _get_global_environment(info: ProjectInfo) -> dict[str, str]:
         "CRAFT_ARCH_TRIPLET_BUILD_FOR": info.arch_triplet_build_for,
         "CRAFT_PARALLEL_BUILD_COUNT": str(info.parallel_build_count),
         "CRAFT_PROJECT_DIR": str(info.project_dir),
+        # Build system configurations for the project.
+        "CARGO_HOME": str(info.dirs.work_dir / ".cargo"),
```
sergiusens, line 164: for these internal directories this could be /cargo, this path is changing anyways, but something to consider.

sergiusens (approved):

craft_parts/executor/environment.py
lengau: I prefer it un-hidden :-)

PR #965, opened 2025-01-10
docs: make plugin links external refs

medubelko:
Fixes the library side of CRAFT-3890.

- Makes the hosted Craft Parts docs an external source for this doc set.
- All interlinks in the plugin pages, which are often used by apps as modular docs, now use intersphinx and treats them as external links.
- Any potentially standalone pages should prefix references to other pages with `:external+craft-parts:ref:`. References to labels within the page can use simple `:ref:` like normal.

@lengau There's a flaw with this approach, though. Only already published pages can be linked to externally. If for example we add a new plugin page and want to link pages to it as part of the same code change, those links will break because the new page hasn't been published yet.

My thinking now is that the only straightforward approach is for the common dir generator script to transform the links after copying the files.

-----

### DIFF
--- a/docs/common/craft-parts/reference/plugins/ant_plugin.rst
+++ b/docs/common/craft-parts/reference/plugins/ant_plugin.rst
@@ -22,8 +22,9 @@ After a successful build, this plugin will:
 Keywords
 --------
 
-In addition to the common :ref:`plugin <part-properties-plugin>` and
-:ref:`sources <part-properties-sources>` keywords, this plugin provides the following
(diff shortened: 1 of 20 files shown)


medubelko:
@lengau Would appreciate your opinion on this.

lengau:
@medubelko Hmm yeah that's a problem :-/ I guess we'll have to go with the common dir generator doing that, although maybe @tigarmo and his sphinx wisdom have another idea.

medubelko:
Closing – not an appropriate solution.

PR #966, opened 2025-01-10
executor: allow capturing error message from scriptlets

cmatsuoka:
Change craftctl communication channel from named pipes to unix
sockets to consolidate the ctl server and error stream capture.

Process execution runs a selector to obtain data from i/o streams
to obtain the isolated error output. Allow this selection loop run
on a caller-specified selector with handlers registered to also
process interactions with craftctl. The captured error output when
executing scriptlets is passed to `ScriptletRunError`. Errors from
`craftctl default` execution in the build step raise `PluginBuildError`.
This is consistent with the exception raised by the built-in build
step handler.

-----
**Notes to reviewers:** see individual commits for pipe-to-socket change, and socket selector consolidation.

### DIFF
--- a/craft_parts/ctl.py
+++ b/craft_parts/ctl.py
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright 2017-2022 Canonical Ltd.
+# Copyright 2017-2025 Canonical Ltd.
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
--- a/craft_parts/ctl.py
+++ b/craft_parts/ctl.py
@@ -72,7 +72,7 @@ def _client(cmd: str, args: list[str]) -> str | None:
             f"utility is designed for use only in part scriptlets."
         ) from err
 
-    logger.debug("ctl socket:", ctl_socket_path)
+    logger.debug(f"ctl socket: {ctl_socket_path}")
 
     data = {"function": cmd, "args": args}
     ctl_socket = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
@@ -89,7 +89,7 @@ def _client(cmd: str, args: list[str]) -> str | None:
     message = feedback[1].strip() if len(feedback) > 1 else ""
     retval = None
 
-    logger.debug("status", status)
+    logger.debug(f"status: {status}")
 
     if status == "OK":
         # command has succeeded
--- a/craft_parts/ctl.py
+++ b/craft_parts/ctl.py
@@ -19,6 +19,7 @@
 import json
 import logging
 import os
+import socket
 import sys
 
 logger = logging.getLogger(__name__)
@@ -64,21 +65,20 @@ def _client(cmd: str, args: list[str]) -> str | None:
     :raise RuntimeError: If the command is invalid.
     """
     try:
-        call_fifo = os.environ["PARTS_CALL_FIFO"]
-        feedback_fifo = os.environ["PARTS_FEEDBACK_FIFO"]
+        ctl_socket_path = os.environ["PARTS_CTL_SOCKET"]
     except KeyError as err:
         raise RuntimeError(
             f"{err!s} environment variable must be defined.\nNote that this "
             f"utility is designed for use only in part scriptlets."
         ) from err
 
+    logger.debug("ctl socket:", ctl_socket_path)
+
     data = {"function": cmd, "args": args}
-
-    with open(call_fifo, "w") as fifo:
-        fifo.write(json.dumps(data))
-
-    with open(feedback_fifo) as fifo:
-        feedback = fifo.readline().split(" ", 1)
+    ctl_socket = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+    ctl_socket.connect(ctl_socket_path)
+    ctl_socket.send(bytes(json.dumps(data), encoding="utf8"))
+    feedback = ctl_socket.recv(1024).decode("utf8").split(" ")
 
     # response from server is in the form "<status> <message>" where
     # <status> can be either "OK" or "ERR".  Previous server versions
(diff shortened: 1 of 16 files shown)


craft_parts/errors.py
```diff
@@ -492,21 +492,17 @@ def __init__(self, *, part_name: str) -> None:
         super().__init__(brief=brief)
 
 
-class PluginBuildError(PartsError):
+class UserExecutionError(PartsError):
```
cmatsuoka, line 495: This is a superclass for both `PluginBuildError` and `ScriptletRunError` and contains the error processing that was previously in `PluginBuildError`. `ScriptletRunError` is raised when there's an error in scriptlet commands, but if something wrong happens during `craftctl default` execution in the build step `PluginBuildError` will be raised.

tests/integration/plugins/test_poetry.py
```diff
@@ -242,7 +242,7 @@ def _get_system_python_interpreter(self) -> str | None:
     actions = lf.plan(Step.PRIME)
 
     out = pathlib.Path("out.txt")
-    with out.open(mode="w") as outfile, pytest.raises(errors.ScriptletRunError):
+    with out.open(mode="w") as outfile, pytest.raises(errors.PluginBuildError):
```
cmatsuoka, line 245: Errors in the built-in handler executed using `craftctl default` now raise the same error as executing it directly.

craft_parts/errors.py
sergiusens: Why UserExecutionError and not something simpler like BuildStepError?. Is this supposed to be an ABC?

craft_parts/ctl.py
```diff
@@ -89,6 +89,8 @@ def _client(cmd: str, args: list[str]) -> str | None:
     message = feedback[1].strip() if len(feedback) > 1 else ""
     retval = None
 
+    logger.debug(f"status: {status}")
+
     if status == "OK":
```
bepri, line 94: It's not letting me comment on the `elif` of this block, so I'm commenting here: Should this raise something like `ScriptletRunError` instead now?

craft_parts/executor/step_handler.py
```diff
@@ -270,70 +269,69 @@ def run_scriptlet(
...
-                call_fifo.close()
-                feedback_fifo.close()
-
-            if process.returncode != 0:
+            except process.ProcessError as process_error:
                 raise errors.ScriptletRunError(
                     part_name=self._part.name,
                     scriptlet_name=scriptlet_name,
-                    exit_code=status,
+                    exit_code=process_error.result.returncode,
```
bepri, line 298: A similar naming scheme between `exit_code` or `returncode` would be nice - up to you which you pick, I like both.

craft_parts/errors.py
cmatsuoka: Will retain the name per reasons explained by @bepri and change the class to abstract.

craft_parts/executor/step_handler.py
```diff
@@ -270,70 +269,69 @@ def run_scriptlet(
...
+    def _ctl_server_selector(
+        self, step: Step, scriptlet_name: str, stream: socket.socket
+    ) -> selectors.BaseSelector:
+        selector = selectors.DefaultSelector()
+
+        def accept(sock: socket.socket, _mask: int) -> None:
+            conn, addr = sock.accept()
+            selector.register(conn, selectors.EVENT_READ, read)
+
+        def read(conn: socket.socket, _mask: int) -> None:
```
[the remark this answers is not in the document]
cmatsuoka: The parameters are passed by the handler dispatcher even if we don't use them in these specific handlers, because they're what `selector.select()` returns.

craft_parts/ctl.py
cmatsuoka: This will cause the `craftctl` execution to fail, which will turn into a `ScriptletRunError` in the main process. Except if this is `craftctl default` and the built-in step execution handler fails, in this case it will raise `PluginBuildError` if this is a build step, or a generic `PartsError` otherwise.

tests/unit/executor/test_step_handler.py
```diff
@@ -325,7 +326,7 @@ def setup(self, new_dir, partitions):
 
     def test_run_scriptlet(self, new_dir, capfd):
         sh = _step_handler_for_step(
-            Step.PULL,
+            Step.BUILD,
```
tigarmo, line 329: how come?

tigarmo (commented):
@cmatsuoka have you run snapcraft's spread tests with these changes? I think we should

tests/unit/executor/test_step_handler.py
cmatsuoka: This test was running a build step scriptlet in the pull step. Not a problem for what was being tested, but it was a non-realistic, inconsistent scenario.

cmatsuoka:
> @cmatsuoka have you run snapcraft's spread tests with these changes? I think we should

I'll prepare a snapcraft PR to to so.

bepri (approved):

pyproject.toml
```diff
@@ -348,6 +348,7 @@ lint.ignore = [
     "D213",  # Multi-line docstring summary should start at the second line (reason: pep257 default)
     "D215",  # Section underline is over-indented (reason: pep257 default)
     "A003",  # Class attribute shadowing built-in (reason: Class attributes don't often get bare references)
+    "A005",  # Standard library shadowing (reason: disambiguate using local names if required)
```
lengau, line 351:
```suggestion
```

This will fix the linting issue.

cmatsuoka: But it would make ruff complain about the file. I added a per-file config ignoring this for platform.py.

tigarmo (approved):

tigarmo (approved):

pyproject.toml
cmatsuoka: I haven't noticed that the noqa came with the master merge. Kept that and removed all my other attempts.

lengau (approved):
Brilliant!

I think a correct conventional-commits title for this would start with "feat(executor):"

Issue #661, opened 2024-02-12
CI failing due to xattrs not supported


lengau:
### Bug Description

It appears a GitHub runner update has disable xattrs or changed filesystems? Regardless, tests are failing now.

### To Reproduce

https://github.com/canonical/craft-parts/actions/workflows/tests.yaml

### part yaml

```shell
n/a
```

### Relevant log output

```shell
=================================== FAILURES ===================================
  ____________ TestOriginStagePackage.test_mark_origin_stage_package _____________
  
  path = '.tests-xattr-test-dir/bar'
  key = 'user.craft_parts.origin_stage_package', value = 'package'
  
      def write_xattr(path: str, key: str, value: str) -> None:
          """Add extended attribute metadata to a file.
      
          :param path: The file to add metadata to.
          :param key: The attribute key.
          :param value: The attribute value.
          """
          if sys.platform != "linux":
              raise RuntimeError("xattr support only available for Linux")
      
          # Extended attributes do not apply to symlinks.
          if os.path.islink(path):
              return
      
          key = f"user.craft_parts.{key}"
      
          try:
  >           os.setxattr(path, key, value.encode())
  E           OSError: [Errno 95] Operation not supported: '.tests-xattr-test-dir/bar'
  
  /home/runner/work/craft-parts/craft-parts/craft_parts/xattrs.py:77: OSError
  
  The above exception was the direct cause of the following exception:
  
  self = <tests.unit.packages.test_base.TestOriginStagePackage object at 0x7f5c3a1ba6f0>
  
      @pytest.mark.usefixtures("new_dir")
      def test_mark_origin_stage_package(self):
          test_dir = Path(".tests-xattr-test-dir")
          test_dir.mkdir()
      
          Path(test_dir / "foo").touch()
          Path(test_dir / "bar").touch()
      
  >       base.mark_origin_stage_package(str(test_dir), "package")
  
  /home/runner/work/craft-parts/craft-parts/tests/unit/packages/test_base.py:119: 
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
  /home/runner/work/craft-parts/craft-parts/craft_parts/packages/base.py:274: in mark_origin_stage_package
      write_origin_stage_package(file_path, stage_package)
[...]

PR #979, opened 2025-01-16
feat: flow for plugin package file mapping and xattr tagging

mattculler:

-----

Implement flow code, defined by https://docs.google.com/document/d/1wVtAYJouIXotd8iCZMZ_1VpmGpN0yoLfKcxeOBIgotw/edit?tab=t.0

(CRAFT-3741)

### DIFF
--- a/craft_parts/executor/part_handler.py
+++ b/craft_parts/executor/part_handler.py
@@ -990,7 +990,7 @@ class PartHandler:
 
     def _annotate_plugin_files(self) -> None:
         """Get the PackageFiles from the plugin and tag all the files it includes."""
-        for pkg, pkg_files in self._plugin.get_file_list().items():
+        for pkg, pkg_files in self._plugin.get_package_files().items():
             for pkg_file in pkg_files:
                 xattrs.write_xattr(
                     pkg_file,
--- a/craft_parts/executor/part_handler.py
+++ b/craft_parts/executor/part_handler.py
@@ -316,7 +316,7 @@ class PartHandler:
                 stdout=stdout,
                 stderr=stderr,
             )
-        
+
         # The plugin has by now built/unpacked packages, now get that file list mapping
         # and xattr the files.
         self._annotate_plugin_files()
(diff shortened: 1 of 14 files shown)


pyproject.toml
```diff
@@ -336,6 +340,9 @@ lint.ignore = [
 [tool.ruff.lint.flake8-annotations]
 allow-star-arg-any = true
 
+[tool.ruff.lint.flake8-builtins]
+builtins-allowed-modules = ["platform"]
```
mr-cal, line 344: Instead of a global ignore, can you add an ignore to the top of `packages.platform.py`?

mr-cal (approved):
Looks good to me, although it's not clear why the integration tests fail.  I wonder if github runners support xattrs.

mattculler:
> Looks good to me, although it's not clear why the integration tests fail. I wonder if github runners support xattrs.

One of the top results on kagi for "github ci runner xattr" was posted by this familiar face:

https://github.com/canonical/craft-parts/issues/661

So it seems I'll need to fix these tests by disabling the plugin xattr writes somehow, I'll investigate more.

pyproject.toml
mattculler: I don't think so, pretty sure it's allowable via configuration only: https://docs.astral.sh/ruff/rules/stdlib-module-shadowing/

craft_parts/xattrs.py
```diff
@@ -58,7 +60,7 @@ def read_xattr(path: str, key: str) -> str | None:
     return value.decode().strip()
 
 
-def write_xattr(path: str, key: str, value: str) -> None:
+def write_xattr(path: str | Path, key: str, value: str) -> None:
```
cmatsuoka, line 63: For general API consistency it would be better to use only Path here, unless convenience suffers too much. Then it would be simpler to test this only with Path instead of testing with str and Path.

craft_parts/executor/part_handler.py
```diff
@@ -47,6 +47,8 @@
 
 logger = logging.getLogger(__name__)
 
+_ORIGIN_PACKAGE_FORMAT = "origin_{plugin_name}_package"
```
cmatsuoka, line 50: This is a bit strange in isolate form but I see it's following the xattr pattern already used for stage packages.

craft_parts/packages/platform.py
```diff
@@ -1,3 +1,5 @@
+# noqa: A005
```
cmatsuoka, line 1: I wonder if this will render the editor modeline below useless.

cmatsuoka (approved):
LGTM!

PR #980, opened 2025-01-17
fix: use source-subdir in go-use plugin

### LINKED ISSUE #978: `go-use` plugin doesn't respect `source-subdir`
### Bug Description

Some go modules are hosted in subdirs of a git repo; for example, https://github.com/elazarl/goproxy/tree/master/ext has the `github.com/elazarl/goproxy/ext` module but the repo itself is github.com/elazarl/goproxy. One can use `source-subdir` to specify the build subdir, but the `go-use` plugin doesn't respect it.

### To Reproduce

.

### part yaml

```shell
parts:
  elazarl/goproxy/ext:
    override-pull: craftctl default; git checkout a0805db90819
    plugin: go-use
    source: https://github.com/elazarl/goproxy.git
    source-subdir: ext
```

### Relevant log output

```shell
.
```

tigarmo:
It's possible, and somewhat common, for a go module to be defined inside a repo's subdir; in this case we can define source-subdir, but the go-use plugin needs to use this - otherwise the repo root gets used.

Fixes #978

-----

### DIFF
--- a/craft_parts/plugins/go_use_plugin.py
+++ b/craft_parts/plugins/go_use_plugin.py
@@ -100,5 +100,5 @@ class GoUsePlugin(Plugin):
                 )
 
         return [
-            f"go work use {self._part_info.part_src_dir}",
+            f"go work use {self._part_info.part_src_subdir}",
         ]
--- a/docs/reference/changelog.rst
+++ b/docs/reference/changelog.rst
@@ -2,6 +2,15 @@
 Changelog
 *********
 
+
+X.Y.Z (2025-MM-DD)
+------------------
+
+Bug fixes:
+
+- Correctly handle ``source-subdir`` values on the ``go-use`` plugin.
+
+
 2.3.0 (2025-01-20)
 ------------------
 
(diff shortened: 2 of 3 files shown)


craft_parts/plugins/go_use_plugin.py
```diff
@@ -100,5 +100,5 @@ def get_build_commands(self) -> list[str]:
                 )
 
         return [
-            f"go work use {self._part_info.part_src_dir}",
+            f"go work use {self._part_info.part_src_subdir}",
```
lengau, line 103: Can you explain to me why these are the source dir/subdir and not the build dir/subdir? It's not a blocker for this, it's just that the latter makes more sense to me.

lengau (approved):
Good catch, thanks!

And thanks for undoing my changelog mistake.

craft_parts/plugins/go_use_plugin.py
sergiusens: in the before times, src was not copied of to build

sergiusens (approved):

PR #987, opened 2025-01-22
feat: backstage directory

lengau:

-----


craft_parts/parts.py
```diff
@@ -68,6 +68,9 @@ class PartSpec(BaseModel):
     build_attributes: list[str] = []
     organize_files: dict[str, str] = Field(default_factory=dict, alias="organize")
     overlay_files: list[str] = Field(default_factory=lambda: ["*"], alias="overlay")
+    backstage_files: list[RelativePathStr] = Field(
```
tigarmo, line 71: hmm do we need this? I thought we weren't going to expose the backstage to the users
it's like disney, looking behind the scenes breaks the magic

uv.lock
```diff
@@ -137,11 +137,11 @@ wheels = [
 
 [[package]]
 name = "cachetools"
-version = "5.5.0"
+version = "5.5.1"
```
tigarmo, line 140: how come this all got bumped?

craft_parts/executor/part_handler.py
```diff
@@ -373,6 +373,7 @@ def _run_stage(
             stdout=stdout,
             stderr=stderr,
         )
+        backstage_contents = self._migrate_to_backstage()
```
tigarmo, line 376: ah, this is interesting. I hadn't thought about what should happen if the user, say, has an `override-stage` without a `craftctl default`. Should the export dir still be migrated?

craft_parts/executor/part_handler.py
```diff
@@ -388,8 +389,51 @@ def _run_stage(
...
+                return
+            packages.fix_pkg_config(
+                prefix_prepend=self._part.backstage_dir,
+                pkg_config_file=Path(file_path),
+                prefix_trim=self._part.part_export_dir,
+            )
+            packages.fix_pkg_config(
+                prefix_prepend=self._part.backstage_dir,
+                pkg_config_file=Path(file_path),
+                prefix_trim=self._part.part_install_dir,
```
tigarmo, line 425: why is this one needed? How come it references the install dir?

craft_parts/executor/part_handler.py
tigarmo: My initial "gut feeling" is that export -> backstage should happen at exactly the same points as install -> stage. So either both happen, or neither happens.

_However_, we apparently always migrate the part's overlay files to stage, so why not do the same with the backstage?

_However however_, the export dir is populated in the same step as the install dir, so conceptually they seem closer and more related (to me).

craft_parts/executor/environment.py
```diff
@@ -91,7 +91,7 @@ def _basic_environment_for_part(part: Part, *, step_info: StepInfo) -> dict[str,
     :return: A dictionary containing the built-in environment.
     """
     part_environment = _get_step_environment(step_info)
-    paths = [part.part_install_dir, part.stage_dir]
+    paths = [part.part_install_dir, part.stage_dir, part.backstage_dir]
```
sergiusens, line 94: I would not add this default now, it would be rather dangerous and put us in a situation we cannot back out of. I would reserve it for plugin use only

tigarmo (commented):
Thanks! I've reviewed the code but I haven't looked at the tests yet.

craft_parts/executor/environment.py
```diff
@@ -162,6 +162,9 @@ def _get_global_environment(info: ProjectInfo) -> dict[str, str]:
         "CRAFT_PROJECT_DIR": str(info.project_dir),
         # Build system configurations for the project.
         "CARGO_HOME": str(info.dirs.work_dir / "cargo"),
+        "CRAFT_BACKSTAGE": str(info.backstage_dir),
```
sergiusens, line 165: similar, this would enable an override-build behavior.

craft_parts/executor/environment.py
```diff
@@ -228,6 +228,7 @@ def _get_step_environment(step_info: StepInfo) -> dict[str, str]:
         "CRAFT_PART_BUILD": str(step_info.part_build_dir),
         "CRAFT_PART_BUILD_WORK": str(step_info.part_build_subdir),
         "CRAFT_PART_INSTALL": str(step_info.part_install_dir),
+        "CRAFT_PART_BUILDOUT": str(step_info.part_buildout_dir),
```
sergiusens: plugins can access this through the property itself

craft_parts/executor/environment.py
```diff
@@ -170,9 +173,6 @@ def _get_global_environment(info: ProjectInfo) -> dict[str, str]:
     if Features().enable_partitions:
         global_environment.update(_get_environment_for_partitions(info))
 
-    global_environment["CRAFT_STAGE"] = str(info.stage_dir)
-    global_environment["CRAFT_PRIME"] = str(info.prime_dir)
-
```
sergiusens, line 175: this was probably here due to an ordered dict thing, reverting this, means we essentially revert this file back to its original state

craft_parts/executor/executor.py
```diff
@@ -165,6 +165,7 @@ def clean(self, initial_step: Step, *, part_names: list[str] | None = None) -> N
                 for stage_dir in self._project_info.stage_dirs.values():
                     if stage_dir.exists():
                         shutil.rmtree(stage_dir)
+                shutil.rmtree(self._project_info.backstage_dir)
```
sergiusens: let's leave cleaning for another body of work. This feels like the wrong place to do this

craft_parts/executor/part_handler.py
```diff
@@ -365,6 +365,7 @@ def _run_stage(
         :return: The stage step state.
         """
         self._make_dirs()
+        backstage_contents = self._migrate_to_backstage()
```
sergiusens: I guess this is the same as clean. I would have thought this migration could happen as the last step of build

craft_parts/executor/part_handler.py
```diff
@@ -388,8 +389,51 @@ def _run_stage(
...
+            packages.fix_pkg_config(
+                prefix_prepend=self._part.backstage_dir,
+                pkg_config_file=Path(file_path),
+                prefix_trim=self._part.part_buildout_dir,
+            )
+            packages.fix_pkg_config(
+                prefix_prepend=self._part.backstage_dir,
+                pkg_config_file=Path(file_path),
+                prefix_trim=self._part.part_install_dir,
+            )
```
sergiusens: do we have a use case now for this?

sergiusens:
All my comments are outdated as I commented on the first commit

sergiusens (changes requested):
Looks generally good, but going to ask that we stick to plugins only as agreed to in the spec

craft_parts/parts.py
lengau: We need this for the migrations to work.

craft_parts/executor/part_handler.py
lengau: If a plugin has something that gets placed in the install dir and then moved to the export dir, it might still have pkgconfig references to the install dir.

craft_parts/executor/environment.py
lengau: I thought I had removed this... That explains the unexpected test breakages!

craft_parts/executor/environment.py
lengau: This allows us to link to things in the backstage directory though.

craft_parts/executor/part_handler.py
lengau: Not directly - I'm just trying to ensure that the migration to stage and backstage works the same way so that when we do have uses for it in plugins we don't have weird issues related to plugins using backstage.

craft_parts/executor/part_handler.py
lengau: Doing it as the last step of build introduces a simultaneity issue, where parts that are supposed to be independent could end up using each other's backstage output based purely on the order in which craft-parts chooses to run the build. Doing it as part of the stage step makes the `after` keyword both necessary and work the same as if the files went into stage rather than backstage.

The following `parts.yaml` should fail (and does under this implementation):

```
parts:
  first:
    source: https://github.com/tomprogrammer/rust-ascii/archive/refs/tags/v1.1.0.tar.gz
    plugin: cargo-package
    cargo-package-cargo-command: /usr/bin/cargo
  second:
    plugin: rust
    source: https://github.com/tormol/encode_unicode/archive/refs/tags/v1.0.0.tar.gz
    rust-channel: none
```

However, migrating the files during the build step makes the build order of these two unrelated parts matter, and if `first` builds before `second`, it succeeds.

craft_parts/executor/part_handler.py
lengau: That's a good point. The intention was to make it the same as the migration to `stage` so I'll need to move this.

uv.lock
lengau: My experimental branch included a sync because I ran into some buggy behaviour (I don't remember what). I'll drop this change.

craft_parts/executor/executor.py
lengau: That was leftover from the experimental branch work. Removed.

craft_parts/parts.py
tigarmo: why? You're passing "*" here:

```
            backstage_files, backstage_dirs = filesets.migratable_filesets(
                Fileset(["(default)/*"], name="backstage"),
                str(self._part.part_export_dir),
                "default",
            )
```

This is the model, not the state

craft_parts/ctl.py
```diff
@@ -64,21 +65,20 @@ def _client(cmd: str, args: list[str]) -> str | None:
     :raise RuntimeError: If the command is invalid.
     """
     try:
-        call_fifo = os.environ["PARTS_CALL_FIFO"]
-        feedback_fifo = os.environ["PARTS_FEEDBACK_FIFO"]
+        ctl_socket_path = os.environ["PARTS_CTL_SOCKET"]
```
tigarmo, line 68: what are these unrelated changes? Is `feature/offline-builds` up to date?

lengau: It looks like they came from the merge I did. I've done a rebase and they disappeared. Not sure what happened.

craft_parts/executor/part_handler.py
lengau: Removed

tigarmo (approved):
🚀 🚀

sergiusens (approved):
:fire:

craft_parts/executor/step_handler.py
```diff
@@ -176,6 +184,24 @@ def pkgconfig_fixup(file_path: str) -> None:
                 prefix_trim=self._part.part_install_dir,
             )
 
+        def backstage_pkgconfig_fixup(
```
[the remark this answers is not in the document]
lengau: Removed :-)

mr-cal (approved):

PR #1002, opened 2025-02-03
feat: Allow slashes in partition names

### LINKED ISSUE #999: Allow slashes in partition names
### What needs to get done

1. Update validation to allow slashes
2. Update and refactor `validate_partition_names()` to detect new possible conflicts.
3. Update `HAS_PARTITION_REGEX` and the partition identification logic in `utils/path_utils.py`.

Example: `test/foo/bar` and `test/foo-bar` should conflicts.

### Why it needs to get done

To allow slashes in partition names and enable imagecraft creating partitions named `volume/pc/efi`. See https://docs.google.com/document/d/12LxcUIUEoXvzK619CvK9eHU_8n_rEKnLPEC4clsriXA/edit?pli=1&tab=t.0#heading=h.fff0e8cg9w1x

upils:

-----

Fixes #999 
CRAFT-4064


mr-cal (approved):
Very well written for a complex algorithm, nice work

craft_parts/utils/partition_utils.py
```diff
@@ -118,58 +115,123 @@ def _validate_partition_naming_convention(partitions: Sequence[str]) -> None:
...
+    conflictings_partitions = detect_conflicts(partitions)
+
+    if not conflictings_partitions:
+        return
+
+    filtered_conflictings_partitions = unique_conflicts(conflictings_partitions)
+
+    # Raise the full list of collisions
+    lines: list[str] = []
+    for conflictings in filtered_conflictings_partitions.values():
```
cmatsuoka, line 142: Shouldn't that be `conflicting` without the `s`? Maybe `for conflict in filtered_conflicting_partitions`?

craft_parts/utils/partition_utils.py
```diff
@@ -118,58 +115,123 @@ def _validate_partition_naming_convention(partitions: Sequence[str]) -> None:
...
+        conflictings_list = list(conflictings)
+        conflictings_list.sort()
+        lines.append(f"- {str(conflictings_list)[1:-1]}")
+
+    msg = "\n".join(lines)
+
+    raise errors.FeatureError(
+        message=f"Following partitions conflicts:\n{msg}",
+        details="Hyphens and slashes are converted to underscores to associate partitions names with environment variables. 'foo-bar' and 'foo/bar' would result in environment variable FOO_BAR.",
```
cmatsuoka, line 151: I think we're using the name "dash" for - elsewhere (@mr-cal, can you confirm?) 
Also change the message to something like "There are partition name conflicts"?

craft_parts/utils/partition_utils.py
```diff
@@ -118,58 +115,123 @@ def _validate_partition_naming_convention(partitions: Sequence[str]) -> None:
...
+
+def detect_conflicts(partitions: Sequence[str]) -> dict[str, set[str]]:
+    """Detect and return every conflicts between partitions.
 
-    For example, `foo` conflicts in ['default', 'foo', 'foo/bar'].
-    Assumes partition names are valid.
+    Rules:
+      1: partition name must not conflict with namespace (or sub-namespace) names
+        - `foo` and 'foo/bar' conflicts
+        - `foo/bar` and 'foo/bar/baz' conflicts
```
cmatsuoka, line 161: These should be `conflict`.

craft_parts/utils/partition_utils.py
```diff
@@ -118,58 +115,123 @@ def _validate_partition_naming_convention(partitions: Sequence[str]) -> None:
...
+      2: environment variables derived from partition names must not conflicts
+        - `foo/bar-baz` and `foo/bar/baz` conflicts (would be converted to FOO_BAR_BAZ)
+        - `foo/bar/baz-qux` and `foo/bar-baz/qux` conflicts (would be converted to FOO_BAR_BAZ_QUX)
```
cmatsuoka, line 164: These three should also be `conflict`.

craft_parts/utils/partition_utils.py
```diff
@@ -118,58 +115,123 @@ def _validate_partition_naming_convention(partitions: Sequence[str]) -> None:
...
 
-    :raises FeatureError: for namespace conflicts
     """
-    namespaced_partitions: set[str] = set()
-    regular_partitions: set[str] = set()
+    conflictings_partitions: dict[str, set[str]] = defaultdict(set)
```
cmatsuoka, line 167: And this one should be `conflicting`.

craft_parts/utils/partition_utils.py
```diff
@@ -118,58 +115,123 @@ def _validate_partition_naming_convention(partitions: Sequence[str]) -> None:
...
+
+    # Rule #1: Detect conflicts between partitions and namespaces partitions
+    for candidate_partition in partitions:
+        for partition in partitions:
```
cmatsuoka, line 171: I think `itertool.combinations()` can help here.

craft_parts/utils/partition_utils.py
```diff
@@ -118,58 +115,123 @@ def _validate_partition_naming_convention(partitions: Sequence[str]) -> None:
...
+            if candidate_partition == partition:
+                continue
+            if namespace_conflicts(candidate_partition, partition):
+                if not conflictings_partitions[partition]:
```
cmatsuoka, line 175: This can be `conflicting_partitions`. Also if this is a set you can just add it without testing, the code will be slightly simpler.

craft_parts/utils/partition_utils.py
```diff
@@ -118,58 +115,123 @@ def _validate_partition_naming_convention(partitions: Sequence[str]) -> None:
...
-    # sort partitions
     for partition in partitions:
-        if _is_valid_partition_name(partition):
-            regular_partitions.add(partition)
+        underscored = partition.translate({ord("-"): "_", ord("/"): "_"})
+
+        for known_partition, known_underscored in underscored_partitions.items():
+            if env_var_conflicts(underscored, known_underscored):
+                if not conflictings_partitions[known_partition]:
+                    conflictings_partitions[known_partition].add(known_partition)
```
cmatsuoka, line 188: Same here.

craft_parts/utils/partition_utils.py
```diff
@@ -118,58 +115,123 @@ def _validate_partition_naming_convention(partitions: Sequence[str]) -> None:
...
+                conflictings_partitions[known_partition].add(partition)
+
+        underscored_partitions[partition] = underscored
+
+    return conflictings_partitions
+
+
+def split_and_detect(a: str, b: str, separator: str) -> bool:
+    """Split candidates with the separator and check if the shorter one is a subset of the other."""
```
cmatsuoka, line 197: If this is performed on a single separator, can we just check if one string `startswith` the other?

craft_parts/utils/partition_utils.py
```diff
@@ -118,58 +115,123 @@ def _validate_partition_naming_convention(partitions: Sequence[str]) -> None:
...
+    return split_and_detect(a, b, "_")
+
+
+def unique_conflicts(
+    conflicting_partitions: dict[str, set[str]],
+) -> dict[str, set[str]]:
+    """Remove subsets of other conflicts."""
+    filtered: dict[str, set[str]] = {}
+    for partition, to_keep in conflicting_partitions.items():
+        for p2, conflicts in conflicting_partitions.items():
```
cmatsuoka, line 223: I wonder if there's a simpler way to do this. Maybe would could just create a set of conflicting names as they are detected? I'm not sure what you are returning here, a more descriptive docstring including the return value would help.

cmatsuoka (changes requested):
Did a pass, I think we can fix a couple of typos (maybe result of some global find-and-replace operation), add some comments, simplify the implementation a bit and check if we could change any algorithms.

craft_parts/utils/partition_utils.py
upils: I reworked this section based on your suggestion. I have now added a proper `register_conflict()` function to avoid registering duplicates (and so avoid having to filter them later).
It now looks a lot cleaner and easier to read.

craft_parts/utils/partition_utils.py
upils: Yep, thanks! The code is now also checking both "rules" at the same time. 
In some cases translating the partition names may be unnecessary but since we will never deal with many partitions I think that is fine.

craft_parts/utils/partition_utils.py
upils: That was the initial implementation but I failed to make it work because:

- `foo` and `foo/bar` must conflict (the previous behavior)
- `foo/bar-baz` and `foo/bar/baz` must now also conflict
- `foo/bar` and `foo/bar/baz` must now also conflict
- `foo/bar` and `foo/barbaz`  must **NOT** conflict (here using `startswith` would detect a conflict)
 
 I just simplified and fixed it though. Let me know what you think.

craft_parts/utils/partition_utils.py
mr-cal: I think "hyphen" is used in most places across our repos.

mr-cal (approved):
Even simpler, nice work.

cmatsuoka (approved):
Nice, thank you!

mr-cal:
@upils - @bepri landed his PR in #1003, so you are welcome to update the changelog with your change and the release date.  Once you land this PR, you can push a signed tag and the release will happen automatically in CI.

PR #1003, opened 2025-02-03
fix: don't require a specific python interpreter for uv

### LINKED ISSUE #991: uv plugin fails if _get_system_python_interpreter returns None
### Bug Description

In the base abstract definition of `_get_system_python_interpreter()` below, it is possible to return `None`. https://github.com/canonical/craft-parts/blob/8907c070500729da894f3ffa43306abc9517e91a/craft_parts/plugins/base.py#L237

However, this line below attempts to stringify the return value of `_get_system_python_interpreter()`, which will always break if set to `None`. https://github.com/canonical/craft-parts/blob/8907c070500729da894f3ffa43306abc9517e91a/craft_parts/plugins/uv_plugin.py#L102 

### To Reproduce

Override `UvPlugin`'s `_get_system_python_interpreter` to return `None`, then attempt to build a project.

### part yaml

```shell

```

### Relevant log output

```shell
N/A
```

bepri:

-----

Closes #991.
CRAFT-4010


tests/integration/plugins/test_uv.py
```diff
@@ -148,9 +148,11 @@ def _should_remove_symlinks(self) -> bool:
     )
     actions = lf.plan(Step.PRIME)
 
-    with lf.action_executor() as ctx, pytest.raises(errors.PluginBuildError):
+    with lf.action_executor() as ctx, pytest.raises(errors.PluginBuildError) as exc:
         ctx.execute(actions)
 
+    assert b"No suitable Python interpreter found" in cast(bytes, exc.value.stderr)
```
bepri, line 154: This test was meant to prevent this bug, but since it was just raising a PluginBuildError anyways it was a false-positive pass. Now it checks that the failure reason is actually the one we want.

uv.lock
tigarmo, line 1: how come this file got modified?

tests/integration/plugins/test_uv.py
tigarmo: seems to be failing (either this or a related test): https://github.com/canonical/craft-parts/actions/runs/13121108288/job/36607013470?pr=1003

uv.lock
bepri: I assume something was out of date. It looks like the version of craft-parts itself was a dirty version of 2.2.1 in the old lockfile, so maybe this isn't a bad change to keep? I'll revert it if you disagree though.

tests/integration/plugins/test_uv.py
bepri: oops! that was indeed a different test, not one I expected to break. should be fixed now

bepri:
Rockcraft PR using this fix and removing its own fix for the problem: https://github.com/canonical/rockcraft/pull/803

uv.lock
tigarmo: the problem is not so much the dirty version, it's the unexpected change to the lockfile at all. What is the command that is updating this?

bepri: iirc, all I did was run `uv sync --extra apt-oracular` before starting this. I'd say it was just because I didn't include `--frozen` (I'm a bit too accustomed to the starbase makefile doing that work for me...). I understand what you're saying though, there were no dependency changes so the lockfile shouldn't change either. I'll revert it

craft_parts/plugins/uv_plugin.py
```diff
@@ -98,8 +98,13 @@ def _get_rewrite_shebangs_commands(self) -> list[str]:
...
 
     def _get_create_venv_commands(self) -> list[str]:
+        # Explicitly request a Python version if provided by the plugin, otherwise use the global
+        # parts interpreter.
+        python_ver = (
+            self._get_system_python_interpreter() or "${PARTS_PYTHON_INTERPRETER}"
+        )
         return [
-            f'uv venv --relocatable --allow-existing --python "{self._get_system_python_interpreter()}" "{self._get_venv_directory()}"',
+            f'uv venv --relocatable --allow-existing --python {python_ver} "{self._get_venv_directory()}"',
```
lengau, line 107: I don't think this is doing what you intend it to do.

image

You probably want the equivalent of this if you're using the interpreter environment variable:

```
$ uv venv -p $(python3 --version | cut -d' ' -f2)
Using CPython 3.13.1 interpreter at: /usr/bin/python3.13
Creating virtual environment at: .venv
Activate with: source .venv/bin/activate
```

lengau: Actually probably what you want is to always ensure you're using the full path...

```suggestion
            f'uv venv --relocatable --allow-existing --python $(which {python_ver}) "{self._get_venv_directory()}"',
```

lengau (approved):
Nice fix!

tigarmo (approved):
Thanks! Please wait for the rockcraft PR to pass before merging this

tigarmo:
@bepri :shipit: !

PR #1004, opened 2025-02-04
fix: [PROTOTYPE] add option to use symlinks with the npm plugin

bepri:

-----

### DIFF
--- a/craft_parts/plugins/base.py
+++ b/craft_parts/plugins/base.py
@@ -25,8 +25,10 @@ from copy import deepcopy
 from typing import TYPE_CHECKING
 
 from overrides import override
+from packaging.version import Version
 
 from craft_parts.actions import ActionProperties
+from craft_parts.utils.os_utils import OsRelease
 
 from .properties import PluginProperties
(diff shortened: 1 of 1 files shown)


craft_parts/plugins/npm_plugin.py
```diff
@@ -301,11 +314,12 @@ def get_build_commands(self) -> list[str]:
             cmd += [
                 dedent(
                     f"""\
-                tar -xzf "{self._node_binary_path}" -C "${{CRAFT_PART_INSTALL}}/" \
+                tar -hxzf "{self._node_binary_path}" -C "${{CRAFT_PART_INSTALL}}/" \
```
bepri, line 317: `tar -h` avoids overwriting symlinks at the location, instead following them when unpacking a tarball

tests/integration/plugins/test_npm.py
bepri, line 1: Much of the diff in this file was a refactor to deduplicate code into a fixture, the tests themselves are unchanged

PR #1024, opened 2025-02-25
feat: allow short commit hashes

### LINKED ISSUE #1008: Support short git hashes in `source-commit`
### What needs to get done

Git does not support fetching commits via the short hash (as opposed to the full hash). One can work around this by doing something like:

```yaml
override-pull: |
  craftctl default
  git checkout <short-hash>
```

... but this is cumbersome and something that craft-parts could do by itself.

### Why it needs to get done

Short hashes are very common when making references to commits in Git.

mr-cal:

-----

Previously, `source-commit` could only accept full length (40 character) hashes. Now, `source-commit` can accept short hashes.

I recommend reviewing by commit because the first commit is a refactor that moves a bunch of tests.

Fixes #1008
(CRAFT-4127)

### DIFF
--- a/craft_parts/sources/git_source.py
+++ b/craft_parts/sources/git_source.py
@@ -174,11 +174,44 @@ class GitSource(SourceHandler):
     ) -> None:
         super().__init__(source, part_src_dir, **kwargs)
 
+    def _expand_commit(self, commit: str) -> str:
+        """Expand a commit hash to full length."""
+        # short-circuit if the commit is already full length
+        if len(commit) == 40:  # noqa: PLR2004 (magic-value)
+            return commit
+
+        try:
+            # 'git rev-parse' expands non-ambiguous commits to full length (40 characters).
+            return subprocess.check_output(
+                [
+                    get_git_command(),
+                    "-C",
+                    str(self.part_src_dir),
+                    "rev-parse",
+                    commit,
+                ],
+                text=True,
+            ).strip()
+        except subprocess.CalledProcessError as err:
+            # 'rev-parse' will fail if the commit is out of the source-depth range, so give a helpful error
+            if self.source_depth:
+                raise errors.VCSError(
+                    message=f"Failed to parse commit {commit!r}.",
+                    resolution=(
+                        "Ensure 'source-commit' is correct, provide a full-length (40 character) commit, "
+                        "or remove the 'source-depth' key from the part."
+                    ),
+                ) from err
+            raise errors.VCSError(
+                message=f"Failed to parse commit {commit!r}.",
+                resolution="Ensure 'source-commit' is correct or provide a full-length (40 character) commit.",
+            ) from err
+
     def _fetch_origin_commit(self) -> None:
         """Fetch from origin, using source-commit if defined."""
         command = [get_git_command(), "-C", str(self.part_src_dir), "fetch", "origin"]
         if self.source_commit:
-            command.append(self.source_commit)
+            command.append(self._expand_commit(self.source_commit))
 
         self._run(command)
 
@@ -211,8 +244,8 @@ class GitSource(SourceHandler):
         elif self.source_tag:
             refspec = "refs/tags/" + self.source_tag
         elif self.source_commit:
-            refspec = self.source_commit
             self._fetch_origin_commit()
+            refspec = self._expand_commit(self.source_commit)
         else:
             refspec = "refs/remotes/origin/" + self._get_current_branch()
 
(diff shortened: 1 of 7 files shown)


craft_parts/sources/git_source.py
```diff
@@ -174,11 +174,25 @@ def __init__(  # pylint: disable=too-many-arguments
     ) -> None:
         super().__init__(source, part_src_dir, **kwargs)
 
+    def _expand_commit(self, commit: str) -> str:
+        """Expand a commit hash to full length."""
+        # 'git rev-parse' expands non-ambiguous commits to full length (40 characters).
+        # Full length commits are unchanged.
```
tigarmo, line 180: Maybe we should early-return if `len(commit) == 40`?

tests/unit/sources/test_git_source.py
```diff
@@ -156,16 +156,25 @@ def test_pull_tag(self, fake_run, new_dir):
...
 
-    def test_pull_commit(self, fake_run, new_dir):
+    def test_pull_commit(self, fake_check_output, fake_run, new_dir):
+        fake_check_output.return_value = "2514f9533ec9b45d07883e10a561b248497a8e3c"
         git = GitSource(
             "git://my-source",
             Path("source_dir"),
             cache_dir=new_dir,
-            source_commit="2514f9533ec9b45d07883e10a561b248497a8e3c",
+            source_commit="2514f9533e",
```
tigarmo, line 165: these feel like *new* test cases, in addition to the existing ones

craft_parts/sources/git_source.py
```diff
@@ -174,11 +174,25 @@ def __init__(  # pylint: disable=too-many-arguments
...
+        return self._run_output(
+            [
+                get_git_command(),
+                "-C",
+                str(self.part_src_dir),
+                "rev-parse",
```
tigarmo, line 186: A consideration, possibly for near-future work; if `source-depth` is non-zero, this call will probably fail:

```
❯ git rev-parse cd483f3
cd483f3
fatal: ambiguous argument 'cd483f3': unknown revision or path not in the working tree.
```

Should we handle this in any particular way?

tigarmo: In particular, the case that will probably be common is trying to shallow-clone to a short-hash

tests/unit/sources/test_git_source.py
mr-cal: As-is, I decided not to add another test because the unit tests are fully mocked and it doesn't matter in the slightest what I pass to `source-commit`. I made this change to ensure the commands used the return value from `fake_check_output.return_value = "2514f9533ec9b45d07883e10a561b248497a8e3c"` and not what was passed to `source_commit`.

If I add an early return for full-length commits like you suggested, then I agree adding a new test would exercise additional code.

craft_parts/sources/git_source.py
tigarmo: I like this idea, it also addresses the case where there's a typo on the short hash
but I'm also ok with doing this a bit later if you prefer

tigarmo (approved):
thanks!

dariuszd21 (approved):
Looks good to me, thanks!

tests/integration/sources/test_git_source.py
```diff
@@ -0,0 +1,448 @@
...
+        self.add_file("test.txt", "Howdy, Partner!", "updated test.txt")
+        _call(["git", "push", "-f", str(remote)])
+
+        # go back to the working tree and pull the new commit
+        os.chdir(working_tree)
+        git.pull()
+
+        # assert we actually pulled the commit
+        with open(Path(working_tree / "test.txt")) as file:
+            assert file.read() == "Howdy, Partner!"
```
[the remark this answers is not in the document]
dariuszd21: I can see that most of those code is moved from `unit` to `integration` so you may ignore my suggestions

tests/unit/sources/test_git_source.py
```diff
@@ -208,46 +156,57 @@ def test_pull_tag(self, fake_run, new_dir):
...
-                        "source_dir",
-                        "fetch",
-                        "origin",
-                        "2514f9533ec9b45d07883e10a561b248497a8e3c",
-                    ]
-                ),
+                    ["git", "-C", "source_dir", "rev-parse", short_commit], text=True
+                )
+            ]
+            * 2
```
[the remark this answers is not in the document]
mr-cal: Once to call `git fetch origin commit` and once to call `git checkout commit`.

I considered:
* using `lrucache` (probably safe)
* refactoring so it only gets called once (I'm not super comfortable refactoring in this PR, so I left it out of scope)

craft_parts/sources/git_source.py
mr-cal: Added :)

mr-cal:
Force pushing for a cutesy little rebase

PR #1031, opened 2025-03-05
docs: add new how-to for parts

bepri:

-----

This PR takes [this Snapcraft documentation page](https://snapcraft.io/docs/adding-parts) and upstream and generalizes it as much as possible.

App owners - please pay attention to the labels in the source rST file. The hope is that this page can be slice 'n' diced to include any extra info perhaps specific to your app.

CRAFT-4260


docs/common/craft-parts/how-to/use_parts.rst
```diff
@@ -0,0 +1,115 @@
+.. _how-to-use-parts:
+
+Using parts in a project file
+=============================
+
+When packing, :ref:`parts <parts>` are used to describe your package, where its various
+components can be found, its build and run-time requirements, and its dependencies.
+Consequently, all package will have one or more parts.
```
lengau, line 8:
```suggestion
Consequently, all packages will have one or more parts.
```

docs/common/craft-parts/how-to/use_parts.rst
```diff
@@ -0,0 +1,115 @@
...
+
+.. _how-to-use-parts_details:
+
+Parts are purposefully flexible to allow for varied and disparate sources. At its
+simplest, a part will locate a project's source code and invoke a
+:ref:`plugin <plugins>` to build the application in the environment. However, a part
+can just as easily be used to source and extract a binary executable from an RPM file.
+They can also download tagged code from a remote repository, pull in dependencies,
```
lengau, line 16:
```suggestion
It can also download tagged code from a remote repository, pull in dependencies,
```

docs/common/craft-parts/how-to/use_parts.rst
```diff
@@ -0,0 +1,115 @@
...
+A simple part that just packages a simple "hello world" shell script might look like
+this:
+
+.. code:: yaml
+
+    parts:
+        my-part:
+            plugin: dump
+            source: hello.sh
+            source-type: file
```
lengau, line 34:
```suggestion
      my-part:
        plugin: dump
        source: hello.sh
        source-type: file
```

docs/common/craft-parts/how-to/use_parts.rst
```diff
@@ -0,0 +1,115 @@
...
+
+A part starts with an arbitrary name, in this case, ``my-part``. They always will go
```
lengau, line 36:
```suggestion
A part starts with an arbitrary name, in this case, ``my-part``. They go
```

docs/common/craft-parts/how-to/use_parts.rst
```diff
@@ -0,0 +1,115 @@
...
+under the top-level ``parts`` key, which can contain as many parts as your application
```
lengau, line 37:
```suggestion
under the top-level ``parts`` key, which can contain as many parts as your package
```

docs/common/craft-parts/how-to/use_parts.rst
```diff
@@ -0,0 +1,115 @@
...
+invoke an entire build system or create a workspace for offline builds.
+
+.. _how-to-use-parts_keys-intro:
+
+Common keywords
+---------------
+
+A typical part may also commonly make use of the following keys.
+
+.. _how-to-use-parts_keys-list:
```
lengau, line 56: We already have a description of parts keys in a reference doc, so probably better to link there.

lengau (changes requested):

docs/common/craft-parts/how-to/use_parts.rst
bepri: I probably went about this the wrong way, but I was aiming to avoid that since I think I remember @medubelko decreeing that we shouldn't link between common docs. Am I remembering wrong?

tigarmo: I vaguely remember this conversation and I *think* the issue is that we have to make sure that the page being referenced will indeed be part of the final docset, otherwise it's a broken reference.
In this case I think it should be fine - I can't think of a case where a project would want this part-oriented how-to but *not* want the part key reference

docs/common/craft-parts/how-to/use_parts.rst
```diff
@@ -0,0 +1,115 @@
...
+simplest, a part will locate a project's source code and invoke a
+:ref:`plugin <plugins>` to build the application in the environment. However, a part
+can just as easily be used to source and extract a binary executable from an RPM file.
+It can also download tagged code from a remote repository, pull in dependencies,
+define a build order, or completely override any stage of the
+:ref:`lifecycle <lifecycle>`.
+
+.. _how-to-use-parts_defining:
+
+Defining a part
```
tigarmo, line 22: Having recently written a how-to, I'll share the feedback that I got that I think also applies to this page - a lot of the content here is more "reference" than "how-to". How-tos should be step-oriented, which you probably know as the paragraphs that start this section go in that direction and seem to be new (as far as I can tell).

In my case, I found it really helpful to do an outline of the sections/steps before-hand and getting an approval on that. Of course, @medubelko is the subject matter expert here and can (please) correct me if I'm wrong

docs/common/craft-parts/how-to/use_parts.rst
bepri: I remember that too, now that you say it. Thanks! I just swapped to linking to the parts reference instead.

medubelko: The prescription is that we must be careful. If as Tiago says we're utterly certain that these docs form an immutable bond and no downstream app will ever isolate them, then we can interlink them. If there's any doubt to such, we shouldn't.

medubelko (approved):

lengau (approved):

docs/common/craft-parts/how-to/use_parts.rst
```diff
@@ -0,0 +1,61 @@
...
+When packing, :ref:`parts <parts>` are used to describe your package, where its various
+components can be found, its build and run-time requirements, and its dependencies.
+Consequently, all packages will have one or more parts.
+
+.. _how-to-use-parts_details:
+
+Parts are purposefully flexible to allow for varied and disparate sources. At its
+simplest, a part will locate a project's source code and invoke a
+:ref:`plugin <plugins>` to build the application in the environment. However, you can
+just as easily use a part to to source and extract a binary executable from an RPM file.
```
tigarmo, line 15:
```suggestion
just as easily use a part to source and extract a binary executable from an RPM file.
```

tigarmo (approved):

PR #1032, opened 2025-03-06
feat: spring boot plugin

yanksyoon:

-----

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,7 +92,7 @@ test-ruff:
 
 .PHONY: test-integrations
 test-integrations: ## Run integration tests.
-	pytest tests/integration
+	pytest --basetemp .tmp tests/integration
 
 .PHONY: test-mypy
 test-mypy:
--- a/Makefile
+++ b/Makefile
@@ -92,7 +92,7 @@ test-ruff:
 
 .PHONY: test-integrations
 test-integrations: ## Run integration tests.
-	pytest --basetemp .tmp tests/integration
+	pytest tests/integration
 
 .PHONY: test-mypy
 test-mypy:
--- a/craft_parts/plugins/spring_boot_plugin.py
+++ b/craft_parts/plugins/spring_boot_plugin.py
@@ -183,4 +183,14 @@ class SpringBootPlugin(Plugin):
             )
         )
 
+        # Install tmp dir (unavailble in bare base)
+        commands.append(
+            textwrap.dedent(
+                """\
+                mkdir -p ${CRAFT_PRIME}/tmp
+                chown 584792 ${CRAFT_PRIME}/tmp
+                """
+            )
+        )
+
         return commands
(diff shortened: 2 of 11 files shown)


Makefile
```diff
@@ -92,7 +92,7 @@ test-ruff:
 
 .PHONY: test-integrations
 test-integrations: ## Run integration tests.
-	pytest tests/integration
+	pytest --basetemp .tmp tests/integration
```
yanksyoon, line 95: Changed `basetemp` dir to avoid using `sudo` to write files to tempdir. See https://docs.pytest.org/en/stable/how-to/tmp_path.html#temporary-directory-location-and-retention

craft_parts/plugins/spring_boot_plugin.py
```diff
@@ -0,0 +1,196 @@
...
+                fi
+                """
+            )
+        )
+        commands.append(
+            textwrap.dedent(
+                """\
+                if [[ $PROJECT_JAVA_VERSION -gt $SYSTEM_JAVA_MAJOR_VERSION ]]; then
+                    echo "Project requires Java version $PROJECT_JAVA_VERSION, but system Java \
+                version is $SYSTEM_JAVA_MAJOR_VERSION."
```
jdkandersson, line 152: Let's add a link explaining that the project requires a Java version which is later than is available in the system which means the app won't start

craft_parts/plugins/spring_boot_plugin.py
```diff
@@ -0,0 +1,196 @@
...
+                fi
+                """
+            )
+        )
+
+        # Install tmp dir (unavailable in bare base)
+        commands.append(
+            textwrap.dedent(
+                """\
+                mkdir -p ${CRAFT_PRIME}/tmp
```
jdkandersson, line 190: Let's check why this is required

tests/integration/plugins/test_spring_boot.py
```diff
@@ -0,0 +1,196 @@
...
+            deps-setup:
+                plugin: nil
+                build-packages:
+                - openjdk-17-jdk
+            my-part:
+                plugin: spring-boot
+                source: {source_location}
+                after: [deps-setup]
+                build-environment:
+                - JAVA_HOME: /usr/lib/jvm/java-17-openjdk-${{CRAFT_ARCH_BUILD_FOR}}
```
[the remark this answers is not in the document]
yanksyoon: @tigarmo Hello Tiago, may I know if this is possible here please? The test uses the system environment and hence many `java` executable exists - which the tooling uses to determine the Java version.
From what I understand, different plugins also depend on the Java (example Jlink) executable and I shouldn't quite manipulate the environment due to other test's dependencies.

yanksyoon:
@tigarmo @vpa1977 
Hello Tiago, Vladimir, may I know if I could get a review on these please? Thank you!
Please help focus more on the code aspects as the documentation is being worked on by @erincon

vpa1977:
I am not sure of the design - the plugin seems to be to do be doing many things at once and have to take into account multiple build system details. 
Maybe it is possible to refactor as:
 - fix to the maven plugin to prefer mvnw and add checks for the Java version
 - a gradle plugin (that supports wrapper at the moment) and the Java version (so that installed Java can run the wrapper).

yanksyoon:
Closing due to redesign of the plugins :)

PR #1033, opened 2025-03-06
fix(jlink): find all jar files in default behavior

### LINKED ISSUE #1027: jlink: default jar search sometimes fails to find all jars
### Bug Description

If `jlink-jars` is empty, the default behavior is that the plugin will use all jar files in CRAFT_STAGE to assemble the Java runtime. In some scenarios, only some jar files are found.

### To Reproduce

The simplest way that I've found to reproduce this is to have a CRAFT_STAGE dir that looks like this:

```
stage/
  |- file1.jar
  |- subdir/
  |  |- file2.jar
```

In this scenario, only `file1.jar` is found. I suspect this has to do with how the glob is done but I haven't really explored it.

### part yaml

```shell

```

### Relevant log output

```shell
n/a
```

bepri:

-----
Note: The actual fix was quoting the "*.jar" in the find command, which was being expanded by bash instead of find. It was moved into a function in the name of testing.

Closes #1027
CRAFT-4266.

### DIFF
--- a/craft_parts/plugins/jlink_plugin.py
+++ b/craft_parts/plugins/jlink_plugin.py
@@ -61,6 +61,15 @@ class JLinkPlugin(Plugin):
     properties_class = JLinkPluginProperties
     validator_class = JLinkPluginEnvironmentValidator
 
+    def _get_find_jars_commands(self) -> str:
+        """Return commands for finding all jarfiles."""
+        options = cast(JLinkPluginProperties, self._options)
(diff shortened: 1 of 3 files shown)


mr-cal (changes requested):
Can you add a changelog entry?

mr-cal (approved):
Thanks!

tigarmo (approved):

PR #1047, opened 2025-03-19
feat: new .NET plugin (v2)

mateusrodrigues:
- [X] Have you added an entry to the changelog (`docs/reference/changelog.rst`)?

-----

Hi folks!

This PR introduces the new .NET plugin in accordance with the FO205 specification (.NET Snapcraft Plugin). I’ve created it as a separate file and included unit tests, integration tests, documentation, and a changelog entry.

There are a few points I’d like to address before finalizing the PR:
- Since the old .NET plugin is only available on `core22`, should we adjust the [current Snapcraft registration logic](https://github.com/canonical/snapcraft/blob/7651d96085a7d715b06764e4e5dff391fb68b4b1/snapcraft/application.py#L139-L141) to register the old one and unregister the new one on `core22`, while keeping the new one registered at the [craft-parts level](https://github.com/canonical/craft-parts/blob/ac6e1d249ee2eeb035f5ed7832ed9c450617c698/craft_parts/plugins/plugins.py#L57)? This should simplify things for newer bases, e.g. `core24` and above, as the new plugin wouldn't need to be manually registered for each.
- Relates to the point above, but since the new plugin is not registered here, the integration tests fails because it only finds the old `dotnet` plugin. Any suggestions on how we can work around this?

**Note:** This is being submitted as a draft PR until the spec is fully approved. In the meantime, we can discuss the points above.


tests/integration/plugins/test_dotnet_v2.py
```diff
@@ -0,0 +1,136 @@
...
+            dotnet-self-contained: true
+            build-environment:
+              - PATH: $CRAFT_STAGE/sdk:$PATH
+            after: [dotnet-deps]
+          dotnet-deps:
+            plugin: dump
+            source: https://download.visualstudio.microsoft.com/download/pr/d2abdb4c-a96e-4123-9351-e4dd2ea20905/e8010ae2688786ffc1ebca4ebb52f41b/dotnet-sdk-8.0.406-linux-x64.tar.gz
+            source-checksum: sha512/d6fdcfebd0df46959f7857cfb3beac7de6c8843515ece28b24802765fd9cfb6c7e9701b320134cb4907322937ab89cae914ddc21bf48b9b6313e9a9af5c1f24a
+            override-build: |
+              # TODO: find out why this is a problem with "organize".XS
```
mr-cal, line 44: What's the problem?

mr-cal (commented):
Thanks, this is looking good. I've next week so we can discuss the timing, expectations, and strategy for managing multiple .net plugins.

I'm interested in a snapcraft PR that integrates this so we can build snaps and see how the documentation looks.  Your first suggestion to register the plugin [here](https://github.com/canonical/snapcraft/blob/7651d96085a7d715b06764e4e5dff391fb68b4b1/snapcraft/application.py#L139-L141) should work for a draft, but we may choose an alternative approach when we meet next week.

tests/integration/plugins/test_dotnet_v2.py
mateusrodrigues: I'm afraid this was a left-over from the original .NET plugin integration tests. I copied over the same [parts definition](https://github.com/canonical/craft-parts/blob/main/tests/integration/plugins/test_dotnet.py#L44) from there to create the tests for the new plugin.

docs/common/craft-parts/reference/plugins/dotnet_plugin.rst
```diff
@@ -5,6 +5,10 @@
 
 The ``dotnet`` plugin builds .NET projects using the ``dotnet`` tool.
 
+.. warning::
```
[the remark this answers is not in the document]
mateusrodrigues: Should I remove both that warning and the [note](https://github.com/mateusrodrigues/craft-parts/blob/edc951f6e6ec8bf1069ac430b1ec4d920adfeeb1/docs/common/craft-parts/reference/plugins/dotnet_v2_plugin.rst?plain=1#L8-L10) in the v2 documentation then?

mateusrodrigues:
@mr-cal sounds good, thanks for your review! I've committed your suggestions and adjusted the tests accordingly.

tests/integration/plugins/test_dotnet_v2.py
mr-cal: I see.  Since this isn't for the dotnet plugin itself, I'm fine with that for now. 

One day it would be nice to try using `organize` and see if it's still failing.

mr-cal (changes requested):

docs/common/craft-parts/reference/plugins/dotnet_plugin.rst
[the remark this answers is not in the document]
mateusrodrigues: Ok, I updated the notes accordingly.

mateusrodrigues:
I just now realized that, per the plugin spec, the `dotnet restore` command should be part of the `pull` step, not `build`. I have updated the plugin and the unit tests to reflect this.

cc @mr-cal

craft_parts/plugins/dotnet_v2_plugin.py
```diff
@@ -0,0 +1,401 @@
...
+        """Return a list of commands to run during the build step."""
+        options = cast(DotnetV2PluginProperties, self._options)
+
+        build_for = self._part_info.project_info.arch_build_for
+        dotnet_rid = _DEBIAN_ARCH_TO_DOTNET_RID.get(build_for)
+
+        if not dotnet_rid:
+            raise ValueError(
+                f"Unsupported architecture {build_for!r}. Supported architectures are {humanize_list(_DEBIAN_ARCH_TO_DOTNET_RID.keys(), 'and')}."
+            )
```
mr-cal, line 296: pedantic: This is the third occurrence of this, so the rule of 3 kicks in.  You could dedupe with something like:

```python
def _get_rid(arch: str) -> str:
   """Get the dotnet rid for an arch.

   :raises ValueError: If the architecture is unsupported.
   """
  if not dotnet_rid:
    raise ValueError(
      f"Unsupported architecture {build_for!r}. Supported architectures are {humanize_list(_DEBIAN_ARCH_TO_DOTNET_RID.keys(), 'and')}."
    )
```

mr-cal (approved):
Nice work!

I created a [feature branch](https://github.com/canonical/snapcraft/tree/refs/heads/feature/dotnet) in snapcraft. It gets [published](https://github.com/canonical/snapcraft/actions/runs/14641910677) to the `latest/edge/dotnet` channel for easy testing.

mateusrodrigues:
> I just now realized that, per the plugin spec, the `dotnet restore` command should be part of the `pull` step, not `build`. I have updated the plugin and the unit tests to reflect this.

There is currently an issue with this. The application part now needs the .NET SDK ready earlier, i.e. at the moment it's pulling the source.

Now, considering the scenario where the SDK is provided by a `dotnet-deps` part, `snapcraft` will pull *all* parts first before it gets around to building them (according to my tests), which means that by the time the application is pulled, `dotnet-deps` will not have staged the SDK yet and `dotnet restore` will then fail with "command not found", even if the app's part has `after: [dotnet-deps]`.

This might be the case to move `dotnet restore` back to the build step.

mateusrodrigues:
Ok, I have moved `dotnet restore` back to the `build` step and fixed the tests accordingly. I have also rebased this branch to fix a merge conflict.

Can we refresh the build available in the feature branch so I can run some more tests?

mr-cal:
@mateusrodrigues, it should be available soon: https://github.com/canonical/snapcraft/actions/runs/14731042368

mr-cal:
Plan from in-person discussion:
- @mateusrodrigues to try changing the temp dir to fix CI
  - If this doesn't work, @mr-cal will figure out another solution. (update: @lengau is already trying this)
- @mr-cal to land and release craft-parts
- @mr-cal to update craft-parts in snapcraft's `feature/dotnet` branch
- @mateusrodrigues to add a spread test for `feature/dotnet` that builds a core24 dotnet snap
- @mr-cal to add to the next monthly release of snapcraft

lengau:
Update to @mr-cal 's update: this is going to take a lot to check why this causes so many things to fail. This change may not be feasible for this PR.

mateusrodrigues:
The linter error seems unrelated, but I'm able to reproduce it locally.
```
(how-to/source-handler: line   52) broken    https://rsync.samba.org/ - 403 Client Error: Forbidden for url: https://rsync.samba.org/
```
Interestingly, the website seems to be reachable.

lengau (approved):

medubelko:
> @medubelko what can we do about this linkcheck issue?

I think more than one component in our infra has trouble with 403s. They're not necessarily bad links. I understand that as of this month the web team is ignoring them across the board in their validators.

Per the docs there's no built-in Sphinx config to ignore 403 errors. Best we can do is [add it to `linkcheck_ignore`](https://github.com/canonical/craft-parts/blob/4359cf3ff6b1c0caeabfba654b4cc64c3aba9b19/docs/conf.py#L73). Not ideal, but works for this PR.

tests/integration/plugins/test_dotnet_v2.py
```diff
@@ -0,0 +1,140 @@
...
+        @override
+        def _execute(self, cmd: str) -> str:
+            if cmd == "dotnet --version" and not self.__already_run:
+                self.__class__.__already_run = True
+                raise subprocess.CalledProcessError(127, cmd)
+            return super()._execute(cmd)
+
+    mocker.patch.object(
+        dotnet_v2_plugin.DotnetV2Plugin, "validator_class", FailSpecificCmdValidator
+    )
```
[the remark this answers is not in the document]
mateusrodrigues: Done, looks much cleaner now.

docs/common/craft-parts/craft-parts.wordlist.txt
```diff
@@ -65,6 +65,7 @@ DotPluginEnvironmentValidator
 Dotnet
 DotnetPlugin
 DotnetPluginProperties
+DotnetV
```
medubelko, line 68: There are several words added here that are:

- In source code
- Marked up as code in the docs (\`\`example\`\`)

The spell check shouldn't be picking these up, so we can safely remove them from this file.

docs/common/craft-parts/reference/plugins/dotnet_plugin.rst
```diff
@@ -5,6 +5,9 @@
...
+    This plugin has been superseded by :ref:`.NET plugin (v2) <craft_parts_dotnet_v2_plugin>`.
```
medubelko, line 9:
```suggestion
.. admonition:: Legacy plugin

    This plugin is superseded by the :ref:`craft_parts_dotnet_v2_plugin`.
```

docs/common/craft-parts/reference/plugins/dotnet_v2_plugin.rst
```diff
@@ -0,0 +1,231 @@
+.. _craft_parts_dotnet_v2_plugin:
+
+.NET plugin (v2)
+================
+
+The ``dotnet`` plugin builds .NET projects using the ``dotnet`` tool.
+
+.. note::
+    This plugin supersedes :ref:`.NET plugin <craft_parts_dotnet_plugin>`.
```
medubelko, line 9: Taking the opportunity here to improve on the intro:

```suggestion
The .NET plugin (v2) builds .NET projects using the `dotnet
<https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet>`_ tool. It's the successor
to the :ref:`craft_parts_dotnet_plugin`.
```

docs/common/craft-parts/reference/plugins/dotnet_v2_plugin.rst
```diff
@@ -0,0 +1,231 @@
...
+
+Keywords
+--------
+
+In addition to the common :ref:`plugin <part-properties-plugin>` and
+:ref:`sources <part-properties-sources>` keywords, this plugin provides the following
+plugin-specific keywords:
+
+.. _global_flags:
```
medubelko, line 18: Sphinx references are global, so we need to make new ones bespoke to this document.

```suggestion
.. _craft_parts_dotnet_v2_plugin-global-flags:
```

docs/common/craft-parts/reference/plugins/dotnet_v2_plugin.rst
```diff
@@ -0,0 +1,231 @@
...
+
+**Default:** ``"Release"``
+
+The .NET build configuration to use. Possible values are ``"Debug"`` and ``"Release"``.
+
+dotnet-project
+^^^^^^^^^^^^^^
+**Type:** string
+
+**Default:** ``None``
```
medubelko, line 35: Should this be `"None"`? Or are we saying it's unset?

docs/common/craft-parts/reference/plugins/dotnet_v2_plugin.rst
```diff
@@ -0,0 +1,231 @@
...
+
+Path to the solution or project file to build relative to the root of the snap source.
+If a project or solution file is not specified, MSBuild will search t he root of the snap
+source for a file that has an extension that ends in either ``proj`` or ``sln``.
```
medubelko, line 39:
```suggestion
The path to the solution or project file to build, relative to the root of the snap
source. If a path isn't specified, MSBuild will search the root of the source for a file
with the ``.proj`` or ``.sln`` extension.
```

docs/common/craft-parts/reference/plugins/dotnet_v2_plugin.rst
```diff
@@ -0,0 +1,231 @@
...
+
+dotnet-properties
+^^^^^^^^^^^^^^^^^
+**Type:** dict of strings to strings
+
+**Default:** Empty
```
medubelko, line 45: For configurations, "Unset" is typically the best word to indicate that the setting isn't used. It avoids the confusion around null/nil/noop.

docs/common/craft-parts/reference/plugins/dotnet_v2_plugin.rst
```diff
@@ -0,0 +1,231 @@
...
+
+A list of MSBuild properties to be appended to the restore, build, and publish commands
+in the format of ``-p:<Key>=<Value>``.
+
+.. _dotnet_self_contained:
```
medubelko, line 50: Let's apply the same uniquifier to all of the new references.

docs/common/craft-parts/reference/plugins/dotnet_v2_plugin.rst
```diff
@@ -0,0 +1,231 @@
...
+
+.. _dotnet_version:
+
+dotnet-version
+^^^^^^^^^^^^^^
+**Type:** string
+
+**Default:** ``None``
+
+Sets the .NET version to build the project with. By setting this property, the plugin
```
medubelko, line 86:
```suggestion
Sets the .NET version to build the project with. By setting this key, the plugin
```

When we mean the named YAML key, let's use "key". This is a separate concern from the MSBuild properties.

docs/reference/changelog.rst
```diff
@@ -62,7 +62,8 @@ New features:
 
 - Previously, ``source-commit`` could only accept full length (40 character)
   hashes. Now, ``source-commit`` can accept short hashes.
-- Allow usage of the overlay and partitions features simultaneously. 
+- Allow usage of the overlay and partitions features simultaneously.
+- New :ref:`.NET plugin (v2) <craft_parts_dotnet_v2_plugin>`.
```
medubelko, line 66:
```suggestion
- :ref:`craft_parts_dotnet_v2_plugin`.
```

docs/common/craft-parts/reference/plugins/dotnet_v2_plugin.rst
```diff
@@ -0,0 +1,231 @@
...
+stage the ``dotnet`` executable to be used in the build step. This can be useful,
+for example, in cases where a specific, unreleased version of ``dotnet`` is desired
+but unavailable as a snap or an Ubuntu package.
+
+.. note::
+    This plugin will validate the presence of .NET by running ``dotnet --version``.
+    Therefore, it is assumed that the ``dotnet`` executable is visible in the PATH.
+    To achieve that, make sure to append the location of the staged .NET SDK from
+    ``dotnet-deps`` to the PATH using the :ref:`build-environment <build_environment>`
+    property.
```
medubelko, line 188:
```suggestion
This plugin will validate the presence of .NET by running ``dotnet --version``.
Therefore, it assumes that the dotnet executable is visible in the PATH. To
achieve that, make sure to append the location of the staged .NET SDK from
``dotnet-deps`` to the PATH using the :ref:`build-environment <build_environment>`
property.
```

We generally don't need callouts. :)

docs/common/craft-parts/reference/plugins/dotnet_v2_plugin.rst
```diff
@@ -0,0 +1,231 @@
...
+
+A list of MSBuild properties to be appended to the publish command in the format of
+``-p:<Key>=<Value>``.
+
+.. _dotnet-v2-details-begin:
+
+Dependencies
+------------
+
+The .NET plugin needs the ``dotnet`` executable to build programs. The plugin will
```
medubelko, line 168:
```suggestion
The .NET plugin needs the dotnet executable to build programs. The plugin will
```

The command name doesn't need to be in code unless we're specifically discussing its use on the command line.

docs/common/craft-parts/reference/plugins/dotnet_v2_plugin.rst
```diff
@@ -0,0 +1,231 @@
...
+provision it by itself if :ref:`dotnet_version` is set.
+
+If not, some common means of providing ``dotnet`` are:
+
+* The ``dotnet8`` Ubuntu package, declared as a ``build-package``.
+* The ``dotnet-sdk-80`` snap, declared as a ``build-snap`` from the desired channel.
```
medubelko, line 174:
```suggestion
* The dotnet8 Ubuntu package, declared as a ``build-package``.
* The dotnet-sdk-80 snap, declared as a ``build-snap`` from the desired channel.
```

Same here. Package and snap names aren't code.

medubelko, line 174: It's probably worth linking to the package and snap pages.

docs/common/craft-parts/reference/plugins/dotnet_v2_plugin.rst
```diff
@@ -0,0 +1,231 @@
...
+
+Another alternative is to define another part with the name ``dotnet-deps``, and
+declare that the part using the ``dotnet`` plugin comes :ref:`after <after>` the
```
medubelko, line 177:
```suggestion
declare that the part using the .NET plugin (v2) comes :ref:`after <after>` the
```

We should insist on a plain-English name for the plugin, like we have in the title.

docs/common/craft-parts/reference/plugins/dotnet_v2_plugin.rst
```diff
@@ -0,0 +1,231 @@
...
+* Call ``dotnet publish --no-restore --no-build`` with the relevant
+  :ref:`global flags <global_flags>` and :ref:`publish-specific flags <publish_flags>`.
+  The generated assets are placed by default into ``${CRAFT_PART_INSTALL}``.
+
+
+Examples
+--------
+
+The following example uses the ``dotnet`` plugin to build an application with .NET 8 using
+the ``Debug`` configuration, generating assets that are self-contained.
```
medubelko, line 215:
```suggestion
the debug configuration, generating assets that are self-contained.
```

docs/common/craft-parts/reference/plugins/dotnet_v2_plugin.rst
```diff
@@ -0,0 +1,231 @@
...
+
+
+.. code-block:: yaml
+
```
medubelko, line 219:
```suggestion
  :caption: Project file

```

docs/reference/plugins.rst
```diff
@@ -11,6 +11,7 @@ lifecycle.
    :maxdepth: 1
 
    /common/craft-parts/reference/plugins/dotnet_plugin.rst
+   /common/craft-parts/reference/plugins/dotnet_v2_plugin.rst
```
medubelko, line 14: I vote we put the new before the old. @mr-cal What do you think?

medubelko (changes requested):
Hi @mateusrodrigues, thanks for documenting your new feature!

I went over your pages in detail, and provided some suggestions about the docs-as-code as well as a few bits about the naming. Let me know what you think. :)

docs/common/craft-parts/reference/plugins/dotnet_v2_plugin.rst
mateusrodrigues: We're saying it's unset, thus not present in the command if not specified. Should we say "Default: unset" instead?

docs/common/craft-parts/reference/plugins/dotnet_v2_plugin.rst
mateusrodrigues: Thanks, this reads much better! One small detail that I added was `.*proj` instead of just `.proj`, since MSBuild project files can be language-specific, like `.csproj` for C# or `.fsproj` for F#.

docs/common/craft-parts/reference/plugins/dotnet_v2_plugin.rst
medubelko: Yes please!

medubelko (changes requested):
Thanks for the quick turnaround, @mateusrodrigues! Just one leftover bit.

docs/common/craft-parts/craft-parts.wordlist.txt
medubelko: Ah, I think it's actually an issue with the way our spellcheck is set up. Not a problem. No further action needed. :)

medubelko (approved):
Looks great!

@mr-cal @lengau Good to merge?

mr-cal:
> Looks great!
> 
> @mr-cal @lengau Good to merge?

I'm good, @lengau?

PR #1048, opened 2025-03-20
docs: Add reference on partition/overlay env vars

upils:

-----


medubelko (commented):
There's a lot of repeated content with a similar scope here, so I think it would be better if we could combine the two sections together.

Otherwise, this seems like the right place to me. I'd appreciate other opinions, though, before we make further changes. @tigarmo?

upils:
> There's a lot of repeated content with a similar scope here, so I think it would be better if we could combine the two sections together.
> 
> Otherwise, this seems like the right place to me. I'd appreciate other opinions, though, before we make further changes. @tigarmo?

Sadly yes. We need to strike the right balance between avoiding repetition and making it clear that: the variable naming convention is slightly different, the conditions to have them set are different and the directory they reference changes based on the part handled.
I will try another way to present the content.

tigarmo:
I think the location is fine

docs/common/craft-parts/reference/partition_specific_output_directory_variables.rst
```diff
@@ -16,3 +16,24 @@ These variable names will contain the (optional) namespace, partition name, and
   $CRAFT_COMPONENT_BAR_BAZ_PRIME -> partitions/component/bar-baz/prime
 
 (Note that the hyphen in the partition ``component/bar-baz`` is converted to an underscore in the corresponding variable names.)
+
+Partition/Overlay-specific output directory environment variables
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If overlay is enabled too, additional environment variables will be created for use during step processing and execution of user-defined scriptlets.
```
tigarmo, line 23: This is a bit awkward to read, as a user. Whether partitions and/or overlays are enabled is a choice of the application, so it would be weird to read this sentence in Rockcraft's docs (because overlays are always enabled). I don't have a good suggestion on how to address this without revamping the whole thing, though. This is the kind of thing that I think we could solve with the "capability-based" approach I mentioned recently

(I understand that the problem is pre-existing)

docs/common/craft-parts/reference/partition_specific_output_directory_variables.rst
```diff
@@ -16,3 +16,24 @@ These variable names will contain the (optional) namespace, partition name, and
...
+
+These variable names will contain the (optional) namespace, partition name, and lifecycle step, formatted as ``CRAFT_[<namespace>_]<partition>_OVERLAY``. The values of these variables will be the directory that corresponds to that partition, step **and** part.  For instance, if the defined partitions are ``default``, ``kernel``, and ``component/bar-baz``, and the provided configuration define two parts ``my-part`` and ``another-part``, the following environment variables will be created:
```
tigarmo, line 25:
```suggestion
These variable names will contain the (optional) namespace, partition name, and lifecycle step, formatted as ``CRAFT_[<namespace>_]<partition>_OVERLAY``. The values of these variables will be the directory that corresponds to that partition, step **and** part.  For instance, if the defined partitions are ``default``, ``kernel``, and ``component/bar-baz``, and the provided configuration defines two parts ``my-part`` and ``another-part``, the following environment variables will be created:
```

(this is a really long line)

medubelko (approved):
Approving. @upils If you have the chance to improve it further beyond @tigarmo's suggestions, that would be fantastic. But if not, I think it's a good addition nonetheless.

lengau:
@copilot resolve the merge conflicts in this pull request

docs/common/craft-parts/reference/partition_specific_output_directory_variables.rst
lengau: I haven't been able to come up with a better way to do this without having three versions of the document:

1. No overlays
2. Overlays without namespaces
3. Overlays with namespaces

cmatsuoka (approved):

PR #1053, opened 2025-03-28
docs: reference lifecycle details

alesancor1:

-----

This PR adds references to the lifecycle docs page on every occurrence of <lifecycle step>


lengau:
Thanks for the PR! Unfortunately we're probably not going to use this, as cross-linking from the `common` directory in our docs causes some problems. @medubelko will probably be able to explain better.

PR #1054, opened 2025-04-01
feat: use-mvnw support

yanksyoon:

-----


tests/integration/plugins/test_maven/mvnw
yanksyoon, line 1: This file was generated using `mvn wrapper:wrapper`

craft_parts/plugins/maven_plugin.py
```diff
@@ -106,22 +110,43 @@ def get_build_packages(self) -> set[str]:
...
+        self, options: MavenPluginProperties
+    ) -> list[str]:
+        """Validate mvnw file before execution."""
+        if not options.maven_use_mvnw:
+            return []
+        return [
+            """[ -e "${CRAFT_PART_BUILD}/mvnw" ] || \
+echo "mvnw file not found, refer to plugin documentation: \
+https://canonical-craft-parts.readthedocs-hosted.com/en/latest/\
+common/craft-parts/reference/plugins/maven_plugin.html"
```
jdkandersson: Is this supposed to be an error scenario? That the `mvnw` file is not found? Perhaps we can check that earlier when we check for the presence of the executable

docs/common/craft-parts/explanation/maven_plugin.rst
```diff
@@ -0,0 +1,30 @@
+.. _maven_plugin_explanation:
+
+Maven Plugin
+============
+
+Maven plugin is a plugin that simplifies building and packaging Jars for projects that use the
```
jdkandersson, line 6: How about `The Maven plugin simplifies building ...

jdkandersson: Actually, let's practice not getting the documentation perfect on the first round and fix it in follow ups

docs/common/craft-parts/explanation/maven_plugin.rst
```diff
@@ -0,0 +1,30 @@
...
+Maven (mvn) tooling. It is equivalent to running the following command:
+
+.. code-block:: shell
+    
+    mvn package
+
+
+Plugin parameter ``maven_parameters`` is used to additionally provide any arguments to the mvn
+package command above. ``maven_parameters`` parameter can be used to further configure any command
```
jdkandersson, line 15: It seems `maven_parameters` are mentioned twice?

docs/common/craft-parts/reference/plugins/maven_plugin.rst
```diff
@@ -4,8 +4,10 @@ Maven plugin
 ============
 
 The Maven plugin builds Java projects using the Maven build tool.
-This plugin will set the ``JAVA_HOME`` environment variable to the
-path to the latest JDK found in the build environment.
+
+The plugin will check for project Java version compatibility with the build system Java version
+compatibility before executing the build. If an incompatible version is detected, the plugin will
+fail early.
```
jdkandersson, line 10: Should this be removed?

craft_parts/plugins/maven_plugin.py
yanksyoon: I think this should be a build-time check since we are accessing `${CRAFT_PART_BUILD}` directory which is applied on the command line execution shell.

jdkandersson: ah ok I got you

jdkandersson: I think the addition of exit 1 is good

docs/common/craft-parts/reference/plugins/maven_plugin.rst
yanksyoon: Removed, thank you!

docs/common/craft-parts/explanation/maven_plugin.rst
```diff
@@ -0,0 +1,30 @@
...
+line arguments. All values are passed in after the initial ``mvn package`` command, delimited by
+spaces.
+
+Another plugin parameter ``maven_use_mvnw`` is used to leverage Maven wrapper files provided by the
+project to run the package command. Initial mvn executable is replaced with project’s mvnw
+executable file, resulting in ``./mvnw package`` command.
+
+The plugin is able to detect and apply the following proxy environment variables:
+``http_proxy``, ``https_proxy``, ``no_proxy``. These environment variables can be supplied through
```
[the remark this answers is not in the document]
yanksyoon: Added case insensitive feature, thank you :)

craft_parts/plugins/maven_plugin.py
```diff
@@ -177,13 +180,14 @@ def _create_settings(settings_path: Path) -> None:
 
     for protocol in ("http", "https"):
         env_name = f"{protocol}_proxy"
-        if env_name not in os.environ:
+        env_name_upper = env_name.upper()
+        if not any(env in os.environ for env in [env_name_upper, env_name]):
```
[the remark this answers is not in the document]
yanksyoon: since possible protocol values "http" and "https" are already lower, i think we've covered both upper and lower cases - unless we decide to support something like `Http_proxy` or `hTtP_pRoXy` which I don't think is valid :/
Perhaps I can update the documentation to say lower/upper case rather than case-insensitive?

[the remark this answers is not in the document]
yanksyoon: I'm not sure I follow - would you be suggesting the following?
```
        if env.upper() not in os.environ or env not in os.environ:
```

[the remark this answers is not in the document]
yanksyoon: May I suggest the [suggestion syntax](https://stackoverflow.com/a/54239943) to reduce communication errors?
```suggestion
<suggestion code here>
```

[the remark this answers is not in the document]
yanksyoon: I think having upper() or lower() might be more performant than iterating over the whole list of environment variable keys - even though `map(str,lower, os.environ.keys())` *might* be a little more readable.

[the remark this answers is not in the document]
jdkandersson: Yeh it would be more performant, though the differences won't be noticeable since I wouldn't expect thousands of environment variables so I would tend to give more weight to readability. Anyway, up to you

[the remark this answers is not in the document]
yanksyoon: Updated, thank you :)

jdkandersson (approved):

docs/common/craft-parts/reference/plugins/maven_plugin.rst
```diff
@@ -32,16 +45,41 @@ maven-parameters
 
 Used to add additional parameters to the ``mvn package`` command line.
 
+maven-use-mvnw
+~~~~~~~~~~~~~~
+**Type:** boolean
+
+Used to determine whether the build should use the project provided Maven
+wrapper (located at ``<project-root>/mvnw``) executable file. The command ``mvn
+package`` is replaced with ``./mvnw package`` command.
```
medubelko, line 54:
```suggestion
Used to determine whether the build should use the Maven wrapper provided by the project
at ``<project-root>/mvnw``. If turned on, the project command is replaced with ``./mvnw
package``.
```

docs/common/craft-parts/reference/plugins/maven_plugin.rst
```diff
@@ -32,16 +45,41 @@ maven-parameters
...
 
 Environment variables
 ---------------------
 
-This plugin reads the ``http_proxy`` and ``https_proxy`` variables from the environment
-to configure Maven proxy access. A comma-separated list of hosts that should not be
-accessed via proxy is read from the ``no_proxy`` environment variable.
+Environment variables can be specified to modify behavior of the build. Three proxy related case
+insensitive environment variables are treated specially.
+``http_proxy``, ``https_proxy`` and ``no_proxy``.
```
medubelko, line 62:
```suggestion
Environment variables can be specified to modify the behavior of the build. Three
proxy-related, case-insensitive environment variables are treated specially:

- ``http_proxy``
- ``https_proxy``
- ``no_proxy``
```

docs/common/craft-parts/reference/plugins/maven_plugin.rst
```diff
@@ -32,16 +45,41 @@ maven-parameters
...
+~~~~~~~~~~
+
+URL to proxy http request to. The value is mapped to the settings file (.parts/.m2/settings.xml)
+under proxy element.
+
+https_proxy
+~~~~~~~~~~~
+
+URL to proxy https request to. The value is mapped to the settings file (.parts/.m2/settings.xml)
+under proxy element.
```
medubelko, line 77:
```suggestion
URL to proxy HTTP request to. The value is mapped to the settings file
(``.parts/.m2/settings.xml``) under the proxy element.

https_proxy
~~~~~~~~~~~

URL to proxy HTTPS request to. The value is mapped to the settings file
(``.parts/.m2/settings.xml``) under the proxy element.
```

yanksyoon: Thank you for the improvements, updated!

craft_parts/plugins/maven_plugin.py
```diff
@@ -106,24 +110,46 @@ def get_build_packages(self) -> set[str]:
         """Return a set of required packages to install in the build environment."""
         return set()
 
+    @property
+    def _maven_executable(self) -> str:
+        """Return the maven executable to be used for build."""
+        options = cast(MavenPluginProperties, self._options)
+        return "${CRAFT_PART_BUILD}/mvnw" if options.maven_use_mvnw else "mvn"
```
tigarmo, line 117:
```suggestion
        return "${CRAFT_PART_BUILD_WORK}/mvnw" if options.maven_use_mvnw else "mvn"
```

This is to handle cases with `source-subdir`, where the actual sources are in a subdirectory of a repo

craft_parts/plugins/maven_plugin.py
```diff
@@ -106,24 +110,46 @@ def get_build_packages(self) -> set[str]:
...
         ]
 
+    def _get_mvnw_validation_commands(
+        self, options: MavenPluginProperties
+    ) -> list[str]:
+        """Validate mvnw file before execution."""
+        if not options.maven_use_mvnw:
+            return []
+        return [
+            """[ -e ${CRAFT_PART_BUILD}/mvnw ] || {
```
tigarmo, line 143:
```suggestion
            """[ -e ${CRAFT_PART_BUILD_WORK}/mvnw ] || {
```

craft_parts/plugins/maven_plugin.py
```diff
@@ -152,21 +178,25 @@ def _create_settings(settings_path: Path) -> None:
...
+        case_insensitive_env = {item[0].lower(): item[1] for item in os.environ.items()}
+        if env_name not in case_insensitive_env:
             continue
 
-        proxy_url = urlparse(os.environ[env_name])
+        proxy_url = urlparse(case_insensitive_env[env_name])
         proxy = ET.Element("proxy")
         proxy_tags = [
-            ("id", env_name),
+            ("id", env_name.lower()),
```
tigarmo, line 188: isn't it already lower-ed?

craft_parts/plugins/maven_plugin.py
```diff
@@ -152,21 +178,25 @@ def _create_settings(settings_path: Path) -> None:
...
             ("active", "true"),
             ("protocol", protocol),
-            ("host", proxy_url.hostname),
+            ("host", str(proxy_url.hostname)),
```
tigarmo, line 191: how come?

tests/integration/plugins/test_maven.py
```diff
@@ -18,11 +18,47 @@
...
 
-def test_maven_plugin(new_dir, partitions):
+@pytest.fixture
+def use_mvnw(request):
+    if request.param:
+        yield request.param
+        return
+    source_location = Path(__file__).parent / "test_maven"
+    mvnw_file = source_location / "mvnw"
+    mvnw_file = mvnw_file.rename(f"{source_location}/mvnw.backup")
```
tigarmo, line 33: why this rename? Shouldn't the wrapper be ignored if `maven-use-mvnw` is False?

craft_parts/plugins/maven_plugin.py
```diff
@@ -38,6 +38,7 @@ class MavenPluginProperties(PluginProperties, frozen=True):
     plugin: Literal["maven"] = "maven"
 
     maven_parameters: list[str] = []
+    maven_use_mvnw: bool = False
```
tigarmo, line 41: maybe this variable should be `maven_use_wrapper` instead? Feels nicer to look at, and then gradle's can be `gradle_use_wrapper`

medubelko (approved):
Doc changes LGTM.

tests/integration/plugins/test_maven.py
yanksyoon: Yes you're right - however, in order to test cases where `maven-use-wrapper=True` without the `mvnw` file in the project, I've created this helper fixture. I am renaming it because I would like to reset it back once that particular test is done.
Would you know a cleaner way to approach this?
Thank you for the review!

craft_parts/plugins/maven_plugin.py
yanksyoon: Reverted, thank you! (I meant to serialize the port below)

craft_parts/plugins/maven_plugin.py
yanksyoon: This is reads a lot better, changed, thank you :)

mr-cal (approved):
LGTM

craft_parts/plugins/maven_plugin.py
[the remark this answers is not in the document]
yanksyoon: Thanks for catching that! Added :)

yanksyoon:
Security scans should be unblocked by: https://github.com/canonical/craft-parts/pull/1081

tigarmo (approved):

PR #1056, opened 2025-04-02
feat: gradle plugin

yanksyoon:

-----


craft_parts/plugins/gradle_plugin.py
```diff
@@ -0,0 +1,165 @@
...
+    gradle_parameters: list[str] = []
+    gradle_task: str = "build"
+
+    # part properties required by the plugin
+    source: str  # pyright: ignore[reportGeneralTypeIssues]
+
+    @model_validator(mode="after")
+    def gradle_task_defined(self) -> Self:
+        """Gradle task must be defined."""
+        if not self.gradle_task:
```
jdkandersson, line 51: It looks like this default to build above so isn’t it always defined?

craft_parts/plugins/gradle_plugin.py
```diff
@@ -0,0 +1,165 @@
...
+
+    plugin: Literal["gradle"] = "gradle"
+
+    gradle_init_script: str = ""
+    gradle_init_script_parameters: list[str] = []
+    gradle_parameters: list[str] = []
+    gradle_task: str = "build"
+
+    # part properties required by the plugin
+    source: str  # pyright: ignore[reportGeneralTypeIssues]
```
jdkandersson, line 46: What is the specific problem here?

craft_parts/plugins/gradle_plugin.py
```diff
@@ -0,0 +1,165 @@
...
+    ) -> None:
+        """Ensure the environment contains dependencies needed by the plugin.
+
+        :param part_dependencies: A list of the parts this part depends on.
+        :raises PluginEnvironmentValidationError: If gradle is invalid
+          and there are no parts named gradle-deps.
+        """
+        # We can't check for gradle executable here because if the project uses gradlew, we don't
+        # necessarily need gradle to be installed on the system. We can't check if the project uses
+        # gradlew since the project would not be pulled yet.
```
jdkandersson, line 71: So do we need the function at all? Or is it a required override?

craft_parts/plugins/gradle_plugin.py
```diff
@@ -0,0 +1,165 @@
...
+        options = cast(GradlePluginProperties, self._options)
+
+        gradle_cmd = [self.gradle_executable, options.gradle_task]
+        self._setup_proxy()
+
+        return [
+            *self._get_gradle_init_command(options=options),
+            " ".join(gradle_cmd + options.gradle_parameters),
+            # remove gradle-wrapper.jar files included in the project if any.
+            f'find {self._part_info.part_build_dir} -name "gradle-wrapper.jar" -type f -delete',
```
jdkandersson, line 122: What is the reason for deleting please?

craft_parts/plugins/gradle_plugin.py
```diff
@@ -0,0 +1,165 @@
...
+            *self._get_java_post_build_commands(),
+        ]
+
+    def _setup_proxy(self) -> None:
```
jdkandersson, line 126: It looks like some of this is similar to maven, do you think the logic can be unified for that part?

craft_parts/plugins/gradle_plugin.py
yanksyoon: That's right - but just in-case it is defined as `gradle-task: ""`

jdkandersson: ah I see, perhaps include that information in the docstring

craft_parts/plugins/gradle_plugin.py
yanksyoon: I'm getting the following error from Pywright:
```
/home/canonical/craft-parts/craft_parts/plugins/gradle_plugin.py:46:5 - error: "source" overrides a field of the same name but is missing a default value (reportGeneralTypeIssues)
```

This is because `PluginProperties` parent class defines source as `source: str | None = None` and we're overriding it. Other plugins also have the same type ignores to help PyWright :)

craft_parts/plugins/gradle_plugin.py
yanksyoon: We don't need it but I thought it better be reasoned why it is not implemented as a comment :)

craft_parts/plugins/gradle_plugin.py
yanksyoon: Oh perhaps I better update the comment. `gradle-wrapper.jar` is included whenever we're using the Gradle wrapper projects. Since Java tooling (specifically, the `JavaPlugin` parent class) provides a post-build script that moves all detected jar files, we don't want to include the non-build-artifact (gradlew-wrapper.jar in this case) to be staged/installed. Hence, the removal of the gradle-wrapper.jar file :)

craft_parts/plugins/gradle_plugin.py
yanksyoon: That is a good point - but unfortunately at the moment, it might be better to wait for either of them to get merged first so that a merge conflict would be avoided and then we can refactor the common functions together :)

craft_parts/plugins/gradle_plugin.py
jdkandersson: Ah ok I see

craft_parts/plugins/gradle_plugin.py
jdkandersson: Perhaps instead of a noop function that comment can go into the class docstring?

craft_parts/plugins/gradle_plugin.py
jdkandersson: What causes `gradle-wrapper.jar` to be created please?

craft_parts/plugins/gradle_plugin.py
jdkandersson: ok makes sense

craft_parts/plugins/gradle_plugin.py
yanksyoon: It's not created - it's checked in with the source code

jdkandersson: ok I guess it makes sense to remove it. Not sure if there is a way to not copy it in the first place when pulling the source?

yanksyoon: I don't think there is at the moment :/

jdkandersson: ok

tests/integration/plugins/test_gradle.py
```diff
@@ -0,0 +1,163 @@
...
+from pathlib import Path
+
+import pytest
+import yaml
+from craft_parts import LifecycleManager, Step
+from craft_parts.infos import ProjectInfo
+
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2023 Canonical Ltd.
```
vpa1977, line 28: It looks like a copy-paste error?

craft_parts/plugins/gradle_plugin.py
```diff
@@ -0,0 +1,161 @@
...
+                    f"""
+systemProp.{protocol}.proxyHost={proxy_url.hostname}
+systemProp.{protocol}.proxyPort={proxy_url.port}
+systemProp.{protocol}.proxyUser={proxy_url.username}
+systemProp.{protocol}.proxyPassword={proxy_url.password}
+systemProp.{protocol}.nonProxyHosts={no_proxy}
+"""
+                )
+
+    def _get_gradle_init_command(self, options: GradlePluginProperties) -> list[str]:
```
vpa1977, line 153: It looks like we are running `gradle --init-script script` first and then `gradle build`. 
I think the  normal use-case is when we are running `gradle --init-script script build` so that changes done in init script are applied in the build process (e.g. we configure additional plugins in init script). 
Running `gradle --init-script script`  has no impact on the build process.

vpa1977 (changes requested):
A minor nit in the test: second copyright statement.
We probably want to drop separate command for gradle --init-script along with the parameter and just append `--init-script` to the build command when the init script is explicitly specified.

craft_parts/plugins/gradle_plugin.py
vpa1977: probably, they are not much different from the `gradle-parameters`

craft_parts/plugins/gradle_plugin.py
vpa1977: There an additional concern here - in some cases jar files are included as a part of test suite. 

I think it is reasonable to delete artefact that we know should not be included in the build, but this also slows down subsequent builds. Probably a topic for wider discussion later if we wont to give Java plugins ability to provide their artifact location. @tigarmo .

craft_parts/plugins/gradle_plugin.py
```diff
@@ -0,0 +1,165 @@
...
+            if env_name not in case_insensitive_env:
+                continue
+            proxy_url = urlparse(case_insensitive_env[env_name])
+
+            with open(
+                gradle_properties, "a+", encoding="utf-8"
+            ) as gradle_properties_file:
+                gradle_properties_file.write(
+                    f"""
+systemProp.{protocol}.proxyHost={proxy_url.hostname}
```
vpa1977, line 144: i think in other plugins we have used [dedent](https://github.com/canonical/craft-parts/blob/87057b6dac82fbcae245c0b2811343341c62913c/craft_parts/plugins/npm_plugin.py#L287) to keep nice formatting.

vpa1977 (approved):
Probably we need a review from repository maintainers on this one, e.g. @tigarmo or @lengau .

craft_parts/plugins/gradle_plugin.py
yanksyoon: Ah yes, thanks!

jdkandersson (approved):

craft_parts/plugins/gradle_plugin.py
```diff
@@ -0,0 +1,163 @@
...
+from .java_plugin import JavaPlugin
+from .properties import PluginProperties
+
+
+class GradlePluginProperties(PluginProperties, frozen=True):
+    """The part properties used by the gradle plugin."""
+
+    plugin: Literal["gradle"] = "gradle"
+
+    gradle_init_script: str = ""
```
tigarmo, line 38: can you add some comments/docstrings on these parameters?

craft_parts/plugins/gradle_plugin.py
```diff
@@ -0,0 +1,163 @@
...
+class GradlePlugin(JavaPlugin):
+    """A plugin to build parts that use Gradle.
+
+    The Gradle build system is commonly used to build Java projects. This
+    plugin requires a build.gradle in the root of the source tree.
+    This plugin uses the common plugin keywords as well as those for "sources".
+    For more information check the 'plugins' topic for the former and the
+    'sources' topic for the latter.
+
+    Additionally, this plugin uses the following plugin-specific keywords:
```
tigarmo, line 75: what about gradle-init-script and gradle-task?

craft_parts/plugins/gradle_plugin.py
```diff
@@ -0,0 +1,163 @@
...
+    """
+
+    properties_class = GradlePluginProperties
+    validator_class = GradlePluginEnvironmentValidator
+
+    @property
+    def gradle_executable(self) -> str:
+        """Use gradlew by default if it exists."""
+        return (
+            f"{self._part_info.part_build_dir}/gradlew"
```
tigarmo, line 90:
```suggestion
            f"{self._part_info.part_build_subdir}/gradlew"
```

craft_parts/plugins/gradle_plugin.py
```diff
@@ -0,0 +1,163 @@
...
+            if (self._part_info.part_build_dir / "gradlew").exists()
```
tigarmo, line 91:
```suggestion
            if (self._part_info.part_build_subdir / "gradlew").exists()
```

craft_parts/plugins/gradle_plugin.py
```diff
@@ -0,0 +1,163 @@
...
+        self._setup_proxy()
+
+        return [
+            " ".join(
+                gradle_cmd
+                + self._get_gradle_init_command_args(options=options)
+                + options.gradle_parameters
+            ),
+            # remove gradle-wrapper.jar files included in the project if any.
+            f'find {self._part_info.part_build_dir} -name "gradle-wrapper.jar" -type f -delete',
```
tigarmo, line 120:
```suggestion
            f'find {self._part_info.part_build_subdir} -name "gradle-wrapper.jar" -type f -delete',
```

tests/integration/plugins/test_gradle.py
```diff
@@ -0,0 +1,147 @@
...
+@pytest.fixture
+def use_gradlew(request):
+    if request.param:
+        yield request.param
+        return
+    source_location = Path(__file__).parent / "test_gradle"
+    gradlew_file = source_location / "gradlew"
+    gradlew_file = gradlew_file.rename(f"{source_location}/gradlew.backup")
+    yield request.param
+    gradlew_file.rename(f"{source_location}/gradlew")
```
tigarmo, line 60: instead of doing this, I think the cleaner way would be to copy the original sources to a place inside `new_dir`, and remove the file if `use_gradlew` is false.
Otherwise, if the test dies (e.g. the process segfaults) before reaching this `rename()` things will get ugly

tests/integration/plugins/test_gradle.py
```diff
@@ -0,0 +1,147 @@
...
+            plugin: gradle
+            gradle-task: testWrite build
+            gradle-init-script: init.gradle
+            source: {source_location}
+            build-packages: [gradle, openjdk-21-jdk]
+            build-environment:
+                - JAVA_HOME: /usr/lib/jvm/java-21-openjdk-${{CRAFT_ARCH_BUILD_FOR}}
+                - http_proxy: {local_proxy_url}
+                - https_proxy: {local_proxy_url}
+                - GRADLE_USER_HOME: {new_dir}/parts/{part_name}/build/.gradle
```
tigarmo, line 82: why do you need to set this?

tests/integration/plugins/test_gradle.py
```diff
@@ -0,0 +1,147 @@
...
+    parts_yaml = textwrap.dedent(
+        f"""
+        parts:
+          {part_name}:
+            plugin: gradle
+            gradle-task: build
+            source: {source_location}
+            build-packages: [gradle, openjdk-11-jdk]
+            build-environment:
+            - JAVA_HOME: /usr/lib/jvm/java-11-openjdk-${{CRAFT_ARCH_BUILD_FOR}}
```
tigarmo, line 118: or this?

tigarmo (changes requested):
thanks!

tests/integration/plugins/test_gradle.py
vpa1977: We want to point gradle to openjdk-11, otherwise it will tryuse default jre:
```
Depends: default-jre-headless (>= 2:1.8) | java8-runtime-headless, libgradle-core-java (>= 4.4.1-20build1), libgradle-plugins-java (>= 4.4.1-20build1)
```

tests/integration/plugins/test_gradle.py
vpa1977: I guess the intention here is to avoid having  test's gradle cache in user's home directory when the tests are run locally to avoid unreliable results.

mr-cal (commented):
Looks good, thanks! Have you tested this plugin with snapcraft, charmcraft, or rockcraft?

mr-cal (approved):

tests/integration/plugins/test_gradle.py
yanksyoon: Added a comment in code with the explanation (thanks @vpa1977 !)

tests/integration/plugins/test_gradle.py
yanksyoon: Added a comment in code with the explanation :)

yanksyoon:
@tigarmo Hi Tiago! Would you be able to help retrigger the test please? It had a fluke in azure package archive mirror it seems like :)

yanksyoon:
Security scans should be unblocked by: https://github.com/canonical/craft-parts/pull/1081

tigarmo (approved):
thanks!

PR #1057, opened 2025-04-02
docs: add examples to the source field in docs

alesancor1:

-----

This PR adds more examples to the `source` field in documentation.

Note that docs now say "every git url format is supported" but SCP formatted string may cause infinite loading - see https://github.com/canonical/craft-parts/issues/1058

----

See [built docs](https://canonical-craft-parts--1057.com.readthedocs.build/en/1057/)

### DIFF
--- a/docs/common/craft-parts/reference/part_properties.rst
+++ b/docs/common/craft-parts/reference/part_properties.rst
@@ -222,7 +222,34 @@ source
 
 **Step:** pull
 
-The location of the source code and data.
+The location of the source code and data. It can refer to a git
+repository, debian package, zip file, local directory, etc. See
+:ref:`source_type`.
+
+The following example points the source field to a git repository:
+
+.. code:: yaml
+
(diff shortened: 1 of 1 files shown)


jahn-junior (changes requested):
Hello! Thanks for the contribution.

This page has been single sourced with [Kitbash](https://github.com/canonical/pydantic-kitbash) since this PR was opened. If you'd like to add any additional examples, please add them to this [key's Pydantic field](https://github.com/canonical/craft-parts/blob/main/craft_parts/parts.py#L78).

I suggest using no more than three representative examples to the field itself and replacing the last line of its docstring with something like:

> The source can be any of the [supported source types](https://documentation.ubuntu.com/craft-parts/latest/common/craft-parts/reference/part_properties/#source-type), declared with a remote URL or local path.

I understand it's been some time since this was opened, so I'm happy to make this change myself since it's quite a bit different from the work done so far. Just LMK! Thanks again.

PR #1067, opened 2025-04-15
feat: add export and backstage directories

lengau:

-----

These directories are for use by plugins, but not directly by users. This PR also modifies the cargo-use plugin to use the export and backstage directories.

CRAFT-4479


craft_parts/executor/environment.py
```diff
@@ -165,6 +165,8 @@ def _get_global_environment(info: ProjectInfo) -> dict[str, str]:
         "CRAFT_ARCH_TRIPLET_BUILD_FOR": info.arch_triplet_build_for,
         "CRAFT_PARALLEL_BUILD_COUNT": str(info.parallel_build_count),
         "CRAFT_PROJECT_DIR": str(info.project_dir),
+        # Build system configurations for the project.
+        "CARGO_HOME": str(info.dirs.work_dir / "cargo"),
```
sergiusens, line 169: reminder to remove

lengau: While I was removing this I came across this: https://github.com/canonical/craft-parts/blob/5418aa74c151570efe537d301e4fcc2186d33ac6/craft_parts/plugins/rust_plugin.py#L223

Re-examining the provenance of both of these leads me to the conclusion that they're two ways to do the same task of making the `rust` plugin work with the `cargo-use` plugin. This one applies it as a blanket change, meaning that other plugins (e.g. the `python` plugin) can use it, while the other applies it only to the `rust` plugin.

docs/common/craft-parts/craft-parts.wordlist.txt
```diff
@@ -24,6 +24,9 @@ BeforeValidator
 BuildPackageNotFound
 BuildPackagesNotInstalled
 BuildState
+CargoPackagePlugin
+CargoPackagePluginEnvironmentValidator
+CargoPackagePluginProperties
```
sergiusens, line 29: should these also be removed?

sergiusens (approved):

docs/common/craft-parts/craft-parts.wordlist.txt
sergiusens, line 29:
```suggestion
```

lengau: Yep, thank you! Double-checked and couldn't find any other leftovers.

PR #1070, opened 2025-04-16
ci: drop focal tests

mr-cal:

-----

Drops focal from CI now that is no longer available as a github runner (https://github.com/actions/runner-images/issues/11101).

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -58,7 +58,7 @@ jobs:
   unit-tests:
     strategy:
       matrix:
-        os: [ubuntu-20.04, ubuntu-22.04, ubuntu-24.04]
+        os: [ubuntu-22.04, ubuntu-24.04]
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout code
@@ -130,9 +130,6 @@ jobs:
         python-version: ["3.10", "3.12"]
         include:
           - adjective: jammy
-          - os: ubuntu-20.04
-            python-version: "3.10"
-            adjective: focal
           - os: ubuntu-24.04
             python-version: "3.12"
             adjective: noble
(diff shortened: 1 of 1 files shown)


upils (approved):

.github/workflows/tests.yaml
```diff
@@ -153,7 +150,7 @@ jobs:
           sudo apt install -y libapt-pkg-dev intltool fuse-overlayfs python3-apt
       - name: Install general python dependencies
         run: |
-          pip install python-apt@https://launchpad.net/ubuntu/+archive/primary/+sourcefiles/python-apt/2.0.1ubuntu0.20.04.1/python-apt_2.0.1ubuntu0.20.04.1.tar.xz
+          pip install python-apt@https://launchpad.net/ubuntu/+archive/primary/+sourcefiles/python-apt/2.4.0ubuntu1/python-apt_2.4.0ubuntu1.tar.xz
```
lengau, line 153: I'm surprised this (the old one) even worked with noble!

.github/workflows/tests.yaml
```diff
@@ -58,7 +58,7 @@ jobs:
   unit-tests:
     strategy:
       matrix:
-        os: [ubuntu-20.04, ubuntu-22.04, ubuntu-24.04]
+        os: [ubuntu-22.04, ubuntu-24.04]
```
lengau, line 61: While we obviously can't do anything about the runners being shut off, I'm mildly concerned that we may now introduce bugs on the still-supported focal, especially with plugin updates. Obviously we're not running in exactly the same environment that craft-parts runs in within an app, but this was for example how I found out that focal doesn't have a `poetry` package.

It's a non-blocker here, but we should consider whether we want to ask for some self-hosted focal runners to use.

lengau (approved):

.github/workflows/tests.yaml
mr-cal: Me too!

.github/workflows/tests.yaml
mr-cal: Agreed, it would be nice to do some further testing on older systems.  I also agree that we can do this in a follow up effort (possibly driven by starflow).

PR #1077, opened 2025-04-23
test: fix directory test strategy

mr-cal:

-----

Craft Parts would occasionally fail when the strategy generated "default" because there would be 2 partitions named "default".


bepri (approved):

tests/unit/test_dirs.py
```diff
@@ -76,7 +76,9 @@ def test_get_stage_dir_with_partitions(partitions):
 @pytest.mark.usefixtures("enable_partitions_feature")
```
upils, line 76: I suspect the test just above, `test_get_stage_dir_with_partitions` may fail in the same way, so I suggest applying the same fix.

upils (approved):
LGTM. I only have a small suggestion.

PR #1093, opened 2025-05-06
Prototype distributing content from overlay to partitions

upils:

-----


craft_parts/executor/migration.py
```diff
@@ -222,14 +237,20 @@ def clean_shared_overlay(
...
+
+    if overlay_migration_state.contents(partition=partition):
+        files, directories = overlay_migration_state.contents(partition=partition)
 
     # Don't remove entries that also belong to a part.
     for other_state in part_states.values():
-        if other_state:
-            files -= other_state.files
-            directories -= other_state.directories
+        if other_state and other_state.contents(partition=partition):
```
cmatsuoka, line 248: Updating the comment above to reflect these changes could be a good idea.

craft_parts/executor/part_handler.py
```diff
@@ -73,34 +78,148 @@ def __call__(
...
+        srcdir: Path,
+        destdirs: Mapping[str | None, Path],
+        permissions: list[Permissions] | None = None,
+    ) -> None:
+        """Migrate layered content from a partition to destination directories.
+
+        If the source partition is the default one, content can be distributed to other
+        partitions using the provided layout.
+        """
+        if self._src_partition == DEFAULT_PARTITION:
```
cmatsuoka, line 129: In the final refactoring we can have partition objects so we won't need to be testing partition name strings.

craft_parts/executor/part_handler.py
```diff
@@ -73,34 +78,148 @@ def __call__(
...
+            # Distribute content into partitions according to the layout
+            for entry in self._layout:
+                # Only migrate content from the subdirectory indicated by the layout entry
+                sub_path = entry.mount.lstrip("/")
```
cmatsuoka, line 133: `sub_path` is the mount point? Maybe call it `mountpoint_path`?

craft_parts/executor/part_handler.py
```diff
@@ -73,34 +78,148 @@ def __call__(
...
             oci_translation=True,
             permissions=permissions,
         )
-        self.migrated_files |= layer_files
-        self.migrated_dirs |= layer_dirs
+        if dst_partition not in self.migrated_files:
+            self.migrated_files[dst_partition] = {}
+
+        for f in layer_files:
+            src_path = str(Path(sub_path) / f)
```
cmatsuoka, line 194: We can resolve this in a better way in the final version.

craft_parts/executor/part_handler.py
```diff
@@ -73,34 +78,148 @@ def __call__(
...
+
+    def all_migrated_files(self) -> set[str]:
+        """Merge lists of files migrated to every partitions.
+
+        Return a list of paths relative to the source partition.
+        """
+        migrated_files: set[str] = set()
+        for m in self.migrated_files.values():
+            migrated_files |= set(m)
+        return migrated_files
```
cmatsuoka, line 212: We can shorten this to `return set(self.migrated_files.values())`. Also this could be a property? And can `self.migrated_files` be private now that we have this?

craft_parts/executor/part_handler.py
```diff
@@ -73,34 +78,148 @@ def __call__(
...
+
+    def all_migrated_directories(self) -> set[str]:
+        """Merge lists of directories migrated to every partitions.
+
+        Return a list of paths relative to the source partition.
+        """
+        migrated_directories: set[str] = set()
+        for m in self.migrated_directories.values():
+            migrated_directories |= set(m)
+        return migrated_directories
```
cmatsuoka, line 222: Same here.

cmatsuoka (commented):
This is a complex change but it looks reasonable, some extra refactoring is possible and maybe update some comments.

craft_parts/executor/part_handler.py
upils: It is indeed the mount point but I suspect that using this name could lead to confusion (especially given the conversation we had last week, showing that this mount point could mean different things based on the context). 
That could be fine to use this name here but I am not sure `_migrate()` should be aware of this, so I opted for a more generic `sub_path` name.

craft_parts/executor/part_handler.py
upils: I do not think we can do what you suggest. `migrated_files` (and `migrated_directories`) are not of type `dict[str, _Migrated_Contents]`, so dicts of dicts. However the name may now be misleading. What do you think of ` partitions_migrated_files`?

> Also this could be a property?

Correct! I will fix that.

> And can self.migrated_files be private now that we have this?

No, because this is still used to call `consolidate_states()` and `_clean_dangling_whiteouts()` in `_migrate_overlay_files_to_stage()`

PR #1099, opened 2025-05-20
ci: reference the OS to build the tox tests result filename

upils:

-----

`matrix.platform` is not defined (anymore?), so use `matrix.os`.


mr-cal (approved):
I crawled through the git history and it doesn't seem like this ever was correct. 

The artifact also doesn't get uploaded, so maybe that's why we've never noticed the name was wrong.

lengau (approved):

PR #1100, opened 2025-05-23
feat: add Filesystem model

upils:
Add FilesystemMount models to be used later to distribute content between partitions
when migrating from the overlay. Defining a `filesystems` section is only valid when
both partition and overlay features are enabled.

-----

Add FilesystemMount models to be used later to distribute content between partitions when migrating from the overlay.

This implements the new `filesystems` key specified in [ST152 - Craft Partitions and Overlays Integration](https://docs.google.com/document/d/1MtDUlGkq2txvylU2Gzym3R8zgVALAWD9w6GSGix4fYs/edit?tab=t.0)

Defining a `filesystems` section is only valid when both partition and overlay features are enabled.

FilesystemMounts can be given to the LifecycleManager and are stored in ProjectInfo.

CRAFT-4529


upils:
Usage in craft-application https://github.com/canonical/craft-application/pull/769

craft_parts/utils/layout_utils.py
```diff
@@ -0,0 +1,73 @@
...
+    """Validate a layout.
+
+    :param data: The repository data to validate.
+    """
+    if not isinstance(data, list):  # pyright: ignore[reportUnnecessaryIsInstance]
+        raise TypeError("value must be a dictionary")
+    layout = [LayoutItem.unmarshal(item) for item in data]
+
+    if layout[0].mount != "/":
+        raise ValueError("A filesystem first entry must map the '/' mount.")
```
mr-cal, line 34: Is there a test for this?

This could also be a model validator.

craft_parts/utils/layout_utils.py
```diff
@@ -0,0 +1,73 @@
...
+    if (
+        not features.Features().enable_partitions
+        or not features.Features().enable_overlay
+    ):
+        raise errors.FeatureError(
+            "Filesystems are defined but partition feature or overlay feature are not enabled."
+        )
+
+    if len(layouts) > 1:
+        raise errors.FeatureError("One and only one filesystem must be defined.")
```
mr-cal, line 57: I think this is more readable, maybe a personal preference. 
```suggestion
        raise errors.FeatureError("Exactly one filesystem must be defined.")
```

craft_parts/utils/layout_utils.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+    if not layouts:
+        return
+
+    if (
+        not features.Features().enable_partitions
+        or not features.Features().enable_overlay
+    ):
+        raise errors.FeatureError(
+            "Filesystems are defined but partition feature or overlay feature are not enabled."
```
mr-cal, line 53: Where is the line drawn for when to use filesystems instead of layouts? Is it similar to the problem we have with craft-parts raising errors about partitions, when a Snapcraft user has an error with components?

If so, one solution would be to create a new class of errors, LayoutError. That way imagecraft could catch LayoutErrors and raise an error that talks specifically about filesystems.

upils: The "layout" term is for us, to avoid confusion when working with it in craft-parts/craft-applications and *craft applications. This is indeed similar to the problem with partitions because I think users should only be aware of "filesystems". However I expect that users will always interact with the "filesystems" concept, whatever the application, and not with an application-specific concept (components for Snapcraft, disk partitions for Imagecraft, etc.) so in a way this is simpler than the "components" problem. So I am not sure we need to translate the errors raised in craft-parts in the *craft applications. 

I am not satisfied with using `FeatureError` though, so I will try introducing a `LayoutError` as you suggested.

craft_parts/utils/layout_utils.py
upils: I am very tempted to remove this function and instead try to unmarshal the received data into the new Layout class I added to rely on the pydantic validation. What do you think?

craft_parts/__init__.py
```diff
@@ -51,6 +52,8 @@
     "Action",
     "ActionProperties",
     "ActionType",
+    "Layout",
+    "Layouts",
```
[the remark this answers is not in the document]
upils: Damn, I forgot this name was already used in Snapcraft. `FilesystemMounts` looks good, but then `LayoutItem` would be renamed `FilesystemMountItem`. Let see how it looks.

craft_parts/main.py
```diff
@@ -98,6 +98,14 @@ def _process_parts(options: argparse.Namespace) -> None:
 
     partitions = options.partitions.split(",") if options.partitions else None
 
+    layouts_data = None
+    if options.layouts:
+        with open(options.layouts) as opt_layouts_file:
+            filesystems_data = yaml.safe_load(opt_layouts_file)
+            if not isinstance(filesystems_data, dict):
+                raise TypeError("layouts definition must be a dictionary")
```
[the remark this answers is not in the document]
upils: To do that I would need to create a copy of the Project model from craft-application, or at least create a pydantic class with a `filesystem` field. This is probably more code than just checking if the loaded data is a dict so I kept it simple here.

craft_parts/filesystem_mounts.py
```diff
@@ -0,0 +1,163 @@
...
+    root: Annotated[UniqueList[FilesystemMountItem], Field(min_length=1)]
+
+    @field_validator("root", mode="after")
+    @classmethod
+    def first_maps_to_slash(
+        cls, value: list[FilesystemMountItem]
+    ) -> list[FilesystemMountItem]:
+        """Make sure the first item in the list maps the '/' mount."""
+        if value[0].mount != "/":
+            raise ValueError("A filesystem first entry must map the '/' mount.")
```
mr-cal, line 87:
```suggestion
            raise ValueError("The first entry in a filesystem must map the '/' mount.")
```

craft_parts/filesystem_mounts.py
```diff
@@ -0,0 +1,163 @@
...
+    def marshal(self) -> dict[str, Any]:
+        """Create a dictionary containing the filesystem_mount item data.
+
+        :return: The newly created dictionary.
+
+        """
+        return self.model_dump(by_alias=True)
+
+
+class FilesystemMount(RootModel):
```
mr-cal, line 75: Nice choice, I hadn't heard of RootModel before.

craft_parts/filesystem_mounts.py
```diff
@@ -0,0 +1,163 @@
...
+
+FilesystemMounts = SingleEntryDict[Literal["default"], FilesystemMount]
+
+
+def validate_filesystem_mount(data: list[dict[str, Any]]) -> None:
+    """Validate a filesystem_mount.
+
+    :param data: The filesystem mount data to validate.
+    """
+    FilesystemMount.unmarshal(data)
```
mr-cal, line 127: Should this have a try/except and raise `errors.FilesystemMountError`?  Otherwise I think this could raise a TypeError or pydantic error.

Since this and `validate_filesystem_mounts()` are designed to raise errors, a `:raises:` entry in the docstring would be useful.

mr-cal (commented):
Looks good, it's easier to read now that everything is named FilesystemMounts. Just some minor feedback.

craft_parts/layouts.py
```diff
@@ -0,0 +1,159 @@
...
+        return str.__hash__(self.mount)
+
+    def __eq__(self, other: object) -> bool:
+        if type(other) is type(self):
+            return self.mount == cast(LayoutItem, other).mount
+
+        return False
+
+    @classmethod
+    def unmarshal(cls, data: dict[str, Any]) -> "LayoutItem":
```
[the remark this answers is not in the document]
upils: We already have this in `CraftBaseModel` but this class is in craft-application. Maybe we could consider moving it to craft-parts?

craft_parts/filesystem_mounts.py
upils: I improved the error handling. Let me know what you think about it.

cmatsuoka (approved):
Thanks. But there are some linting errors to check.

craft_parts/filesystem_mounts.py
```diff
@@ -0,0 +1,183 @@
...
+        """Create and populate a new ``FilesystemMount`` object from list.
+
+        The unmarshal method validates entries in the input list, populating
+        the corresponding fields in the data object.
+
+        :param data: The list to unmarshal.
+
+        :return: The newly created object.
+
+        :raise TypeError: If data is not a list.
```
mr-cal, line 112:
```suggestion
        :raise TypeError: If data is not a list.
        :raise pydantic.ValidationError: If the data fails validation.
```

craft_parts/filesystem_mounts.py
mr-cal: The handling for pydantic errors looks good but what would happen if `FilesystemMount.unmarshal` raises a `TypeError` on line 115?  Would this be caught on line 180?

mr-cal (commented):
Looks good, one more question about catching errors.

medubelko (commented):
@upils The code is a bit difficult for me to follow. Can you which parts of the model will have keys? Or should I only review the error messages?

upils:
> @upils The code is a bit difficult for me to follow. Can you which parts of the model will have keys? Or should I only review the error messages?

@medubelko I think you can focus on:
- `craft_parts/filesystem_mounts.py` (docstrings and error messages)
- the new parameter to the LifecycleManager https://github.com/canonical/craft-parts/pull/1100/files#diff-68d2086e111409ee4298829260c916b031d71f8b27081f83b135ce59d5ff8a66R86

craft_parts/filesystem_mounts.py
```diff
@@ -0,0 +1,189 @@
...
+    if not filesystem_mounts:
+        return
+
+    if (
+        not features.Features().enable_partitions
+        or not features.Features().enable_overlay
+    ):
+        raise errors.FilesystemMountError(
+            brief="Missing features to use filesystems",
+            details="Filesystems are defined but partition feature or overlay feature are not enabled.",
```
jahn-junior, line 163: I notice that the other errors offer resolutions. Is there a reason this one does not?

```suggestion
            resolution="Enable both the partition and overlay features.",
```

craft_parts/filesystem_mounts.py
```diff
@@ -0,0 +1,189 @@
...
+        )
+
+    if len(filesystem_mounts) > 1:
+        raise errors.FilesystemMountError(
+            brief="Exactly one filesystem must be defined.",
+            resolution="Define a single entry in the filesystems section of the project file.",
+        )
```
jahn-junior, line 170:
```suggestion
    if len(filesystem_mounts) > 1:
        raise errors.FilesystemMountError(
            brief="Only one filesystem can be defined.",
            resolution="Reduce the filesystems section of the project file to a single entry.",
        )
```
I think these messages should specifically point out that there are too many entries.

jahn-junior (commented):
A couple small tweaks to the error messages.

jahn-junior (approved):
Made some tweaks to the error messages and updated the relevant tests. :+1:

PR #1101, opened 2025-05-25
Improve uv plugin doc

astrojuanlu:

-----

I see that the uv plugin was partly inspired by the Go plugin (looking at the first few commits of #930) so I went ahead and expanded the explanation on how to actually get `uv` installed. I've seen some people define a custom `uv-deps` (like https://github.com/nsg/immich-distribution/blob/75d8511d2ca62b11ba25e66a37869eb06ff767e3/snap/snapcraft.yaml#L301-L310 or @lengau 's own https://github.com/lengau/astral-ty/blob/20a4f83c6fa369a189cdfe15e92f460b7c08f560/snap/snapcraft.yaml#L25-L34 ) but I guess `snap install astral-uv` is just easier?

A couple more things:
- `uv run --extra dev --extra apt-noble --extra docs sphinx-build ...` initially didn't work because of a missing `setuptools` dependency, so I added it.
- Since the checklist asks for a changelog entry I added it, but looking at the history of the file it was not immediately clear to me how you folks handle unreleased entries and there's no `CONTRIBUTING.md` either so I did my best, but let me know if you prefer another approach.

### DIFF
--- a/.github/workflows/cla-check.yaml
+++ b/.github/workflows/cla-check.yaml
@@ -3,7 +3,7 @@ on: [pull_request]
 
 jobs:
   cla-check:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     steps:
       - name: Check if CLA signed
-        uses: canonical/has-signed-canonical-cla@v1
+        uses: canonical/has-signed-canonical-cla@v2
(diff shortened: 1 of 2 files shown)


astrojuanlu:
- `main` branch: https://canonical-craft-parts.readthedocs-hosted.com/en/latest/common/craft-parts/reference/plugins/uv_plugin/#dependencies
- This PR: https://canonical-craft-parts--1101.com.readthedocs.build/en/1101/common/craft-parts/reference/plugins/uv_plugin/#dependencies

lengau:
@astrojuanlu thanks! You uncovered that we're still using the old workflow. Once https://github.com/canonical/craft-parts/pull/1115 is merged it should succeed.

PR #1106, opened 2025-05-30
chore: add questing python-apt

lengau:

-----

This adds python-apt for Ubuntu Questing so we can build/test on 25.10


lengau:
(rebased on main)

mr-cal (approved):
Assuming only python-apt change. It's hard to see since the `upload-time` changed on every dep in the lockfile.

PR #1108, opened 2025-06-03
Convert FilesystemMounts to a class

upils:
- [ ] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?
- [ ] Have you added an entry to the changelog (`docs/reference/changelog.rst`)?

-----

PR #1109, opened 2025-06-03
feat: track content migrated to partitions

upils:

-----

Enable migrating content from the overlay to different partitions using the provided FilesystemMounts.

Migrated content is tracked per partition in the default state file via a new `partitions_contents` key in state files.

To be merged **after** https://github.com/canonical/craft-parts/pull/1108

CRAFT-4584


craft_parts/executor/part_handler.py
```diff
@@ -73,34 +79,134 @@ def __call__(
     ) -> None: ...
 
 
+DEFAULT_PARTITION = "default"
+
+# map the source path to the destination path in the partition
+_Migrated_Contents = dict[str, str]
```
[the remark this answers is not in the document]
upils: Good catch -_-

craft_parts/executor/step_handler.py
```diff
@@ -45,21 +45,44 @@
...
+class StepPartitionContents:
     """Files and directories to be added to the step's state."""
 
     files: set[str] = dataclasses.field(default_factory=set)
     dirs: set[str] = dataclasses.field(default_factory=set)
 
 
 @dataclasses.dataclass(frozen=True)
-class StageContents(StepContents):
+class StagePartitionContents(StepPartitionContents):
```
cmatsuoka, line 56: This extends StepPartitionContents adding backstage states to the regular stage contents. Backstage isn't a separate step to have its own instance of StepPartitionContents.

craft_parts/executor/part_handler.py
```diff
@@ -292,14 +396,29 @@ def _run_overlay(
...
+            default_directories = default_contents.dirs
+
         return states.OverlayState(
             part_properties=self._part_properties,
             project_options=step_info.project_options,
-            files=contents.files,
-            directories=contents.dirs,
+            partitions_contents=partitions_contents,
+            files=default_files,
+            directories=default_directories,
```
cmatsuoka, line 421: Maybe this could become a bit more readable if `default_files` and `default_directories` are not declared at all, and `default_contents` defaults to an empty instance of `MigationContents` (set in line 411)?

craft_parts/executor/part_handler.py
```diff
@@ -422,14 +538,31 @@ def _run_stage(
...
-            files=contents.files,
-            directories=contents.dirs,
+            partitions_contents=migration_partitions_contents,
+            files=contents.partitions_contents[DEFAULT_PARTITION].files,
+            directories=contents.partitions_contents[DEFAULT_PARTITION].dirs,
             overlay_hash=overlay_hash.hex(),
-            backstage_files=contents.backstage_files,
-            backstage_directories=contents.backstage_dirs,
+            backstage_files=default_backstage_files,
+            backstage_directories=default_backstage_directories,
```
cmatsuoka, line 565: The same strategy can be applied here by not declaring `default_backstage_files` and `default_backstage_directories`.

craft_parts/executor/part_handler.py
```diff
@@ -464,16 +597,24 @@ def _run_prime(
         ):
             prime_dirs = list(self._part.prime_dirs.values())
             primed_stage_packages = _get_primed_stage_packages(
-                contents.files, prime_dirs=prime_dirs
+                contents.partitions_contents[DEFAULT_PARTITION].files,
+                prime_dirs=prime_dirs,
```
cmatsuoka, line 601: Shouldn't `_migrate overlay_files_to_prime` above happen inside `builtin_prime` so that it won't happen in case of prime overrides? (to be consistent with standard part migration to prime)

Also why is this only taking the default partition into account?

cmatsuoka (commented):
Did an initial pass and added some notes (including notes to myself to not get too confused during review).

craft_parts/executor/step_handler.py
```diff
@@ -45,21 +45,44 @@
...
     backstage_dirs: set[str] = dataclasses.field(default_factory=set)
 
 
+@dataclasses.dataclass(init=False)
+class StepContents:
+    """Contents mapped to partitions."""
+
+    partitions_contents: dict[str, StepPartitionContents | StagePartitionContents] = (
+        dataclasses.field(default_factory=dict)
+    )
```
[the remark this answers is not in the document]
upils: Yes. At some point I thought `StepPartitionContents` could hold the `backstage_files`/`backstage_dirs` fields but that might be error prone and confusing to only use these at the stage step.

craft_parts/executor/part_handler.py
upils: > Shouldn't _migrate overlay_files_to_prime above happen inside builtin_prime so that it won't happen in case of prime overrides? (to be consistent with standard part migration to prime)

I think that is unrelated to this PR but is a good question. Changing that behavior would change the user experience and so might surprise user. It is unclear to me what was the intended behavior initially. The way the code manages the overlay content differs in several ways with the standard migration:
- as you noted this is not part of `_builtin_{stage,prime}`
- migrated files are tracked in another file (`overlay/{stage,prime}_overlay`)
- something else?

I do not know if we want to change that now at the risk of introducing subtle differences in behavior.

> Also why is this only taking the default partition into account?

It was already limited to the default partition, but this is now more explicit since `StepContents` tracks contents for every partition (so the default partition must be selected).

cmatsuoka (approved):
Thanks. The overall processing is starting to get really complex so we may want to document this while we still have all of this fresh in our memories.

medubelko (changes requested):

medubelko (approved):

PR #1139, opened 2025-06-23
fix(maven-use): always set version

bepri:

-----

The maven-use plugin will always overwrite the version number of a dependency with whatever is available (unless, of course, what's available is already being used). In the case that a version number isn't specified at all, instead of raising an error, the plugin should just simply write in a version that is available.

CRAFT-4601

### DIFF
--- a/docs/reference/changelog.rst
+++ b/docs/reference/changelog.rst
@@ -16,6 +16,15 @@ Changelog
 
   For a complete list of commits, check out the `X.Y.Z`_ release on GitHub.
 
+
+X.Y.Z (2025-MM-DD)
+------------------
+
+Bug fixes:
+
+- With the maven-use plugin, do not raise errors if versions are not specified
+  in a project's :file:`pom.xml`.
+
 .. _release-2.14.0:
 
 2.14.0 (2025-06-20)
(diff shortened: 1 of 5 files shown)


docs/reference/changelog.rst
```diff
@@ -16,6 +16,15 @@ Changelog
...
   For a complete list of commits, check out the `X.Y.Z`_ release on GitHub.
 
+
+X.Y.Z (2025-MM-DD)
+------------------
+
+Bug fixes:
+
+- With the maven-use plugin, do not raise errors if versions are not specified
+  in a project's :file:`pom.xml`.
```
```suggestion
- With the maven-use plugin, don't raise errors if dependency versions aren't specified
  in the software's :file:`pom.xml` file.
```

medubelko (approved):

tests/integration/plugins/test_maven_use/java-main-part/pom.xml
```diff
@@ -27,8 +27,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-shade-plugin</artifactId>
-        <!-- invalid version on purpose - must be replaced with the archive one -->
-        <version>X.Y.Z</version>
+        <!-- omitted version on purpose - must be replaced with the archive one -->
```
tigarmo, line 30: 👍🏾

tigarmo (approved):
thanks!

PR #1141, opened 2025-06-24
fix(maven-use): don't duplicate distributionManagement tags

bepri:

-----

In projects that already have a `<distributionManagement>` tag in their `pom.xml` file, maven-use would append a second one instead of modifying the existing one, which caused a build error from Maven. This PR changes the behavior to always overwrite the existing tag, if present. This is acceptable because the tag's purpose is configuring remote indexes to upload builds of the project, but the only place that the maven-use plugin should deploy is to the local backstage directory.

CRAFT-4603

### DIFF
--- a/tests/unit/utils/test_maven_utils.py
+++ b/tests/unit/utils/test_maven_utils.py
@@ -514,6 +514,14 @@ def test_update_pom_no_pom(part_info: PartInfo) -> None:
         update_pom(part_info=part_info, add_distribution=False, self_contained=False)
 
 
+def create_project(part_info: PartInfo, project_xml: str) -> Path:
+    part_info.part_build_subdir.mkdir(parents=True)
+    pom_xml = part_info.part_build_subdir / "pom.xml"
+    pom_xml.write_text(project_xml)
+
+    return pom_xml
+
+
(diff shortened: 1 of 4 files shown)


tests/unit/utils/test_maven_utils.py
```diff
@@ -544,6 +551,50 @@ def test_update_pom_add_distribution(part_info: PartInfo) -> None:
...
+    tree = ET.parse(pom_xml)  # noqa: S314
+
+    # The new distributionManagement is in place
+    project = tree.getroot()
+    distro_element = cast("ET.Element", project.find("distributionManagement"))
+    distro_repo = distro_element.find("repository")
+    assert distro_repo is not None
+    distro_id = distro_repo.find("id")
+    assert distro_id is not None
+    assert distro_id.text == "craft"
```
lengau, line 595: This is ugly.

Approved.

lengau (approved):

tests/unit/utils/test_maven_utils.py
bepri: Yes.

Thanks.

medubelko (approved):

PR #1144, opened 2025-06-26
feat: Use the first partition in the list as the default

upils:

-----

Enable setting a custom name as the first partition in the partition list provided to the lifecycle, and consider it the default one.
Only the first partition in the list can be named 'default'.
Crafters can still reference the default partition with `(default)` in the project file. 
Migration states now also track the partition name to properly track migrated content per partition.

CRAFT-4638

### DIFF
--- a/docs/reference/changelog.rst
+++ b/docs/reference/changelog.rst
@@ -25,6 +25,8 @@ New features:
   local build environment. This attribute requires explicit plugin support and can
   currently be declared for parts using the Maven and Maven-use plugins.
 - Add methods to let MavenPlugin subclasses extend the default behavior of the plugin.
+- Accept a custom name as the first partition in the partition list provided
+  to the lifecycle, and use it to name the default partition.
 
 Bug fixes:
 


craft_parts/infos.py
```diff
@@ -269,6 +270,24 @@ def partitions(self) -> list[str] | None:
...
+    def is_default_partition(self, partition: str | None) -> bool:
+        """Check if given partition is the default one."""
+        if self._partitions is None and (
+            partition is None or partition == DEFAULT_PARTITION
+        ):
+            return True
+        if self._partitions is not None:
+            return partition == self._partitions[0]
+
+        return False
```
upils, line 289: These are sadly slightly different to accommodate their respective usage and due to the handling of the `None`/`"default"` values when partitions are defined or not.

craft_parts/parts.py
```diff
@@ -841,6 +841,13 @@ def has_chisel_as_build_snap(self) -> bool:
         """Return whether this part has chisel in its build-snaps."""
         return self.spec.has_chisel_as_build_snap
 
+    @property
+    def default_partition(self) -> str:
+        """Get the "default" partition from a partition list."""
+        if self._partitions:
+            return self._partitions[0]
+        return DEFAULT_PARTITION
```
upils, line 849: Also slightly different than the implementation in `infos.py` in purpose.

craft_parts/utils/partition_utils.py
```diff
@@ -230,11 +228,10 @@ def get_partition_dir_map(
     """
     if partitions:
         return {
-            "default": base_dir / suffix,
+            partitions[0]: base_dir / suffix,
```
upils, line 231: Almost all the magic of this implementation lies here.

craft_parts/state_manager/step_state.py
```diff
@@ -46,6 +46,7 @@ class MigrationState(BaseModel):
     files from shared areas on step cleanup.
     """
 
+    partition: str | None = None
```
upils, line 49: Tracking the partition name in the state to know where to get the content from when `contents()` is called.

craft_parts/executor/collisions.py
```diff
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright 2015-2021,2024 Canonical Ltd.
+# Copyright 2015-2021,2024-2025 Canonical Ltd.
```
cmatsuoka, line 3: It could be simpler to just say 2015-2025 in cases like this.

craft_parts/infos.py
```diff
@@ -269,6 +270,24 @@ def partitions(self) -> list[str] | None:
...
+        return None
+
+    def is_default_partition(self, partition: str | None) -> bool:
+        """Check if given partition is the default one."""
+        if self._partitions is None and (
+            partition is None or partition == DEFAULT_PARTITION
+        ):
+            return True
+        if self._partitions is not None:
+            return partition == self._partitions[0]
```
cmatsuoka, line 287: When we have a real partition set as the default, using `(default)` should also work (even if not explicitly declared in the partition list -- using `default` and the default parition name should be equivalent). The case is `self._partitions` is not `None`, and `partition == DEFAULT_PARTITION`.

In practice this would only affect `organize`, because internally the default partition name is normalized to the first item in the partition list. Only when organizing the user would refer to the default partition by its name, or literal `(default)`, or nothing. But since this is a method to say if a given partition is the default one, it seems reasonable that it would accept the literal `default` name.

craft_parts/executor/organize.py
```diff
@@ -57,14 +58,16 @@ def organize_files(
         the default partition.
     """
     for key in sorted(file_map, key=lambda x: ["*" in x, x]):
-        src_partition, src_inner_path = path_utils.get_partition_and_path(key)
+        src_partition, src_inner_path = path_utils.get_partition_and_path(
+            key, default_partition
+        )
 
-        if src_partition and src_partition != "default":
+        if src_partition and src_partition != default_partition:
```
cmatsuoka, line 65: The `default` alias should also be valid when organizing. If literal `default` is found, we could do the internal logic using the default partition name.

craft_parts/utils/partition_utils.py
```diff
@@ -66,9 +67,6 @@ def validate_partition_names(partitions: Sequence[str] | None) -> None:
             "Partition feature is enabled but no partitions are defined."
         )
 
-    if partitions[0] != "default":
-        raise errors.FeatureError("First partition must be 'default'.")
```
cmatsuoka, line 70: We can do a slightly different check here: the name `default` can only appear at index 0. You can't use a named default partition, and call something else `default`.

cmatsuoka:
This is simpler than other solutions and should have equivalent functionality. The only thing we need to consider is that the user could organize using either the name of the default partition (if not `default`), literal `default`, or nothing.

craft_parts/executor/organize.py
upils: I wondered about that, it could be nice but also confusing for our users to still use `(default)` in their project file when they expect content to come from `(foo)` (the "renamed" `default`). I will implement what you suggest and I think we can discuss it with Gustavo when playing with the prototype.

cmatsuoka (approved):
Thanks! I think this does what we want it to do.

mr-cal (approved):
LGTM

PR #1147, opened 2025-06-27
feat(maven-use): add support for the parent tag

bepri:

-----

Some projects feature a `<parent>` tag at the top level of their `pom.xml`. This is effectively a pom-as-a-package, meant for inheriting configuration from an "upstream" `pom.xml` file. Since it is a "package", it has a version number that needs to be rewritten by maven-use.
CRAFT-4602

### DIFF
--- a/tests/integration/plugins/test_maven_use.py
+++ b/tests/integration/plugins/test_maven_use.py
@@ -100,7 +100,7 @@ def test_maven_use_plugin_self_contained(new_dir, partitions, monkeypatch, caplo
     expected_bumps = [
         ("org.starcraft.add", "2.0.0", "2.2.0"),
         ("org.starcraft.print-addition", "1.0.0", "1.1.0"),
-        ("org.starcraft.parent", "1.1.0", "1.0.0")
+        ("org.starcraft.parent", "1.1.0", "1.0.0"),
     ]
 
     # Check that versions are adjusted as necessary


docs/reference/changelog.rst
```diff
@@ -22,7 +22,8 @@ X.Y.Z (2025-MM-DD)
 New features:
 
 - Make the error message more detailed and traceable when the maven-use plugin
-  encounters invalid XML in the software's ``pom.xml`` file.
+  encounters invalid XML in the software's :file:`pom.xml` file.
+- Add support for the `parent` tag on a :file:`pom.xml` file with the maven-use plugin.
```
```suggestion
- Add support for the ``parent`` tag in the :file:`pom.xml` file with the maven-use plugin.
```

There's only one, right?

medubelko (approved):

tests/integration/plugins/test_maven_use/self-contained/java-dep-print-addition/pom.xml
```diff
@@ -3,6 +3,12 @@
     xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
+    <parent>
+        <groupId>org.starcraft</groupId>
+        <artifactId>parent</artifactId>
+        <version>1.1</version>
```
tigarmo, line 9: can you add a check that this was updated, on the test?

tigarmo (approved):

mr-cal (approved):

PR #1148, opened 2025-06-27
feat: Plugin-constrained build-attributes

tigarmo:
This commit introduces the concept of build-attributes that craft-parts is aware of, and that need to be explicitly supported by a part's plugin in order to be useable in said part.

Currently the only such attribute is 'self-contained', and only the 'maven' and 'maven-use' plugins support it.

-----


docs/reference/changelog.rst
```diff
@@ -23,6 +23,10 @@ New features:
 
 - Make the error message more detailed and traceable when the maven-use plugin
   encounters invalid XML in the software's ``pom.xml`` file.
+- Add the ``self-contained`` build-attribute. This attribute declares that a plugin's
+  build must be done without accessing resources outside the build system (like the
+  Internet), and as such requires explicit plugin support. Currently the ``maven`` and
+  ``maven-use`` support this attribute.
```
jahn-junior, line 29:
```suggestion
- Add the ``self-contained`` build attribute, which constrains a part's build to the
  local build environment. This attribute requires explicit plugin support and can
  currently be declared for parts using the Maven and Maven-use plugins.
```

lengau (approved):
LGTM!

cmatsuoka (approved):
Add some comments but no blockers.

PR #1151, opened 2025-07-01
perf: Conditional imports for dependencies only used for type checks

upils:

-----

Add 3 linter rules to check for imports for dependencies only used for type checks.
Fixes errors found by these rules (and 2 other unrelated lint errors).

Also rely on deferred type checking to import some deps only for type checking. These are not currently helping to reduce import time (because the same deps are imported in other modules) but I am experimenting to see if we could foster this habit and over time avoid some imports are runtime.


craft_parts/executor/environment.py
```diff
@@ -19,21 +19,23 @@
...
+    from craft_parts.infos import ProjectInfo, StepInfo
+    from craft_parts.parts import Part
+    from craft_parts.plugins import Plugin
+
 logger = logging.getLogger(__name__)
 
 
 def generate_step_environment(
-    *, part: Part, plugin: Plugin, step_info: StepInfo
+    *, part: "Part", plugin: "Plugin", step_info: "StepInfo"
```
mr-cal, line 38: Have you considered a `from __future__ import annotations` to not have to quote imports?  I think that's the more modern approach, if I recall correctly.

pyproject.toml
```diff
@@ -330,6 +330,9 @@ lint.select = [  # Base linting rule selections.
     # The team have chosen to only use type-checking blocks when necessary to prevent circular imports.
     # As such, the only enabled type-checking checks are those that warn of an import that needs to be
     # removed from a type-checking block.
```
mr-cal, line 332: At some point, it may be good to update this guidance in starbase.

mr-cal (approved):

bepri (approved):

PR #1164, opened 2025-07-09
feat: symlinks for aliased partition to default dirs

upils:

-----

When using an alias for the default partition, create symlinks for this partition targeting the default directories in the working directories. 
Properly prime stage packages when the default partition is aliased.
Add a fixture case with an aliased default partition and fix tests to properly handle it.


docs/reference/changelog.rst
```diff
@@ -16,13 +16,22 @@ Changelog
 
   For a complete list of commits, check out the `X.Y.Z`_ release on GitHub.
 
+.. _release-2.18.0:
+
+2.18.0 (2025-MM-DD)
+-------------------
+
+- When the default partition is aliased, create symlinks in the ``partitions``
+  directory to the default directories at the root of the working directory.
```
medubelko, line 25: Trying to clarify this a bit:

```suggestion
- When the default partition is aliased, create symlinks in the ``partitions``
  directory to the lifecycle directories in the working directory. The links are
  intended to help users during debugging. For example, ``/root/stage`` in the
  container filesystem is linked to ``/root/partitions/<partition-id>/stage``.
```

medubelko (approved):

craft_parts/executor/part_handler.py
```diff
@@ -1146,6 +1145,33 @@ def _clean_shared(
...
+        partition_dir = self._part.dirs.work_dir / "partitions" / default_partition
+        partition_dir.mkdir(parents=True, exist_ok=True)
+
+        for src, dst in [
+            (self._part_info.stage_dir, partition_dir / "stage"),
+            (self._part_info.prime_dir, partition_dir / "prime"),
+            (self._part_info.overlay_dir, partition_dir / "overlay"),
+            (self._part_info.parts_dir, partition_dir / "parts"),
+        ]:
+            if dst.exists():
```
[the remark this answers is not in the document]
upils: If between 2 runs the alias for the default partition changes then symlinks in a new directory in `partitions/` would be created and existing one should be deleted. But we are currently already not solving that with regular directories for concrete partitions so I think it should be fixed later.

As @cmatsuoka pointed out I missed the cleaning case, where some top-level dirs (stage, prime, etc.) could be removed. I will look into improving that.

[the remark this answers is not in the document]
mr-cal: Makes sense to me, thanks!

Are you going to address cleaning in this PR or a follow-up?

craft_parts/executor/part_handler.py
```diff
@@ -1146,6 +1145,39 @@ def _clean_shared(
...
+            (self._part_info.stage_dir, self._part_info.stage_alias_symlink),
+            (self._part_info.prime_dir, self._part_info.prime_alias_symlink),
+            (self._part_info.overlay_dir, self._part_info.overlay_alias_symlink),
+        ]:
+            if dst.exists():
+                if not dst.is_symlink():
+                    # Between two runs of the lifecycle, the default partition alias name
+                    # can be changed to a previously concrete partition by the user.
+                    raise RuntimeError(
+                        f"cannot create symlinks {dst}, a concrete directory already exists."
```
upils, line 1171: @mr-cal I am not sure about this error handling. I am tempted to advise the user to clean everything but  it sounds wrong. What do you think?

mr-cal: Looking at the other runtime errors in craft_parts/executor/, I agree with you.

I'm not sure if we should suggest to clean. If this error happens unexpectedly and cleaning works, the user may not report the bug.

craft_parts/infos.py
```diff
@@ -281,6 +281,50 @@ def default_partition(self) -> str | None:
...
+        if not self.default_partition:
+            return None
+        return self._dirs.work_dir / "partitions" / self.default_partition
+
+    @property
+    def parts_alias_symlink(self) -> Path | None:
+        """Parts directory for the default alias partition."""
+        if self.alias_partition_dir:
+            return self.alias_partition_dir / "parts"
+        return None
```
lengau, line 305: This is a micro-optimisation, but if you don't want to call `alias_partition_dir` twice you could do:


```suggestion
        if alias_partition_dir := self.alias_partition_dir:
            return alias_partition_dir / "parts"
        return None
```

tests/unit/features/partitions/executor/test_executor.py
```diff
@@ -85,6 +107,14 @@ def test_clean(self, new_dir, partitions):
             (prime_dir / "primed-file").exists() is False for prime_dir in prime_dirs
         )
 
+        # Test symlinks for alias to default partition
+        if info.stage_alias_symlink:
+            assert info.stage_alias_symlink.exists() is False
```
lengau, line 112: Same idea, but we fail if something unexpected changes in the future.

```suggestion
        assert info.stage_alias_symlink is None or info.stage_alias_symlink.exists() is False
```

craft_parts/executor/part_handler.py
medubelko: I defer to Alex on this one, unless we're uncertain about the wording.

tests/unit/features/partitions/executor/test_executor.py
```diff
@@ -67,6 +69,26 @@ def test_clean(self, new_dir, partitions):
         info = ProjectInfo(
             application_name="test", cache_dir=new_dir, partitions=partitions
         )
+
+        # Create symlinks if default partition aliased
```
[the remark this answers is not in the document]
upils: Yes, `_symlink_alias_to_default` but I did not want to call it on purpose to not rely on its behavior. I followed the same logic as the rest of this test, creating dirs/symlinks in the test without relying on `_make_dirs`.

medubelko (approved):
Docs LGTM.

upils:
Not waiting for the tests to pass since the last change was a typo fix.

PR #1166, opened 2025-07-09
feat(maven-use): support multi-module projects

bepri:

-----

Adds support for multi-module Maven projects with the maven-use plugin.

Multi-module projects are ones that declare the `<modules>` tag in their `pom.xml`, with nested `<module>` tags that give a relative path to another directory with its own `pom.xml`. These in turn can feature their own modules with their own `pom.xml`s. Maven-use now recursively discovers these `pom.xml` files and patches them just as it would normally patch the top-level one.


craft_parts/utils/maven/common.py
```diff
@@ -147,46 +148,43 @@ def update_pom(
...
 
-    if deploy_to is not None:
-        # Add a distributionManagement element, to tell "maven deploy" to deploy the
-        # artifacts (jars, poms, etc) to the export dir.
-        distribution_dir = deploy_to / "maven-use"
-        distribution_element = ET.fromstring(  # noqa: S314, unsafe parsing with xml
-            DISTRIBUTION_REPO_TEMPLATE.format(repo_uri=distribution_dir.as_uri())
-        )
+    # We should at least have a single pom file
+    if len(poms) == 0 or not poms[0].is_file():
```
lengau, line 155:
```suggestion
    if not poms not poms[0].is_file():
```

craft_parts/utils/maven/common.py
```diff
@@ -43,9 +43,10 @@
 
 ArtifactDict = dict[str, set["MavenArtifact"]]
 GroupDict = dict[str, ArtifactDict]
+Namespaces = dict[str, str]
```
lengau, line 46: Bonus points for making this more specific

medubelko (approved):
Changelog LGTM.

craft_parts/utils/maven/common.py
```diff
@@ -559,3 +551,74 @@ def _format_xml_str(element: ET.Element) -> str:
...
+    try:
+        modules = _find_element(project, "modules", namespaces)
+    except MavenXMLError:
+        return
+
+    # For every module found:
+    for module in modules.findall("module", namespaces):
+        # - Append it to the list of poms that need patching
+        path_str = _get_element_text(module)
+        pom_path = parent_pom.parent / path_str / "pom.xml"
```
tigarmo, line 614: what if `pom_path` is not a valid file here for some reason? For example: the submodule comes from an actual git submodule which wasn't pulled.
I think in this case we should `log` and `continue` (and not add it to `all_poms`), because it might not be an error

craft_parts/utils/maven/common.py
```diff
@@ -559,3 +551,74 @@ def _format_xml_str(element: ET.Element) -> str:
...
+    for prefix, uri in namespaces.items():
+        ET.register_namespace(prefix, uri)
+    return namespaces
+
+
+def _get_poms(part_info: PartInfo, existing: GroupDict) -> list[Path]:
+    """Get a list of poms on a project.
+
+    Each submodule is added to the list of existing artifacts because Maven's build
+    will put them into the backstage as they are built. If a build somehow fails
```
tigarmo, line 570: I don't think this is correct: the modules will be deployed to the **export** dir (and then migrated to the backstage) in the case of the maven-use plugin, but that's not the only place we deploy to currently. Might be simpler to say that the submodules are visible to maven during the build itself, or something like that.

Also, are you sure they are deployed "as they are built"?

craft_parts/utils/maven/common.py
```diff
@@ -147,46 +148,43 @@ def update_pom(
...
 
-    if deploy_to is not None:
-        # Add a distributionManagement element, to tell "maven deploy" to deploy the
-        # artifacts (jars, poms, etc) to the export dir.
-        distribution_dir = deploy_to / "maven-use"
-        distribution_element = ET.fromstring(  # noqa: S314, unsafe parsing with xml
-            DISTRIBUTION_REPO_TEMPLATE.format(repo_uri=distribution_dir.as_uri())
-        )
+    # We should at least have a single pom file
+    if not poms or not poms[0].is_file():
```
tigarmo, line 155: When would `poms[0]` not be a file? Following up on my other comment, I feel like this list should only contain existing poms

tests/integration/plugins/test_maven_use.py
```diff
@@ -112,3 +113,51 @@ def test_maven_use_plugin_self_contained(new_dir, partitions, monkeypatch, caplo
...
+    project_dir = Path(new_dir) / "project"
+    shutil.copytree(SOURCE_DIR / "multi-module", project_dir)
+    monkeypatch.chdir(project_dir)
+    work_dir = Path(new_dir)
+    parts_yaml = (project_dir / "parts.yaml").read_text()
+    parts = yaml.safe_load(parts_yaml)
+
+    lf = LifecycleManager(
+        parts,
+        application_name="test_go",
```
tigarmo, line 134:
```suggestion
        application_name="test_maven_use_with_modules",
```

tests/integration/plugins/test_maven_use.py
```diff
@@ -112,3 +113,51 @@ def test_maven_use_plugin_self_contained(new_dir, partitions, monkeypatch, caplo
...
+    jar_dir = lf.project_info.prime_dir / "jar"
+    jars = {i.name for i in jar_dir.iterdir()}
+    # Only the two jars produced by the main part
+    # (the shade plugin creates the "original-*" one)
+    assert jars == {"hello-world-0.1.0.jar", "original-hello-world-0.1.0.jar"}
+
+    jar = jar_dir / "hello-world-0.1.0.jar"
+    assert jar.is_file()
+
+    output = subprocess.check_output(["java", "-jar", str(jar)], text=True)
```
tigarmo, line 153: I don't think this test needs to check and run the final jar, feels a bit redundant with the other test
Instead, I think this test should look into the backstage and see if the artifacts for all modules are there

tests/unit/utils/test_maven_utils.py
tigarmo, line 1: I would say we're missing a basic unittest for `_get_poms()` here. Something that builds a basic tree with a few submodules and then checks the list of returned poms

tigarmo (changes requested):

craft_parts/utils/maven/common.py
bepri: Good catch - this is a great place to validate each pom being valid. In what situation would it not be an error, though? The module paths are relative paths to a module that must be present. If Python can't find the pom file, neither can Maven and it'll cause a build time error either way.

craft_parts/utils/maven/common.py
```diff
@@ -571,3 +554,92 @@ def _format_xml_str(element: ET.Element) -> str:
...
+    poms: list[Path] = []
+
+    if base_pom is None:
+        base_pom = part_info.part_build_subdir / "pom.xml"
+
+    if not base_pom.is_file():
+        raise MavenXMLError("'pom.xml' does not exist")
+    poms.append(base_pom)
+
+    _recurse_submodules(part_info, base_pom, poms, existing)
```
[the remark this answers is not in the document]
bepri: I agree with this. I can't tell if you're asking me for anything actionable though or if you're making a note?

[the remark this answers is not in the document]
tigarmo: I started writing the comment to ask you to disable this recursion for deployed (so not-source) poms, and then convinced myself that it's actually ok 

so I'm just making a note!

craft_parts/utils/maven/common.py
bepri: Discussed offline. Even if it's a Maven error, it's better to let Maven detect and report the error than try to model its behavior ourselves

tigarmo (approved):
thanks!

PR #1168, opened 2025-07-11
feat: add call to obtain the prime state timestamp

cmatsuoka:

-----

### DIFF
--- a/tests/unit/test_lifecycle_manager.py
+++ b/tests/unit/test_lifecycle_manager.py
@@ -18,6 +18,7 @@
 
 import sys
 import textwrap
+import time
 from pathlib import Path
 from typing import Any
 from unittest.mock import ANY, call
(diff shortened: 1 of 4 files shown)


tests/unit/test_lifecycle_manager.py
```diff
@@ -288,6 +289,30 @@ def test_strict_plugins_error(self, new_dir, mock_available_plugins):
             )
         assert "p1" in str(exc.value)
 
+    def test_get_prime_state_timestamp(self, new_dir, mock_available_plugins):
```
bepri, line 292: Actually another thought, a test that exercises the `max()` piece by having multiple different files would be good

cmatsuoka: It currently has more than one so `max()` needs to pick the correct one, but it seems that we have no test for the case where there are no files! I'll add one.

lengau (approved):
LGTM!

One comment about how we're selecting the most recent prime timestamp, but it's a trade-off that I'm not sure we actually care about right now.

PR #1169, opened 2025-07-13
feat(jlink): introduce add_modules parameter

vpa1977:

-----

### DIFF
--- a/craft_parts/plugins/jlink_plugin.py
+++ b/craft_parts/plugins/jlink_plugin.py
@@ -30,6 +30,7 @@ class JLinkPluginProperties(PluginProperties, frozen=True):
 
     plugin: Literal["jlink"] = "jlink"
     jlink_jars: list[str] = []
+    jlink_add_modules: list[str] = []
 
 
 class JLinkPluginEnvironmentValidator(PluginEnvironmentValidator):
@@ -69,6 +70,11 @@ class JLinkPlugin(Plugin):
             return f"PROCESS_JARS={jars}"
 
         return 'PROCESS_JARS=$(find ${CRAFT_STAGE} -type f -name "*.jar")'
+
+    def _get_extra_module_list(self) -> str:
+        """Return additional modules."""
+        options = cast(JLinkPluginProperties, self._options)
+        return ",".join([x.strip() for x in options.jlink_add_modules])
 
     @override
     def get_build_packages(self) -> set[str]:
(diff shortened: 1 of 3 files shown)


craft_parts/plugins/jlink_plugin.py
```diff
@@ -70,6 +71,11 @@ def _get_find_jars_commands(self) -> str:
 
         return 'PROCESS_JARS=$(find ${CRAFT_STAGE} -type f -name "*.jar")'
 
+    def _get_extra_module_list(self) -> str:
+        """Return additional modules."""
+        options = cast(JLinkPluginProperties, self._options)
+        return ",".join([x.strip() for x in options.jlink_add_modules])
```
tigarmo, line 77: I think I understand the need, now I'm trying to understand the feature: how does jlink find these modules listed in `jlink_add_modules`?

tests/integration/plugins/test_jlink.py
```diff
@@ -89,6 +89,40 @@ def test_jlink_plugin_embedded_jar(new_dir, partitions):
...
+@pytest.mark.usefixtures("build_test_jar")
+def test_jlink_plugin_add_modules(new_dir, partitions):
+    parts_yaml = textwrap.dedent(
+        """
+        parts:
+            my-part:
+                plugin: jlink
+                source: .
+                jlink-jars: ["test.jar"]
+                jlink-add-modules: ["jdk.crypto.ec", "java.sql"]
```
tigarmo, line 101: What is the difference between a module and a package? The internet tells me that a module is a collection of packages, is that it?
For "java.sql" I see that apparently there's both a module and a package: https://docs.oracle.com/en/java/javase/11/docs/api/java.sql/module-summary.html . Does `jlink-add-modules` only apply to modules?

craft_parts/plugins/jlink_plugin.py
```diff
@@ -30,6 +30,7 @@ class JLinkPluginProperties(PluginProperties, frozen=True):
 
     plugin: Literal["jlink"] = "jlink"
     jlink_jars: list[str] = []
+    jlink_add_modules: list[str] = []
```
tigarmo, line 33: Since the command line argument is called `--add-modules`, I wonder if `jlink_extra_modules` would make it clearer that these modules are included *in addition to* the plugin's default behavior of crawling the build and stage directories

craft_parts/plugins/jlink_plugin.py
vpa1977: They are appended on line 157 to the list of modules discovered by jdeps.

craft_parts/plugins/jlink_plugin.py
vpa1977: Makes sense, this also makes naming in the code consistent.

vpa1977:
Unit tests failing on `FAILED tests/unit/features/partitions/executor/test_step_handler.py::TestStepHandlerBuiltins::test_run_builtin_build[partitions2]`

tigarmo (approved):
Thank you! Looks good to me

PR #1172, opened 2025-07-15
refactor(maven-use): use lxml instead of base xml for all parsing

bepri:

-----

Refactors maven-use to use the `lxml` python library, which is safer and allows for (slightly) cleaner XML code. Of note:

- The global parser object is configured with what appear to be the most secure options, informed by [this page](https://lxml.de/FAQ.html#how-do-i-use-lxml-safely-as-a-web-service-endpoint) and [this page](https://lxml.de/parsing.html#parsers). It is then used on every instance of `parse()` and `fromstring()`
- An additional option is used to strip whitespace from XML files, which allows all output XML to be pretty-printed. This aids both tests and future debugging
- XML nodes became doubly-linked-lists, which allows for prepending elements. This is mainly useful for adding comments to an XML file about changes.
- LXML's inner implementation is such that `lxml.etree.Element` is not a type, but rather a function that returns `_Element`. Therefore, it's unfortunately required to typehint with a private element. This seems to be the generally agreed upon approach online, though (Ruff doesn't even mind!).

CRAFT-4593.

### DIFF
--- a/craft_parts/utils/maven/common.py
+++ b/craft_parts/utils/maven/common.py
@@ -573,7 +573,8 @@ def _get_namespaces(project: etree._Element) -> Namespaces:
         # this particular case and move on.
         except ValueError as ve:  # noqa: PERF203
             if "Invalid namespace URI" in str(ve):
-                pass
+                continue
+            raise
     return namespaces
 
 


craft_parts/utils/maven/common.py
```diff
@@ -46,7 +45,19 @@
...
 GroupDict = dict[str, ArtifactDict]
-Namespaces = dict[str, str]
+Namespaces = dict[str | None, str]
+
+_XML_PARSER = etree.XMLParser(
+    # Attempt to recover if an error is encountered
+    recover=True,
+    # Entities can resolve into zip-bomb-like packages, so ignore them
+    resolve_entities=False,
+    # Removing blank text allows the `pretty_printing` kwarg to work later
```
mr-cal, line 55:
```suggestion
    # Removing blank text allows the `pretty_print` kwarg to work later
```

mr-cal (approved):
LGTM

craft_parts/utils/maven/common.py
```diff
@@ -46,7 +45,19 @@
...
+Namespaces = dict[str | None, str]
+
+_XML_PARSER = etree.XMLParser(
+    # Attempt to recover if an error is encountered
+    recover=True,
+    # Entities can resolve into zip-bomb-like packages, so ignore them
+    resolve_entities=False,
+    # Removing blank text allows the `pretty_print` kwarg to work later
+    remove_blank_text=True,
+    # Just no
```
lengau, line 57: 🤣

lengau (approved):

upils (approved):

PR #1185, opened 2025-07-24
fix: Set CRAFT_DEFAULT_* env vars when the default partition is aliased

### LINKED ISSUE #1171: Missing CRAFT_DEFAULT_* env var when aliasing the default partition
### Bug Description

When using an alias for the default partition (using a different name than `default` for the first entry in the partitions list), `CRAFT_DEFAULT_*` env vars are not created. These env vars should always be defined when the partition feature is enabled. 

### To Reproduce

python3 -mcraft_parts --partitions test --overlay-base /tmp/overlay --trace -v prime

### part yaml

```shell
parts:
  part1:
    plugin: nil
    overlay-script: |
      echo $CRAFT_DEFAULT_STAGE
```

### Relevant log output

```shell
DEBUG:craft_parts.utils.os_utils:umount mountpoint='/home/paul/scratch/craft-parts/overlay/overlay', args=()
Error: 'overlay-script' in part 'part1' failed with code 1.

:: /tmp/tmphwhrhumu/scriptlet.sh: line 5: CRAFT_DEFAULT_STAGE: unbound variable
Review the scriptlet and make sure it's correct.
```

upils:

-----

Fix #1171 
CRAFT-4653

### DIFF
--- a/docs/reference/changelog.rst
+++ b/docs/reference/changelog.rst
@@ -16,16 +16,6 @@ Changelog
 
   For a complete list of commits, check out the `X.Y.Z`_ release on GitHub.
 
-.. _release-2.20.0:
-
-2.20.0 (2025-MM-DD)
-------------------
-
-Bug fixes:
-
-- ``CRAFT_DEFAULT_*`` environment variables are set when the default partition 
-  is aliased.
-
 .. _release-2.19.0:
 
 2.19.0 (2025-MM-DD)
@@ -43,6 +33,8 @@ Bug fixes:
 
 - The ``--recurse-submodules`` link in the ``source-submodules`` docstring
   no longer produces a linkcheck error.
+- ``CRAFT_DEFAULT_*`` environment variables are set when the default partition 
+  is aliased.
 
 For a complete list of commits, check out the `2.19.0`_ release on GitHub.
 
--- a/docs/reference/changelog.rst
+++ b/docs/reference/changelog.rst
@@ -16,6 +16,16 @@ Changelog
 
   For a complete list of commits, check out the `X.Y.Z`_ release on GitHub.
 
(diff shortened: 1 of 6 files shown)


docs/reference/changelog.rst
```diff
@@ -16,6 +16,16 @@ Changelog
 
   For a complete list of commits, check out the `X.Y.Z`_ release on GitHub.
 
+.. _release-2.20.0:
+
+2.20.0 (2025-MM-DD)
```
tigarmo, line 21: well, this is a bugfix so it should be at least a patch release (2.19.1)
But I suggest we roll this back into 2.19.0

tests/unit/features/partitions/executor/test_environment.py
```diff
@@ -36,6 +36,18 @@
                 "CRAFT_FOO1_BAR_BAZ2_PRIME",
             },
         ),
+        (
+            # exercise aliased default partition
```
tigarmo, line 40: 👍🏾

tigarmo (changes requested):

docs/reference/changelog.rst
```diff
@@ -33,6 +33,8 @@ Bug fixes:
 
 - The ``--recurse-submodules`` link in the ``source-submodules`` docstring
   no longer produces a linkcheck error.
+- ``CRAFT_DEFAULT_*`` environment variables are set when the default partition
```
tigarmo, line 36: Just to confirm: is there a released version of craft-parts that has this bug? Or did the bug + fix happen between releases?

upils: This bug is here since 2.16.0, when the alias mechanism was added.

PR #1187, opened 2025-07-25
[PROTOTYPE] feat: build first the overlay-populating part

upils:

-----

If a part `foo` organizes content into the overlay AND at least one other part declares a dependency on this part, then the OVERLAY and BUILD steps of this `foo` part run before the OVERLAY step of other dependent parts.

### DIFF
--- a/craft_parts/executor/part_handler.py
+++ b/craft_parts/executor/part_handler.py
@@ -46,7 +46,7 @@ from craft_parts.state_manager import (
 from craft_parts.state_manager.stage_state import StageState
 from craft_parts.steps import Step
 from craft_parts.utils import file_utils, os_utils
-from craft_parts.utils.partition_utils import DEFAULT_PARTITION
+from craft_parts.utils.partition_utils import DEFAULT_PARTITION, OVERLAY_IDENTIFIER
 
 from . import filesets, migration
 from .environment import generate_step_environment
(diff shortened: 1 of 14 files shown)


craft_parts/sequencer.py
```diff
@@ -216,11 +216,29 @@ def _get_project_vars(self, part: Part, step: Step) -> dict[str, ProjectVar] | N
...
+            steps.ovl_bootstrap_dependency_prerequisite_step(step)
+        )
+        if prerequisite_step_bootstrap_ovl:
+            all_overlay_init_deps = parts.part_dependencies_overlay_bootstrap(
+                part, part_list=self._part_list
+            )
+            self._add_deps_actions(
+                part, step, all_overlay_init_deps, prerequisite_step_bootstrap_ovl
+            )
+
```
upils, line 230: This is the heart of the PR.

craft_parts/executor/part_handler.py
```diff
@@ -457,6 +457,32 @@ def _run_build(
                     stdout=stdout,
                     stderr=stderr,
                 )
+
+                # Organize the installed files as requested. We do this in the build step for
```
tigarmo, line 461: This (useful) comment got moved and now it kind of looks like it only applies to this branch (the one with overlay visibility), which isn't true. How about keeping a single `organize_files()` call and only defining the parameters in this `if` branch?

craft_parts/parts.py
```diff
@@ -848,6 +857,14 @@ def default_partition(self) -> str:
             return self._partitions[0]
         return DEFAULT_PARTITION
 
+    @property
+    def bootstrap_overlay(self) -> bool:
```
tigarmo, line 861:
```suggestion
    def is_bootstrap_overlay(self) -> bool:
```

Makes it clearer that it's a query, and not something else

craft_parts/parts.py
```diff
@@ -1073,6 +1097,20 @@ def part_dependencies(
     return dependencies
 
 
+def part_dependencies_overlay_bootstrap(
+    part: Part, *, part_list: list[Part]
+) -> set[Part]:
+    """Return a set of all the parts upon which the named part depends and bootstrapping the overlay.
```
tigarmo, line 1103: This is a bit confusing to me, do you mean "and _that are_ bootstrapping the overlay"?

craft_parts/parts.py
```diff
@@ -1073,6 +1097,20 @@ def part_dependencies(
     return dependencies
 
 
+def part_dependencies_overlay_bootstrap(
```
tigarmo, line 1100: Would `dependencies_that_bootstrap_overlay` be a clearer name?

tigarmo (commented):
the approach seems to work, thanks. I added some comments about naming but they don't affect the meat of the PR

craft_parts/executor/part_handler.py
upils: I do no like where the comment is either but this call to `organize_files` is not only in the `if` branch but more importantly under the `overlays.LayerMount` context because it needs the overlay to be mounted.  I will create a method calling `self._run_step` and `organize_files` to avoid the duplication.

craft_parts/parts.py
upils: Yes. I struggled finding proper names for some methods because they are doing very specific things.

PR #1193, opened 2025-07-30
feat: exclude files already distributed to another partition

upils:

-----

### DIFF
--- a/craft_parts/executor/part_handler.py
+++ b/craft_parts/executor/part_handler.py
@@ -103,6 +103,7 @@ class _Squasher:
         self.migrated_directories: dict[str | None, _MigratedContents] = {partition: {}}
         self._src_partition = partition
         self._default_partition = default_partition
+        self._sub_paths_distributed: set[str] = set()
         if filesystem_mount:
             self._filesystem_mount = filesystem_mount
 
@@ -126,6 +127,7 @@ class _Squasher:
                 # entry
                 sub_path = entry.mount.lstrip("/")
                 dst_partition = entry.device
+                logger.debug(f"distribute content to {dst_partition}")
 
                 # Migrate to the destination partition indicated by the filesystem mounts entry
                 self._migrate(
@@ -134,6 +136,8 @@ class _Squasher:
                     sub_path=sub_path,
                     dst_partition=dst_partition,
                 )
+                if len(sub_path) > 0:
+                    self._sub_paths_distributed.add(sub_path)
         else:
             # Ignore the filesystem mounts and migrate from/to the same partition
             self._migrate(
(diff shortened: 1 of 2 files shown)


craft_parts/executor/part_handler.py
```diff
@@ -134,6 +136,8 @@ def migrate(
                     sub_path=sub_path,
                     dst_partition=dst_partition,
                 )
+                if len(sub_path) > 0:
+                    self._sub_paths_distributed.add(sub_path)
```
cmatsuoka, line 140: If `sub_path` is not changed after creation we can move this conditional to a point right after creation. Then `sub_paths_distributed` will actually list all (non-empty) subpaths I presume?

craft_parts/executor/part_handler.py
```diff
@@ -187,6 +194,26 @@ def _migrate(
...
             self.migrated_directories[dst_partition][src_path] = f
 
+    def _filter_already_distributed(
+        self, visible_files: set[str], visible_dirs: set[str]
+    ) -> tuple[set[str], set[str]]:
+        """Exclude files in sub path already distributed to another partition."""
+        files: set[str] = set()
+        dirs: set[str] = set()
+        if not self._sub_paths_distributed:
+            return visible_files, visible_dirs
```
cmatsuoka, line 204: We can also move this block to the top of this method, before declaring `files` and `dirs`.

craft_parts/executor/part_handler.py
```diff
@@ -187,6 +194,26 @@ def _migrate(
...
+        for f in visible_files.copy():
+            for d in self._sub_paths_distributed:
+                if not Path(f).is_relative_to(Path(d)) or Path(f) == Path(d):
```
cmatsuoka, line 207: This checks if the visible file was not already distributed (because it's under a distributed subpath, or it's the subpath itself). Maybe adding a small utility function `_not_already_distributed()` could make this easier to read/understand.

craft_parts/executor/part_handler.py
```diff
@@ -187,6 +194,26 @@ def _migrate(
...
+                    logger.debug(f"keeping file {f}")
+                    files.add(f)
+        for f in visible_dirs.copy():
+            for d in self._sub_paths_distributed:
+                if not Path(f).is_relative_to(Path(d)) or Path(f) == Path(d):
```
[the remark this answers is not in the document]
cmatsuoka: It seems that this is a false positive? We're building a list of files that were not already distributed.

craft_parts/executor/part_handler.py
cmatsuoka: A comment paragraph explaining what we're trying to achieve here can be useful.

craft_parts/executor/part_handler.py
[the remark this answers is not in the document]
upils: I think the solution proposed by Copilot would work but may not be more readable in the end (using a `else`). So I think I will stick to the current approach.

mr-cal (approved):

PR #1262, opened 2025-09-08
feat: allow creating usrmerged dirs by default

tigarmo:
This commit introduces a "usrmerged_by_default" parameter to the LifecycleManager so that applications can opt into a new behavior where a part's install dir gets an usrmerged directory structure by default. This lets the lifecycle create a set of files that is better suited for modern Ubuntu bases, but is a breaking change for some plugins and specially for user-provided "override-build" scripts.

Two build-attributes are also introduced to let users have finer, per-part control:

- "enable-usrmerge" can be used to have the directory structure created for the part even if "usrmerged_by_default" is False;
- "disable-usrmerge" can be used to disable the structure creation even if "usrmerged_by_default" is True.

-----

### DIFF
--- a/craft_parts/parts.py
+++ b/craft_parts/parts.py
@@ -1059,6 +1059,19 @@ def sort_parts(part_list: list[Part]) -> list[Part]:
     # simplest way to do this is to sort them by name.
     all_parts = sorted(part_list, key=lambda part: part.name, reverse=True)
 
+    # Change the implicit order so that parts that organize to them
+    # are at the end of the list (because the order is reversed).
+    organize_to_overlay_parts: list[Part] = []
+    other_parts: list[Part] = []
+    for part in all_parts:
+        if part.organizes_to_overlay:
+            organize_to_overlay_parts.append(part)
+        else:
+            other_parts.append(part)
+
+    all_parts = [*other_parts, *organize_to_overlay_parts]
+
+    # Process explicit ordering set using the "after" key.
     while all_parts:
         top_part = None
 
--- a/craft_parts/sequencer.py
+++ b/craft_parts/sequencer.py
@@ -121,7 +121,7 @@ class Sequencer:
                     rerun_target_step=rerun_target_step,
                 )
 
-    def _add_step_actions(
+    def _add_step_actions(  # noqa: PLR0912
         self,
         *,
         current_step: Step,
@@ -141,6 +141,14 @@ class Sequencer:
         # check if step already ran, if not then run it
         if not self._sm.has_step_run(part, current_step):
             self._run_step(part, current_step, reason=reason)
+
+            # Parts that organize to the overlay run the BUILD step immediately
+            # after the OVERLAY step. Organization happens at the end of the
+            # BUILD step, and organized files may be needed by another part that
+            # installs overlay packages. See ST-158.
+            if part.organizes_to_overlay and current_step == Step.OVERLAY:
+                self._run_step(part, Step.BUILD, reason="organize contents to overlay")
+
             return
 
         # If the step has already run:
@@ -194,6 +202,15 @@ class Sequencer:
                     outdated_files=outdated_files,
                     outdated_dirs=outdated_dirs,
                 )
+
+                if part.organizes_to_overlay and current_step == Step.OVERLAY:
+                    self._update_step(
+                        part,
+                        Step.BUILD,
+                        reason="organize contents to overlay",
+                        outdated_files=outdated_files,
+                        outdated_dirs=outdated_dirs,
+                    )
             else:
                 self._rerun_step(part, current_step, reason=outdated_report.reason())
 
--- a/craft_parts/sources/local_source.py
+++ b/craft_parts/sources/local_source.py
@@ -76,6 +76,7 @@ class LocalSource(SourceHandler):
             self._ignore_patterns.append(self._dirs.parts_dir.name)
             self._ignore_patterns.append(self._dirs.stage_dir.name)
             self._ignore_patterns.append(self._dirs.prime_dir.name)
+            self._ignore_patterns.append(self._dirs.overlay_dir.name)
             if self._dirs.partition_dir:
                 self._ignore_patterns.append(self._dirs.partition_dir.name)
         else:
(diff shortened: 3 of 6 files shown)


upils (approved):
I have the same concern as Claudio but otherwise LGTM. Maybe it should be tested in a one of our *craft application before merging though.

craft_parts/executor/part_handler.py
```diff
@@ -1175,6 +1175,43 @@ def _symlink_alias_to_default(self) -> None:
...
+
+        usrmerged_by_default = self._part_info.usrmerged_by_default
+        enabled_attr = "enable-usrmerge" in self._part_info.build_attributes
+
+        usrmerged = usrmerged_by_default or enabled_attr
+        if not usrmerged:
+            # usrmerged not enabled by default, nor for the individual
+            return
+
+        root_dir = self._part.part_install_dirs[self._part_info.default_partition]
```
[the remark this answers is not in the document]
upils: It maps to `workdir/parts/<part_name>/install` so with the magic of symlinks it will end up in the correct place if the default partition is renamed, and is the correct place if not.

docs/reference/changelog.rst
```diff
@@ -31,6 +31,12 @@ New features:
   pseudo-partition. To make use of this feature, the project must support both overlays
   and partitions.
 
+- Add a new LifecycleManager creation parameter to control whether parts should have a
+  "usrmerged" directory structure pre-created in their install directories.
+
+- Add two new build-attributes, "enable-usrmerge" and "disable-usrmerge", to control the
+  creation of the "usrmerged" directory structure in a part's install dir.
```
medubelko, line 38:
```suggestion
- Add a new ``LifecycleManager`` creation parameter to control whether parts should have 
  a merged ``/usr`` directory scheme in their install directories.

- Add two new build attributes, ``enable-usrmerge`` and ``disable-usrmerge``, to control 
  the creation of a merged ``/usr`` directory scheme in a part's install directory.
```

medubelko, line 38: I feel like in some quarters this will be seen as a _big_ change, but I'm not really sure how best to promote it in Craft Parts itself. It will certainly earn a major release note in our apps.

craft_parts/executor/part_handler.py
```diff
@@ -1193,6 +1230,7 @@ def _make_dirs(self) -> None:
             os.makedirs(dir_name, exist_ok=True)
 
         self._symlink_alias_to_default()
+        self._create_usrmerge_scaffolding()
```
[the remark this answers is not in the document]
tigarmo: Claudio gave me a longer version of his comment that is a compelling argument: it's quite common to have parts that just `dump` some files and then `organize` the whole thing into a specific subdir, like this:

```yaml
config-files:
  plugin: dump
  source: config/
  organize:
    "*": etc/config
```

That is, the assumption that only files dumped by the plugin exist in the part's install dir is likely very strong and ingrained. In this scenario, `etc/config` would also get the usrmerged scaffolding, like `etc/config/usr/bin`.

When we (Claudio, Paul and I) talked about this earlier we decided to *disable* this behavior for `nil` and `dump` at least initially. Since this is only going to be used in a development base for now (in Rockcraft), we can consider this feature "socially experimental" and iterate on it a bit.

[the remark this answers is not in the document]
lengau: @tigarmo it'll still be enabled if you add the build attribute though, right?

docs/reference/changelog.rst
tigarmo: thanks, applied it manually

tigarmo: Definitely. The plan is this: Rockcraft will be the first app to enable the new default behavior, only for rocks targeting base `ubuntu@25.10` and later. This gives us some time while the base is still unstable to experiment with the feature, and then once the 26.04 LTS rolls around this should be solid and the impact reasonable well-known.

This will feature prominently in Rockcraft's release notes regarding 25.10 support.

tigarmo:
> @tigarmo Per our discussion, let's add `build-attributes` as a mainstay key to _Part properties_, and list the UsrMerge options in a table à la [Snapcraft](https://documentation.ubuntu.com/snapcraft/stable/reference/project-file/snapcraft-yaml/#parts.%3Cpart-name%3E.build-attributes).
> 
> I've tasked us for updating Snapcraft itself in [canonical/snapcraft#5769](https://github.com/canonical/snapcraft/issues/5769).

sorry I missed the "table" part. Can I do tables in the docstrings?

craft_parts/parts.py
```diff
@@ -337,7 +337,19 @@ class PartSpec(BaseModel):
         examples=['[{MESSAGE: "Hello world!"}, {NAME: "Craft Parts"}]'],
     )
 
-    build_attributes: list[str] = []
+    build_attributes: list[str] = Field(
+        default=[],
+        description="Identifiers that control specific behaviors during the build.",
+        examples=["[enable-usrmerge]", "[disable-usrmerge]"],
+    )
+    """Special identifiers that change some features and behaviors during the build.
```
medubelko, line 345: So I'm clear, we don't want to say "build step"?

craft_parts/parts.py
```diff
@@ -337,7 +337,19 @@ class PartSpec(BaseModel):
...
+    Possible values are:
+
+    * ``enable-usrmerge``: Pre-fill the ``${CRAFT_PART_INSTALL}`` directory with a basic
+      usrmerged directory structure before running the part's build step.
+    * ``disable-usrmerge``: Do *not* pre-fill ``${CRAFT_PART_INSTALL}`` wih the usrmerged
+      directory structure.
```
medubelko, line 351: If we can, let's follow the pattern from the spec:

```suggestion
    **Values**

    .. list-table::
        :header-rows: 1

        * - Value
          - Description
        * - ``enable-usrmerge``
          - Fills the ``${CRAFT_PART_INSTALL}`` directory with a merged ``/usr`` 
            directory before running the part's build step.
        * - ``disable-usrmerge``
          - Prevents a merged ``/usr`` directory from being assembled for the build 
            step. Available in lifecycles in which the directory would be merged by
            default.
```

I also tweaked the wording.

medubelko (commented):
Thanks for adding the field!

PR #1275, opened 2025-09-10
ci: separate out java tests

lengau:

---


.github/workflows/qa.yaml
```diff
@@ -9,14 +9,48 @@ on:
...
+      setup-vars: NO_PYTHON=1 NO_PLUGIN=1
+  test-python-plugins:
+    uses: canonical/starflow/.github/workflows/test-python.yaml@work/setup-vars
+    with:
+      fast-test-platforms: "[]"
+      slow-test-platforms: '["ubuntu-24.04"]'
+      slow-test-python-versions: '["3.12"]'
+      lowest-python-version: ""
+      pytest-markers: python
+      setup-vars: NO_JAVA=1 NO_PLUGIN=1
```
upils, line 47: thought: I wonder how hard that will be to maintain without getting it wrong at some point, especially if we want to split another group of test out of the common group. We will need to change the setup-vars of every job. However I do not see a better solution :/.

.github/workflows/qa.yaml
```diff
@@ -9,14 +9,48 @@ on:
...
+  cancel-in-progress: true
+
 jobs:
   lint:
     uses: canonical/starflow/.github/workflows/lint-python.yaml@main
   test:
-    uses: canonical/starflow/.github/workflows/test-python.yaml@main
+    uses: canonical/starflow/.github/workflows/test-python.yaml@work/setup-vars
     with:
       fast-test-platforms: '["ubuntu-22.04", "ubuntu-24.04"]'
```
upils, line 23: issue: that is probably unrelated to this PR but I noticed that tests on plugins are still executed in "fast tests" jobs, see https://github.com/canonical/craft-parts/actions/runs/17602626438/job/50008521822?pr=1275#step:4:782. Unless I am missing something, I suspect passed markers are not respected, or not properly passed to the `make test-coverage` command.

.github/workflows/qa.yaml
```diff
@@ -9,14 +9,48 @@ on:
...
       fast-test-python-versions: '["3.10", "3.12", "3.13"]'
       slow-test-platforms: '["ubuntu-24.04"]'
       slow-test-python-versions: '["3.12"]'
       lowest-python-version: ""
+      pytest-markers: not java and not python and not plugin
+      setup-vars: NO_JAVA=1 NO_PYTHON=1 NO_PLUGIN=1
+  test-java-plugins:
```
upils, line 30: thought: I know we do not focus much on coverage, but splitting tests result in split coverage. Aside from the issue mentioned above, I wonder if we should still keep a job running everything to produce a consolidated coverage. Or we can collect the coverage files and consolidate them before uploading to codecov.

.github/workflows/qa.yaml
```diff
@@ -9,14 +9,48 @@ on:
...
 
+# Cancel existing jobs for this branch/PR
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref_name }}
+  cancel-in-progress: true
+
 jobs:
   lint:
     uses: canonical/starflow/.github/workflows/lint-python.yaml@main
   test:
```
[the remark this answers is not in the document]
lengau: I like `test-common` this. Thanks!
```suggestion
  test-common:
```

.github/workflows/qa.yaml
lengau: I made it opt-out to preserve the UX for the more important use case of running setup on one's own machine. If we forget an opt out when changing CI, the only real downside is the installation of additional unnecessary packages.

tests/integration/plugins/test_poetry.py
```diff
@@ -57,6 +57,7 @@ def parts_dict(poetry_part):
     return {"parts": {"foo": poetry_part}}
 
 
+@pytest.mark.python
```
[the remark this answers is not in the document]
lengau: Playing with it, what do you think of this fixtures in the plugins directory's `conftest.py`? This would ensure that we run all of the plugin integration tests with all feature combinations without having to copy them into the features directory (which I've noticed some of our plugins aren't in anyway).

If a test needs to run under a specific feature set, the test can request that fixture and skip if the feature set is incorrect. (This also means the test could, for example, check only for partitions and would run regardless of the overlay state)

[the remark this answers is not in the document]
lengau: It's not, because each `import` statement simply overrides what `pytestmark` does.

bepri (approved):

PR #1276, opened 2025-09-10
ci: run Python plugin tests on more platforms

lengau:

---

Requires #1275


upils:
Oh I am not complaining that we have a lot of tests! I compared the time needed to run the whole test suite in CI before and after these changes and the added time is negligible I think, so it is fine. 

However I am worried about our blind spots. I find it increasingly difficult to determine if we are testing a specific combination of python version / ubuntu version / arch / plugin group other than actually looking at an executed CI workflow. This is fine for now but might become hard to maintain at some point.

PR #1294, opened 2025-09-12
style: enable strict type checking with pyright

bepri:

---

Enables strict type checking with pyright and then resolves all of the resulting messages. This is part of a series of PRs to improve the typing and linting status of Craft Parts to be more on-par with our other libraries.

This branch has been installed and had linters ran on {Charm,Rock,Snap}craft and Craft Application.


craft_parts/overlays/chroot.py
```diff
@@ -43,6 +49,7 @@ def chroot(path: Path, target: Callable, *args: Any, **kwargs: Any) -> Any:  # n
     :returns: The target function return value.
     """
     logger.debug("[pid=%d] parent process", os.getpid())
+    parent_conn: Connection[Any, tuple[_T, str | None]]
```
bepri, line 52: This typehint technically should be `Connection[Any, tuple[_T, None] | tuple[None, str]]`, but types surrounding multiprocessing are finnicky at best and the way we handle it now makes this type hint effectively true, since we don't try to access the first field of the tuple if the second is `str`.

craft_parts/plugins/plugins.py
```diff
@@ -82,7 +82,7 @@
     "uv": UvPlugin,
 }
 
-_PLUGINS = copy.deepcopy(_BUILTIN_PLUGINS)
+_Plugins = copy.deepcopy(_BUILTIN_PLUGINS)
```
bepri, line 85: Pyright didn't like that we were modifying an all-caps variable.

craft_parts/state_manager/build_state.py
```diff
@@ -48,7 +48,7 @@ def unmarshal(cls, data: dict[str, Any]) -> "BuildState":
 
         :raise TypeError: If data is not a dictionary.
         """
-        if not isinstance(data, dict):
+        if not isinstance(data, dict):  # pyright: ignore[reportUnnecessaryIsInstance]
```
bepri, line 51: I feel like these could potentially be removed, but I wanted to leave that investigation up to another PR

craft_parts/state_manager/state_manager.py
```diff
@@ -455,7 +458,8 @@ def get_step_state_overlay_hash(self, part: Part, step: Step) -> bytes:
             # step didn't run yet
             return b""
 
-        overlay_hash = stw.state.overlay_hash  # type: ignore[reportGeneralTypeIssues, attr-defined]
+        state = cast("stage_state.StageState | build_state.BuildState", stw.state)
```
bepri, line 461: Already asserted by the above `if step not in ...` check, but (understandably) pyright didn't know that

craft_parts/errors.py
```diff
@@ -185,19 +191,17 @@ def from_validation_error(
         return cls(part_name=part_name, message="\n".join(formatted_errors))
 
     @classmethod
-    def _format_loc(cls, loc: "Loc") -> str:
+    def _format_loc(cls, loc: tuple[int | str, ...]) -> str:
```
bepri, line 194: `Loc` isn't exported anymore by Pydantic, but this is its type internally.

craft_parts/pydantic_schema.py
```diff
@@ -76,8 +76,8 @@ def model_json_schema(
             pydantic.ConfigDict(extra="allow"),
         ]
 
-        source_adapter: pydantic.TypeAdapter = pydantic.TypeAdapter(sources.SourceModel)
-        plugin_adapter: pydantic.TypeAdapter = pydantic.TypeAdapter(plugin_model)
+        source_adapter = pydantic.TypeAdapter(sources.SourceModel)  # type: ignore[reportUnknownVariableType, var-annotated]
```
bepri, line 79: These types weren't effective because we're doing something really weird here. Generating types at runtime like this is hated by Pydantic, mypy, and Pyright (a triple whammy!) so I just ignore it completely instead.

craft_parts/utils/file_utils.py
```diff
@@ -69,8 +69,7 @@ def write(self, data: str) -> int:
 
     def close(self) -> None:
         """Close the FIFO."""
-        if self._fd is not None:
-            os.close(self._fd)
+        os.close(self._fd)
```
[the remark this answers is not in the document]
bepri: I don't see any reason for this change, as it would just let you call `.close()` multiple times on a FIFO. But we can accept it if somebody sees a strong benefit.

craft_parts/callbacks.py
```diff
@@ -19,33 +19,39 @@
...
+
+
+@dataclass(frozen=True)
+class CallbackHook(Generic[_T_cb_co]):
+    """A callback hook."""
+
+    function: _T_cb_co
+    step_list: list[Step] | None
+
+
```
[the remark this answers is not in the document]
bepri: The change from a `NamedTuple` was necessary as Python [doesn't allow generic `NamedTuples`](https://bugs.python.org/issue43923). To preserve the API, I set it to a frozen `dataclass`, as this should still give the same constructors and immutable properties.

craft_parts/permissions.py
```diff
@@ -204,18 +204,17 @@ def _squash_permissions(permissions: list[Permissions]) -> Permissions:
...
-        for key in keys:
+    squashed_permissions = Permissions(
+        path=permissions[-1].path,
+        owner=permissions[0].owner,
+        group=permissions[0].group,
+        mode=permissions[0].mode,
+    )
+
+    for permission in permissions[1:]:
+        for key in ["path", "owner", "group", "mode"]:
```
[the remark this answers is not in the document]
bepri: This doesn't change anything, it just skips the first iteration of the loop.

craft_parts/errors.py
```diff
@@ -168,10 +171,13 @@ def from_validation_error(
         formatted_errors: list[str] = []
 
         for error in error_list:
-            loc = error.get("loc")
-            msg = error.get("msg")
+            try:
+                loc = error["loc"]
+                msg = error["msg"]
+            except KeyError:
+                continue
```
cmatsuoka, line 178: We could use `get` to default to `None` and try-catch would be unnecessary here, slightly improving readability. The conditional below would handle the absence of keys.

craft_parts/errors.py
```diff
@@ -837,19 +841,22 @@ def __init__(
...
```
cmatsuoka, line 857: Is a loc value of `None` allowed here? Because if the key exists and the value is `None` the conditional below won't catch it. If it's the same case as the previous similar block we could use `get` instead.

craft_parts/overlays/chroot.py
cmatsuoka: Is this worth mentioning in the code itself as a comment?

craft_parts/utils/file_utils.py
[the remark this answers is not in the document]
cmatsuoka: Setting the descriptor to `None` doesn't seem necessary, but check if not `None` still looks useful to avoid a crash.

craft_parts/plugins/plugins.py
lengau:
```suggestion
_plugins = copy.deepcopy(_BUILTIN_PLUGINS)
```

craft_parts/utils/file_utils.py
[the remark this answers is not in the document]
cmatsuoka: If we're guaranteed to have a value there I'm good with removing the check.

craft_parts/callbacks.py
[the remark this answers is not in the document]
lengau: Also if anyone was using the `CallbackHook`'s iterability I am deeply concerned.

craft_parts/state_manager/state_manager.py
bepri: sadly, nope

cmatsuoka (approved):
Thanks! Here's your badge:

image

PR #1295, opened 2025-09-12
docs: add more readme flair

lengau:
We need to talk about your flair.

# We were only using the minimum amount of flair.

What do you think of a person who only does the bare minimum?

Rendered: https://github.com/canonical/craft-parts/blob/work/more-flair/README.md

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ before it is packaged into the final artifact.
 
 This library is most useful for app developers in the Starcraft family.
 
-## [![Documentation Status][rtd-badge]][rtd-latest]
+## Documentation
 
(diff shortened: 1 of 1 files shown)


medubelko (commented):
And in the proper spirit of the 90s, it's like I'm looking at the site footer on Geocities.

I'm going to be the devil's advocate, assuming we want this to become a pattern across our projects. It's not what I had in mind with the README model.

Do we need _all_ of these? There are several that from what I can tell don't provide a status, feel like TF2 hats.

I feel there ought to be a rational order to these, at minimum. Something like `license` > `builds` > `security checks` > `build checks`.

There's no way that download count is an honest metric, right?

PR #1311, opened 2025-09-16
ci: test non java/python plugins on more platforms

lengau:

---

Hey look! There are [Jammy-based ARM runners](https://github.com/actions/partner-runner-images#available-images) now!

NOTE: I don't plan to make the additional workflows required unless I get pushback on that. I'm more interested in "notice that this doesn't work" than "we need to let this finish running before we merge your PR because you fixed a typo in the changelog" for this stuff, so it would be on the person doing the merge to decide about failing tests that don't meet our core requirements. I definitely won't be making s390x or ppc64el required until those are no longer marked as experimental (I don't want CI to fail because there aren't any of those runners available), but I'm happy to discuss our feelings on focal amd64 runners and jammy/noble arm64 runners (I can see arguments both ways for them).


tests/integration/plugins/test_cargo_use/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "test_cargo_use"
 version = "0.1.0"
-edition = "2024"
+edition = "2018"
```
lengau, line 4: So we can build with cargo 1.75 on focal. Ensuring that the base has the relevant cargo edition is out of scope.

tests/integration/plugins/test_go.py
```diff
@@ -124,6 +126,10 @@ def test_go_generate(new_dir, partitions):
     assert output == "This is a generated line\n"
 
 
+@pytest.mark.skipif(
```
lengau, line 129: This is solvable in the makefile, but it's become a big enough hassle that it deserves its own ticket.

lengau:
Flaky tests are resolved in #1308 and #1316

tests/integration/plugins/test_cargo_use/Cargo.toml
bepri: I'm not sure if you're saying this should be done in another PR or not, so I thought I'd put a small argument here - I think it's completely out of Craft Parts' responsibility to ensure the relevant Cargo version, a bit like how it isn't Craft Parts' job to make sure you aren't using Python `match` statements while building a focal snap.

I'd say this is a fine change because the Rust maintainers are veeeery good about backwards compatibility and I can't imagine this ever not working.

lengau: No, this is just a change to enable running this test with older cargo.

cmatsuoka (approved):

PR #1347, opened 2025-09-30
feat: Enforce filesystem mounts to be ordered in increasing mount nesting

### LINKED ISSUE #1342: Force order of entries in FilesystemMount
### What needs to get done

Implement a validator on `FilesystemMount` checking:
- the order of entries in based on depth of the `mount` value
- uniqueness of `mount` values

### Why it needs to get done

In the `filessystems` section (see Imagecraft project file), entries can currently be defined in an arbitrary order, as long as the first one maps to `/`. This is leading to configurations producing surprising results.

Enforce the following rules:
- Entries must be ordered in increasing depth.
- `mount` values must be unique (preventing 2 devices to be mounted at the same location)

upils:

---

Fixes #1342 
CRAFT-4803

### DIFF
--- a/craft_parts/filesystem_mounts.py
+++ b/craft_parts/filesystem_mounts.py
@@ -17,6 +17,7 @@
 """FilesystemMounts models."""
 
 from collections.abc import Iterable, Iterator
+from pathlib import Path
 from typing import Annotated, Any, Literal, cast
 
 from pydantic import (
@@ -86,6 +87,16 @@ class FilesystemMountItem(BaseModel):
         return self.model_dump(by_alias=True)
 
 
+def _format_conflicts(conflicts: list[tuple[str, str]]) -> str:
+    """Format conflicts between FilesystemMountItems."""
+    formatted: list[str] = [
+        f"- {misplaced} must be listed before {parent}"
+        for misplaced, parent in conflicts
+    ]
+
+    return "\n".join(formatted)
+
+
 class FilesystemMount(
     RootModel[Annotated[UniqueList[FilesystemMountItem], Field(min_length=1)]]
 ):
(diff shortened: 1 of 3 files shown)


docs/reference/changelog.rst
```diff
@@ -21,6 +21,10 @@ Changelog
 2.25.0 (unreleased)
 -------------------
 
+New features:
+
+- Enforce filesystem mounts to be ordered in increasing ``mount`` depth.
```
```suggestion
- Order filesystem mounts by increasing ``mount`` depth.
```

Just a nitpick, not required.

medubelko (approved):

docs/reference/changelog.rst
tigarmo: How about something more spelled out, like "The LifecycleManager will now validate that filesystem mounts are ordered in increasing 'mount' depth"?

craft_parts/filesystem_mounts.py
```diff
@@ -107,6 +118,36 @@ def first_maps_to_slash(
             raise ValueError("The first entry in a filesystem must map the '/' mount.")
         return value
 
+    @field_validator("root", mode="after")
+    @classmethod
+    def validate_entries_order(
```
tigarmo, line 123: I think the code itself is fine but is there a reason to do this in craft-parts, as opposed to in Imagecraft itself? If I remember the flow correctly, if we do the validation only in craft-parts it means that it will only be done when the LifecycleManager is created, right? 

Or is this model also used in Imagecraft directly?

docs/reference/changelog.rst
upils: It looks better, thanks! But I realized that the "ordered by `depth`" is an oversimplification because the following would be valid:
- /
- /a
- /a/foo
- /b
- /b/bar

`/a/foo` is deeper than `/b` but is not conflicting, and the code properly deal with that.

So I will rework this sentence to clarify that.

PR #1370, opened 2025-10-15
fix: make sure the destination does not exist before organizing

### LINKED ISSUE #1366: Unhandled error when organizing a file to a directory where the file already exists
### Bug Description

The `organize_files` function in craft-parts does not properly detect collisions when the destination is a directory. If an entry organize a file to a destination and then a second entry also tries to organize to the same destination but expressed as a directory, it fails because the call to `file_utils.move(src,dst)` fails. However this error is not handled and is "swallowed".

The function should probably check if the destination exists when `overwrite` is `False` before calling `file_utils.move(src,dst)`.

### To Reproduce

Running `python3 -mcraft_parts  --trace -v prime` with this project file fails with `Error: None`.

### part yaml

```shell
parts:
  rootfs:
    plugin: nil
    override-build: |
      mkdir $CRAFT_PART_INSTALL/foo
      mkdir $CRAFT_PART_INSTALL/bar
      echo "a from foo" > $CRAFT_PART_INSTALL/foo/a
      echo "a from bar" > $CRAFT_PART_INSTALL/bar/a
    organize:
      'bar/a': /qux/a
      'foo/a': /qux/
```

### Relevant log output

```shell
Execute: Pull rootfs
Execute: Overlay rootfs
Execute: Build rootfs
+ mkdir /home/paul/scratch/craft-parts/parts/rootfs/install/foo
+ mkdir /home/paul/scratch/craft-parts/parts/rootfs/install/bar
+ echo 'a from foo'
+ echo 'a from bar'
Error: None.
```

upils:

---

Fixes #1366
CRAFT-4870

### DIFF
--- a/docs/reference/changelog.rst
+++ b/docs/reference/changelog.rst
@@ -27,10 +27,9 @@ Bug fixes:
 - Trim down debug logs to make the output more useful when handling numerous
   overlay files.
 - Suppress detached HEAD warnings when cloning Git remotes.
-- `#1366 <https://github.com/canonical/craft-parts/issues/1366>`_ When the destination
-  of an ``organize`` entry is a directory, properly detect collisions with existing
-  files in this directory. A proper error is now returned, giving necessary information
-  to the user to resolve the conflict.
+- `#1366 <https://github.com/canonical/craft-parts/issues/1366>`_ When organizing a
+  file to a directory, detect collisions with existing files in the directory. If a
+  collision occurs, an error is produced with a message listing the colliding paths.
 
 For a complete list of commits, check out the `2.26.0`_ release on GitHub.
 


docs/reference/changelog.rst
```diff
@@ -27,6 +27,10 @@ Bug fixes:
 - Trim down debug logs to make the output more useful when handling numerous
   overlay files.
 - Suppress detached HEAD warnings when cloning Git remotes.
+- `#1366 <https://github.com/canonical/craft-parts/issues/1366>`_ When the destination
+  of an ``organize`` entry is a directory, properly detect collisions with existing
+  files in this directory. A proper error is now returned, giving necessary information
+  to the user to resolve the conflict.
```
jahn-junior:
```suggestion
- `#1366 <https://github.com/canonical/craft-parts/issues/1366>`_ When organizing a file
  to a directory, detect collisions with existing files in the directory. If a collision
  occurs, an error is produced with a message listing the colliding paths.
```

jahn-junior (approved):

upils:
> This looks good in general, but there's a part of me that wonders whether it could cause re-run issues. Do we have any integration tests that would cover that case?

@lengau in case of a rerun the build step is either skipped, updated or cleaned and re-executed.

If it is updated, the `overwrite` param is set to true and the previous branch (just before my additional change) is executed, removing the existing file. https://github.com/canonical/craft-parts/blob/4c4c228d7d65f1d712b3b80e502f86f91201d5b4/tests/unit/features/partitions/executor/test_organize.py#L203 is testing that.

If it is cleaned, the existing file is also removed.

So I do not see a case where it could prevent from reruning.

PR #1383, opened 2025-10-29
fix: ignored outdated files should still be pulled

### LINKED ISSUE #1373: `ignore_outdated` glob list for parts
### What needs to get done

Add an `ignore_outdated` list that works like `ignore_local_sources` for the purpose of checking whether a local part is outdated, but still copies the files into the build directory.

### Why it needs to get done

Blocker for https://github.com/canonical/craft-application/issues/942

cmatsuoka:
During the execution of the lifecycle there are files or directories
that should not be pulled, such as the `parts` or `stage`. There are
also files that should be pulled and ignored by the outdated file
checker, such as `spread.yaml`. If we don't pull files that are in
the repository parts that set `version` using `git describe --dirty`
will always describe the state of the repository as dirty.

---
CRAFT-4884
Fixes #1373

### DIFF
--- a/craft_parts/sources/git_source.py
+++ b/craft_parts/sources/git_source.py
@@ -270,6 +270,21 @@ class GitSource(SourceHandler):
 
     def _clone_new(self) -> None:
         """Clone a git repository, using submodules, branch, and depth if defined."""
+        git_version = self._get_git_version()
+
+        # Attempt a shallow fetch for source_commits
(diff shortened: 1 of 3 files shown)


cmatsuoka:
@lengau This is a very simple fix, could you check if this actually solves the problem you detected?

lengau (approved):
Seems to work for me. Thanks!

PR #1395, opened 2025-11-12
feat: add ruby plugin 

### LINKED ISSUE #338: Add ruby plugin
Port from Snapcraft PR: https://github.com/snapcore/snapcraft/pull/3511

steinbro:

---
Fixes #338. The goal is to update the abandoned PR #341 as follows:

1. Don't recompile Ruby every time, i.e. allow for a ruby-deps part that can be shared by other parts
2. Add tests
3. Add docs

### DIFF
--- a/craft_parts/plugins/plugins.py
+++ b/craft_parts/plugins/plugins.py
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright 2021,2024 Canonical Ltd.
+# Copyright 2021,2024-2025 Canonical Ltd.
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
@@ -16,7 +16,8 @@
 
 """Definitions and helpers to handle plugins."""
 
-import copy
+import enum
+from collections.abc import Mapping
 from typing import TYPE_CHECKING, Any
 
 from craft_parts import errors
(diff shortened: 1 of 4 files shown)


steinbro:
Implementation, tests and docs are ready for a first round of review.

craft_parts/plugins/ruby_plugin.py
```diff
@@ -0,0 +1,207 @@
...
+    - ``ruby-configure-options``
+      (list of str)
+      Defaults to []
+    - ``ruby-use-bundler``
+      (bool)
+      Defaults to False
+    """
+
+    # NOTE: To update ruby-install version, go to https://github.com/postmodern/ruby-install/tags
+    RUBY_INSTALL_VERSION = "0.10.1"
```
cmatsuoka: We usually define this type of constant at top level, local to the module.

craft_parts/plugins/ruby_plugin.py
```diff
@@ -0,0 +1,207 @@
...
+    ruby_shared: bool = False
+    ruby_configure_options: list[str] = []
+
+
+class RubyPlugin(Plugin):
+    """A plugin for Ruby based projects.
+
+    The desired Ruby interpreter is compiled using ruby-install.
+
+    The ruby plugin uses the common plugin keywords, plus the following ruby-
```
medubelko: > common plugin keywords

A minor point, but there aren't really any common keys for plugins. Plugin keys are parallel with part keys, so they feel like they're part of the same collection, but there's no standard set, AFAIK.

docs/common/craft-parts/reference/plugins/ruby_plugin.rst
```diff
@@ -0,0 +1,125 @@
...
+
+This plugin provides the following unique keys.
+
+
+ruby-gems
+~~~~~~~~~
+
+**Type:** list of strings
+
+List of gems to be installed.
```
medubelko:
```suggestion
The gems to install.
```

docs/common/craft-parts/reference/plugins/ruby_plugin.rst
```diff
@@ -0,0 +1,125 @@
...
+ruby-flavor
+~~~~~~~~~~~
+
+**Type:** string
+
+**Default:** ruby
+
+Specifies which implementation of the Ruby interpreter the plugin
+should build. Supported values are ``ruby``, ``jruby``, ``rbx``,
+``truffleruby``, and ``mruby``.
```
medubelko:
```suggestion
The Ruby interpreter to compile and include. The supported interpreters are:

- ``ruby``
- ``jruby``
- ``rbx``
- ``truffleruby``
- ``mruby``
```

I don't believe we need to keep the word "implementation", but I'm not very familiar.

docs/common/craft-parts/reference/plugins/ruby_plugin.rst
```diff
@@ -0,0 +1,125 @@
...
+
+ruby-use-jemalloc
+~~~~~~~~~~~~~~~~~
+
+**Type:** boolean
+
+**Default:** False
+
+When set to ``true``, the plugin compiles Ruby with support for the jemalloc
+memory allocator.
```
medubelko:
```suggestion
Whether to compile Ruby with support for jemalloc.
```

docs/common/craft-parts/reference/plugins/ruby_plugin.rst
```diff
@@ -0,0 +1,125 @@
...
+.. _ruby-details-begin:
+
+Dependencies
+------------
+
+By default, this plugin downloads and runs the `ruby-install
+<https://github.com/postmodern/ruby-install>` tool. If recompiling the
+interpreter for every Ruby-based part is not desired, a ``ruby-deps`` part
+can install Ruby and any gems in the build system. Other parts that use this
+interpreter must run ``after`` the ``ruby-deps`` part:
```
medubelko:
```suggestion
<https://github.com/postmodern/ruby-install>` tool.

If a project has multiple Ruby parts, by default each compiles and bundles its
own Ruby interpreter. If a special Ruby part named ``ruby-deps`` is defined, it
creates a shared interpreter and shared gems that the other Ruby parts can use.
For a Ruby part to use these shared files, it must list ``ruby-deps`` in its
``after`` key.
```

medubelko (commented):
@steinbro I went into a little more detail than I usually do, so I hope this isn't too much to chew on. Let me know what you think.

docs/common/craft-parts/reference/plugins/ruby_plugin.rst
```diff
@@ -0,0 +1,125 @@
...
+When set to ``true``, the plugin compiles Ruby with support for the jemalloc
+memory allocator.
+
+
+ruby-configure-options
+~~~~~~~~~~~~~~~~~~~~~~
+
+**Type:** list of strings
+
+Extra arguments to pass to the compiler when building the Ruby interpreter.
```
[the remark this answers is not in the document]
steinbro: I would expect so, since each is an entirely different implementation with its own optional features.. There are some examples in the ruby-install README: https://github.com/postmodern/ruby-install/blob/master/README.md#synopsis

docs/common/craft-parts/reference/plugins/ruby_plugin.rst
steinbro: Thank you for a much better explanation. My attempt to adapt language from the poetry plugin reference did not turn out well.

bepri:
Do we know which bash script is failing, exactly? My first guess is some sort of `bash`/`sh` inequality

steinbro:
I worked around the inscrutable syntax error by switching the integration test to compile the mruby variant of the Ruby interpreter, which uses Ruby's own Rakefile syntax instead of Makefiles. The only failing tests now are unrelated, coming from a different plugin (dotnet_v2).

tests/integration/plugins/test_ruby.py
```diff
@@ -0,0 +1,107 @@
...
+    # from bundle install
+    ruby_root = Path(lf.project_info.prime_dir, "ruby")
+    # e.g. "3.2.0"; will vary based on version in archive
+    version_dir = next(ruby_root.iterdir())
+    primed_script = version_dir / "bin" / "mytest"
+    assert primed_script.exists()
+
+
+@pytest.mark.slow
+def test_ruby_custom_flavor(new_dir, partitions):
```
lengau: Could we get a version of this test that doesn't set any configuration options?

steinbro:
Reworked to match the behavior of other plugins: by default, don't attempt to build an interpreter and raise an exception if none is present in the build environment. Also re-requesting docs review from @medubelko since I updated the Dependencies section of the plugin reference accordingly.

docs/common/craft-parts/reference/plugins/ruby_plugin.rst
```diff
@@ -0,0 +1,128 @@
...
+interpreter.
+
+
+.. _ruby-details-begin:
+
+Dependencies
+------------
+
+The Ruby plugin needs the ``ruby`` interpreter to run Ruby programs but does not
+provision it by default, to allow flexibility in the choice of interpreter flavor
```
medubelko:
```suggestion
provide it by default, to give you flexibility in the choice of interpreter flavor
```

I think "provide" is a less expensive word for what we mean here.

docs/common/craft-parts/reference/plugins/ruby_plugin.rst
```diff
@@ -0,0 +1,128 @@
...
+Extra arguments to pass to the ``configure`` script when building the Ruby
+interpreter.
+
+
+.. _ruby-details-begin:
+
+Dependencies
+------------
+
+The Ruby plugin needs the ``ruby`` interpreter to run Ruby programs but does not
```
medubelko: > `ruby` interpreter

Nitpick: do we really need to specify the binary name here and in the next paragraph? Doesn't "Ruby interpreter" mean the same thing?

docs/common/craft-parts/reference/plugins/ruby_plugin.rst
steinbro: I agree and I'll accept this suggestion, but FWIW this language came straight from the Go plugin reference.

PR #1414, opened 2025-12-08
fix: standardize paths in ruby plugin

### LINKED ISSUE #1413: Ruby plugin paths are inconsistent
### Bug Description

The Ruby plugin as implemented in #1395 treats paths inconsistently:
1. The install prefix for the interpreter is `$CRAFT_PART_INSTALL/usr`.
2. `GEM_HOME` is `$CRAFT_PART_INSTALL`.
3. `BUNDLE_PATH` is `$CRAFT_PART_INSTALL`.

This means:
1. The ruby executable is in `/usr/bin/`, and system-global gems are in ``/usr/lib/ruby/gems/3.2.0/`.
2. Gem-installed executables are in `/bin`
3. Bundle-installed executables are in `/ruby/<ruby_version>/bin`

Instead, we should set the build environment so that `gem install` and `bundle install` go into the system gem path.

### To Reproduce

N/A

### part yaml

```shell

```

### Relevant log output

```shell
N/A
```

steinbro:
FIxes #1413.

---

### DIFF
--- a/tests/integration/plugins/test_ruby.py
+++ b/tests/integration/plugins/test_ruby.py
@@ -1,6 +1,6 @@
 # -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
 #
-# Copyright 2023-2025 Canonical Ltd.
+# Copyright 2025 Canonical Ltd.
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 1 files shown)


steinbro:
I addressed the items above, which required more changes than expected. Still need to fix the hard-coded Ruby version in the RUBYLIB search paths. Also need to address the failing tests on older Ubuntu bases, where bundler from rubygems isn't compatible with the ruby version from apt.

lengau (approved):

PR #1418, opened 2025-12-18
docs: update overlay filter field description

### LINKED ISSUE #1401: Docs: Rewrite `overlay` field description
## Situation

In canonical/rockcraft/pull/1041, we're transcluding the docs of the `overlay` key for the first time.

The key's documentation needs improvement and additional context:

https://github.com/canonical/craft-parts/blob/f2227aac531a1de6cf454081717c3256b2663462/craft_parts/parts.py#L387

## Request

- [ ] Fix "overly" typo.
- [ ] Remove "overlay filesystem" as a term, as it's unhelpful to users.
- [ ] Rewrite the description so that it informs how the user will use the key to productive effect:
    - Describe what happens when the list is set at all. It's an _exhaustive_ list.
    - Describe the path negation syntax.
- [ ] Review the second example. Why would `-etc/...` be needed when it automatically excludes everything that's not listed?

cmatsuoka:
Closes #1401.

---

### DIFF
--- a/craft_parts/parts.py
+++ b/craft_parts/parts.py
@@ -272,13 +272,10 @@ class PartSpec(BaseModel):
         description="The packages to install in the part's layer.",
         examples=["[ed]"],
     )
-    """The packages to install in the part's overlay filesystem.
+    """The packages to install in the part's layer.
 
-    During the overlay step, these packages are installed into the part's overlay
-"""The packages to install in the part's layer.
-
-   During the overlay step, these packages are installed into the part's layer
-   using the base layer's package manager.
+    During the overlay step, these packages are installed into the part's layer
+    using the base layer's package manager.
     """
 
     stage_snaps: list[str] = Field(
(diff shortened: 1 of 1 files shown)


craft_parts/parts.py
```diff
@@ -384,9 +384,18 @@ class PartSpec(BaseModel):
...
     )
+    """During the overlay step, any files specified in this field will remain in the part's
+    layer, or be removed if preceded by ``-``. This operation applies only to files that
+    are present in the part's layer and files in lower layers will not be affected. If a
+    file is removed and a file with the same name exists in the same path in a lower layer,
+    the latter will be copied to the stage directory instead.
+
+    Paths support wildcards (``*``) and must be relative to the working directory where
+    they will be used.
+    """
```
jahn-junior:
```suggestion
    """The files to copy from the part's layer to the stage directory.
    
    During the overlay step, files listed under this key are kept in the
    part’s layer unless prefixed with ``-``, which removes them. Any
    files left in the part's layer are copied to the stage directory during
    the stage step.
    
    This operation only applies to files that are present in the part's layer
    — files in lower layers aren't affected. If a file is removed and a lower
    layer contains a file with the same path, the latter will be copied to
    the stage directory.

    Paths support wildcards (``*``) and must be relative to the working
    directory where they will be used.
    """
```
The pattern we've established is to reiterate the `description`, as the docstring takes precedent. I also tried to clarify the relationship between the part's layer and the stage directory, as there wasn't any mention of it beyond the description. Let me know if you think this could be improved.

(I really wish we could link to the filesets explanation, but it'd be too fragile downstream.)

craft_parts/parts.py
jahn-junior:
```suggestion
        description="The packages to install in the part's layer.",
```

jahn-junior (approved):
Thanks!

bepri (approved):

PR #1421, opened 2026-01-05
fix(maven-use): infer groupId from parent

bepri:

---

Fixes a bug where valid Maven projects that inherited a `groupId` from a `parent` tag instead of defining their own would be rejected by the Maven Use plugin.

CRAFT-4928.


bepri:
OSV warning is addressed in #1411

PR #1449, opened 2026-02-06
fix: Use `UV_LOCKED` instead of `UV_FROZEN`

carlcsaposs-canonical:
`UV_FROZEN` ensures uv.lock is not modified https://docs.astral.sh/uv/reference/cli/#uv-sync--frozen
`UV_LOCKED` ensures uv.lock is not modified & validates that uv.lock is a valid solution to the current pyproject.toml https://docs.astral.sh/uv/reference/cli/#uv-sync--locked

Fixes https://github.com/canonical/charmcraft/issues/2564

### DIFF
--- a/tests/integration/plugins/test_uv/pyproject.toml
+++ b/tests/integration/plugins/test_uv/pyproject.toml
@@ -6,6 +6,7 @@ build-backend = "setuptools.build_meta"
 name = "plugin_test_uv"
 description = "My package description"
 version = "0.0.1"
+requires-python = ">=3.10"
 
 [project.scripts]
 mytestuv = "mytestuv:main"
--- a/tests/integration/plugins/test_uv/uv.lock
+++ b/tests/integration/plugins/test_uv/uv.lock
@@ -1,6 +1,6 @@
 version = 1
 revision = 3
-requires-python = ">=3.12"
+requires-python = ">=3.10"
 
 [[package]]
 name = "plugin-test-uv"


docs/common/craft-parts/reference/plugins/uv_plugin.rst
```diff
@@ -47,7 +47,7 @@ own special variables.
     variables for uv, see the `uv environment documentation
     <https://docs.astral.sh/uv/configuration/environment/>`_.
 
-UV_FROZEN
+UV_LOCKED
```
bepri: The description under this header isn't accurate anymore

carlcsaposs-canonical: I wasn't sure what to put here—do you have any suggestions?

For reference, here are the upstream descriptions:
UV_FROZEN
> Run without updating the `uv.lock` file
>
> Instead of checking if the lockfile is up-to-date, uses the versions in the lockfile as the source of truth. If the lockfile is missing, uv will exit with an error. If the `pyproject.toml` includes changes to dependencies that have not been included in the lockfile yet, they will not be present in the environment.

https://docs.astral.sh/uv/reference/cli/#uv-run--frozen

UV_LOCKED
> Assert that the `uv.lock` will remain unchanged
>
> Requires that the lockfile is up-to-date. If the lockfile is missing or needs to be updated, uv will exit with an error.


https://docs.astral.sh/uv/reference/cli/#uv-run--locked

And the current description in the charmcraft docs:
> Whether or not to update the `uv.lock` file. If true, `uv.lock` must exist and will be used as the single source of truth for dependency versions, with no attempt made to update them before installation.

I noticed that the description in the charmcraft docs is a bit different from upstream—and was wondering if that's intentional (since the user is not interfacing directly with the `uv sync` command). I believe the current description is also fairly accurate for UV_LOCKED—if set to `false`, the uv.lock file will be updated

Should I remove the reference to "single source of truth" or do you think the description should be changed further?

PR #1452, opened 2026-02-09
feat: Add `UV_LINK_MODE=copy`

carlcsaposs-canonical:
Companion to https://github.com/canonical/craft-providers/pull/900

When the cache dir is mounted from the LXD host, hardlinking from the global cache (uv's default behavior) is not possible

e.g.
```
:: warning: Failed to hardlink files; falling back to full copy. This may lead to degraded performance.
::          If the cache and target directories are on different filesystems, hardlinking may not be supported.
::          If this is intentional, set `export UV_LINK_MODE=copy` or use `--link-mode=copy` to suppress this warning.
```


bepri (approved):
Thanks! Could you make a changelog entry for this and update the reference docs for the plugin with a sentence or two explaining why it's set by us?

bepri (approved):
Thanks - just minor nitpicks on the docs entries now.

medubelko (approved):

mr-cal (approved):
LGTM but would like to get @lengau's review, since this is a charmcraft-motivated feature.

docs/common/craft-parts/reference/plugins/uv_plugin.rst
```diff
@@ -93,6 +93,17 @@ UV_PYTHON_PREFERENCE
 Whether uv should prefer (or exclusively use) system or uv-managed Python versions.
 
 
+UV_LINK_MODE
+~~~~~~~~~~~~
+
+**Type:** ``clone``, ``copy``, ``hardlink``, or ``symlink``
+
+**Default:** "copy"
```
lengau: Setting this to copy by default will slow down the default case where the cache isn't mounted from the outer machine. While I think it's a good idea to document this here and provide a motivation for using `copy`, I don't think we should change the default value.

lengau (changes requested):
Let's leave the default value unset still.

docs/common/craft-parts/reference/plugins/uv_plugin.rst
lengau: That depends on the user's LXD configuration. If their storage type is `dir` and it's on the same filesystem as the host, I believe it'll still do hardlinks correctly.

In the intervening time though I've changed my opinion regardless.

lengau (approved):

PR #1471, opened 2026-02-23
ci: upgrade flask

smethnani:

---

Resolves an osv. An alternative could be to skip test directories for osv scanning


bepri (approved):
I agree with ignoring these in the future. I think the best way to approach this would be to add the file to Starbase at `tests/osv-scanner.toml` and then merge that into each of the repositories, but we probably have some out of date repos where we can't "just" merge

PR #1474, opened 2026-02-25
docs: fix Ruby plugin docs examples

steinbro:

---
The Ruby plugin examples in the docs were broken (thanks @atandrewlee for pointing this out):

1. The `ruby-deps` part was missing the `source` property.
2. The `mruby` interpreter flavor does not support gems

The docs now present separate examples for building a custom Ruby interpreter, and for creating a `ruby-deps` part.

### DIFF
--- a/docs/common/craft-parts/reference/plugins/ruby_plugin.rst
+++ b/docs/common/craft-parts/reference/plugins/ruby_plugin.rst
@@ -157,7 +157,6 @@ are included in the output artifact.
   parts:
     ruby-deps:
       plugin: nil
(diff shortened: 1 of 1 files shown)


docs/common/craft-parts/reference/plugins/ruby_plugin.rst
medubelko: I didn't catch this the first time around, but I think this would be better with an _Examples_ section, with introductory text for what each example shows. With two now it's not clear what they do and how they're related to dependencies.

PR #1485, opened 2026-02-28
fix(deb): avoid python-apt for installed package checks

### LINKED ISSUE #976: Package installer fails if `/etc/apt/sources.list` is absent
### Bug Description

On a fresh Ubuntu Plucky install, `/etc/apt/sources.list` is not included, as Ubuntu has migrated to fully utilising deb822. However, craft-parts

### To Reproduce

Clone the repository on a fresh plucky machine and run the cmake integration tests. It does not matter whether cmake is installed - it just needs any plugin that has a build dependency.

### part yaml

```shell
n/a
```

### Relevant log output

```shell
(craft-parts) lengau@ratel:~/Work/Code/craft-parts$ pytest tests/integration/plugins/test_cmake.py::test_cmake_plugin[-bin]
========================================================================================= test session starts =========================================================================================
platform linux -- Python 3.11.10, pytest-8.3.4, pluggy-1.5.0
rootdir: /home/lengau/Work/Code/craft-parts
configfile: pyproject.toml
plugins: subprocess-1.5.3, cov-5.0.0, requests-mock-1.12.1, check-2.4.1, hypothesis-6.123.7, mock-3.14.0
collected 1 item

tests/integration/plugins/test_cmake.py F                                                                                                                                                       [100%]

============================================================================================== FAILURES ===============================================================================================
_______________________________________________________________________________________ test_cmake_plugin[-bin] _______________________________________________________________________________________

new_dir = local('/tmp/pytest-of-lengau/pytest-15/test_cmake_plugin__bin_0'), partitions = None, prefix = '', install_path = 'bin'

@pytest.mark.parametrize(
("prefix", "install_path"),
[
(None, "usr/local/bin"),
("", "bin"),
("/usr/local", "usr/local/bin"),
("/something/else", "something/else/bin"),
],
)
def test_cmake_plugin(new_dir, partitions, prefix, install_path):
parts_yaml = textwrap.dedent(
"""\
parts:
foo:
plugin: cmake
source: .
"""
)
if prefix is not None:
parts_yaml += f"    cmake-parameters: [-DCMAKE_INSTALL_PREFIX={prefix}]"

Path("cmake.build").write_text(
textwrap.dedent(
"""\
project('cmake-hello', 'c')
executable('hello', 'hello.c', install : true)
"""
)
)

Path("hello.c").write_text(
textwrap.dedent(
"""\
#include <stdio.h>

int main()
{
printf(\"hello world\\n\");
return 0;
}
"""
)
)

Path("CMakeLists.txt").write_text(
textwrap.dedent(
"""\
cmake_minimum_required(VERSION 2.6)
project(cmake-hello C)
add_executable(cmake-hello hello.c)
install(TARGETS cmake-hello RUNTIME DESTINATION bin)
"""
)
)

lf = LifecycleManager(
parts, application_name="test_cmake", cache_dir=new_dir, partitions=partitions
)
actions = lf.plan(Step.PRIME)

>       with lf.action_executor() as ctx:

/home/lengau/Work/Code/craft-parts/tests/integration/plugins/test_cmake.py:88:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/home/lengau/Work/Code/craft-parts/craft_parts/executor/executor.py:304: in __enter__
self._executor.prologue()
/home/lengau/Work/Code/craft-parts/craft_parts/executor/executor.py:88: in prologue
self._install_build_packages()
/home/lengau/Work/Code/craft-parts/craft_parts/executor/executor.py:246: in _install_build_packages
packages.Repository.install_packages(sorted(build_packages))
/home/lengau/Work/Code/craft-parts/craft_parts/packages/deb.py:562: in install_packages
if not cls._check_if_all_packages_installed(package_names):
/home/lengau/Work/Code/craft-parts/craft_parts/packages/deb.py:293: in wrapped
return method(*args, **kwargs)
/home/lengau/Work/Code/craft-parts/craft_parts/packages/deb.py:466: in _check_if_all_packages_installed
with AptCache() as apt_cache:  # pyright: ignore[reportPossiblyUnboundVariable]
/home/lengau/Work/Code/craft-parts/craft_parts/packages/apt_cache.py:103: in __enter__
self.cache = apt.cache.Cache(rootdir="/")
/home/lengau/Work/Code/craft-parts/.venv/lib/python3.11/site-packages/apt/cache.py:148: in __init__
self._check_and_create_required_dirs(rootdir)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <apt.cache.Cache object at 0x7a99f22c1fd0>, rootdir = '/'

def _check_and_create_required_dirs(self, rootdir: str) -> None:
"""
check if the required apt directories/files are there and if
not create them
"""
files = [
"/var/lib/dpkg/status",
"/etc/apt/sources.list",
]
dirs = [
"/var/lib/dpkg",
"/etc/apt/",
"/var/cache/apt/archives/partial",
"/var/lib/apt/lists/partial",
]
for d in dirs:
if not os.path.exists(rootdir + d):
# print "creating: ", rootdir + d
os.makedirs(rootdir + d)
for f in files:
if not os.path.exists(rootdir + f):
>               open(rootdir + f, "w").close()
E               PermissionError: [Errno 13] Permission denied: '//etc/apt/sources.list'

/home/lengau/Work/Code/craft-parts/.venv/lib/python3.11/site-packages/apt/cache.py:188: PermissionError
======================================================================================= short test summary info =======================================================================================
FAILED tests/integration/plugins/test_cmake.py::test_cmake_plugin[-bin] - PermissionError: [Errno 13] Permission denied: '//etc/apt/sources.list'
========================================================================================== 1 failed in 0.11s ==========================================================================================
```

gcomneno:
Fixes #976

## Background

On Ubuntu Plucky, `/etc/apt/sources.list` may be absent because the system uses deb822-based configuration exclusively.

When `python-apt` initializes its cache (`apt.cache.Cache(rootdir="/")`), it attempts to ensure certain files exist, including `/etc/apt/sources.list`.
If the file is missing, it attempts to create it. When running as an unprivileged user (e.g. during integration tests), this results in:

PermissionError: [Errno 13] Permission denied: '//etc/apt/sources.list'

This failure happens in the installed-package check path, before any actual installation occurs.

## What this change does

For the host `install_packages()` path only:

- Replace the use of `python-apt` for:
  - checking whether packages are already installed
  - constructing the returned `package=version` list
- Use `dpkg-query` instead for installed-state detection.

Actual installation still uses:

    apt-get install ...

Stage package handling and other apt-cache–based functionality remain unchanged.

## Rationale

For determining whether a package is installed, `dpkg-query` is:

- sufficient
- does not require root
- independent of apt source configuration
- unaffected by deb822 migration

Using `python-apt` for this specific optimization is unnecessary and causes failures on systems without `/etc/apt/sources.list`.

## Trade-offs

- `dpkg-query` does not resolve virtual packages in the same way as `python-apt`.
- However:
  - `apt-get install` still performs dependency and virtual package resolution.
  - The returned `package=version` list is already documented as “best effort”.
  - Typical build-packages (cmake, gcc, pkg-config, etc.) are concrete packages.

## Scope

This change only affects the installed-check optimization and version reporting for host installs.
All apt-cache usage for stage packages remains intact.


mr-cal (changes requested):
Thanks, this looks like a nice improvement.

mr-cal (commented):
Thanks, this is looking good. Can you run `make format` and `make lint` locally and resolve the linting errors?

https://github.com/canonical/craft-parts/actions/runs/22774342100/job/66068884396?pr=1485#step:6:8

mr-cal (commented):
I think you'll need to run make format and make lint again.

craft_parts/packages/deb.py
```diff
@@ -544,25 +650,36 @@ def install_packages(
...
             install_required = True
 
-        # Collect the list of marked packages to later construct a manifest
-        marked_packages = cls._get_packages_marked_for_installation(package_names)
-        marked_package_names = [name for name, _ in sorted(marked_packages)]
+        if list_only:
+            try:
+                marked = _get_packages_marked_for_installation_apt_get(package_names)
+            except subprocess.CalledProcessError as err:
+                raise errors.BuildPackageNotFound(package_names[0]) from err
```
mr-cal: Won't this just show the first package? How will we know it's the erroneous package?

gcomneno:
Hi! Following up on this one.

I know things were busy around the 26.04 lead-up, so no rush — I just wanted to check whether there has been any chance to revisit it, or if there is anything else you’d like me to update on my side.

gcomneno:
Hi! Just gently following up on this one.

As far as I can tell, the review comments have been addressed in the current branch, including the installed-packages manifest path and the removal of the no-longer-used helper. The PR is currently green on CI.

Could you please take another look when convenient? Happy to adjust anything else if I missed something, or to close it if this direction should not move forward.

craft_parts/packages/deb.py
```diff
@@ -284,6 +284,8 @@
 
 _T_wrapper = TypeVar("_T_wrapper")
 
+_DPKG_QUERY_INSTALLED_FIELDS = 3
```
mr-cal: Can you drop the const? If the linters flag it for a magic number, then an # ignore comment.

craft_parts/packages/deb.py
```diff
@@ -370,6 +372,135 @@ def _is_list_of_slices(names: list[str]) -> bool:
...
+
+
+def _get_installed_packages_dpkg_query() -> list[str]:
+    """Return installed packages using dpkg-query.
+
+    Prefer dpkg-query because python-apt may try to create
+    /etc/apt/sources.list when it is missing, which fails in unprivileged
+    environments. Fall back to parsing /var/lib/dpkg/status when dpkg-query
+    is unavailable or fails.
+    """
```
mr-cal: I think this is a dupe of the docstring for get_installed_packages
```suggestion
    """Return installed packages using dpkg-query."""
```

craft_parts/packages/deb.py
```diff
@@ -370,6 +372,135 @@ def _is_list_of_slices(names: list[str]) -> bool:
...
+
+
+def _get_packages_marked_for_installation_apt_get(
+    package_names: list[str],
+) -> list[tuple[str, str]]:
+    """Return packages apt-get would install using simulation output."""
+    result = subprocess.run(
+        [
+            "apt-get",
+            "-s",
```
mr-cal: For readability:
```suggestion
            "--simulate",
```

craft_parts/packages/deb.py
```diff
@@ -370,6 +372,135 @@ def _is_list_of_slices(names: list[str]) -> bool:
...
+            "install",
+            *package_names,
+        ],
+        check=True,
+        capture_output=True,
+        text=True,
+    )
+
+    marked: list[tuple[str, str]] = []
+    for line in result.stdout.splitlines():
```
mr-cal: I don't think this parser will work there is an upgrade. Consider:

```
apt-get --simulate --no-install-recommends --allow-downgrades install systemd
...
Inst systemd [255.4-1ubuntu8.14] (255.4-1ubuntu8.15 Ubuntu:24.04/noble-updates [amd64])
...
```

The split would grab the old version, not the new one.  Can you add unit tests for this?

craft_parts/packages/deb.py
mr-cal, line 1: Would you be able to add a simple integration test to `tests/integration/packages/test_deb.py`?  I think these commands would benefit from an integration test.

gcomneno:
Thanks for the review!

I pushed follow-up fixes for the requested changes:

- replaced the apt simulation call with `apt-get --simulate`
- fixed the simulation parser so upgrade lines report the target version rather than the currently installed version
- improved `BuildPackageNotFound` handling so it reports the package mentioned by apt-get output when detectable, instead of always falling back to the first requested package
- simplified the dpkg-query helper code by dropping the field-count constant and shortening the duplicated docstring
- added unit coverage for upgrade simulation parsing and apt-get reported-package errors
- added a simple integration test for Debian installed-package queries

Local checks pass:

- `uv run pytest -q tests/unit/packages/test_deb.py tests/integration/packages/test_deb.py -rs`
- `uv run ruff check craft_parts/packages/deb.py tests/unit/packages/test_deb.py tests/integration/packages/test_deb.py`
- `uv run ruff format --check craft_parts/packages/deb.py tests/unit/packages/test_deb.py tests/integration/packages/test_deb.py`
- `git diff --check`

tests/integration/packages/test_deb.py
```diff
@@ -0,0 +1,27 @@
...
+
+from craft_parts.packages import deb
+
+
+def test_get_installed_packages_returns_system_packages():
+    """Verify installed Debian packages can be queried without python-apt."""
+    packages = deb.Ubuntu.get_installed_packages()
+
+    assert packages
+    assert all("=" in package for package in packages)
```
mr-cal: Can we verify the contents a little better, like this?

```python
assert all(                                                                                                                                                          
    len(parts) == 2 and all(parts)                                                                                                                                   
    for pkg in packages                                                                                                                                              
    for parts in [pkg.split("=", 1)]                                                                                                                                 
)
````

File: craft_parts/packages/deb.py [hunk 7]
```diff
@@ -544,25 +662,38 @@ def install_packages(
         if not cls._check_if_all_packages_installed(package_names):
             install_required = True
 
-        # Collect the list of marked packages to later construct a manifest
-        marked_packages = cls._get_packages_marked_for_installation(package_names)
-        marked_package_names = [name for name, _ in sorted(marked_packages)]
+        if list_only:
```
mr-cal, line 782: Unrelated, but it turns out this we have no known usages of `list_only`.  I wonder if this is some forgotten use-case that was replaced by `get_installed_packages`.

craft_parts/packages/deb.py
```diff
@@ -435,7 +578,6 @@ def refresh_packages_list(  # pyright: ignore[reportIncompatibleMethodOverride]
             ) from call_error
 
     @classmethod
-    @_apt_cache_wrapper
     def _check_if_all_packages_installed(cls, package_names: list[str]) -> bool:
```
mr-cal, line 676: Can you add a comment to this docstring about how this doesn't resolve virtual packages? 

I believe the impact of this is that it might make a redundant apt-get call (i.e. if you need gcc, it won't see that gcc is installed and will call `apt-get install gcc`. It will be a no-op because gcc-13 is installed.)

I'm OK with this impact in favor of not making this PR more complex.

craft_parts/packages/deb.py
```diff
@@ -544,25 +662,38 @@ def install_packages(
...
-                cls._install_packages(package_names)
-            else:
-                logger.debug(
-                    "Requested build-packages already installed: %s", package_names
-                )
-
-        # This result is a best effort approach for deps and virtual packages
-        # as they are not part of the installation list.
-        # Tell static type checkers to ignore until we can use PEP 612 (Python 3.10)
-        return cls._get_installed_package_versions(  # type: ignore[no-any-return]
```
mr-cal, line 663: This is interesting too - there's no known usages of this return value being consumed.

craft_parts/packages/deb.py
```diff
@@ -544,25 +662,38 @@ def install_packages(
...
+                raise errors.BuildPackageNotFound(failed_package) from err
+            cls._install_packages(package_names)
+            installed_package_names = [name for name, _ in marked]
+        else:
+            logger.debug(
+                "Requested build-packages already installed: %s", package_names
+            )
+            installed_package_names = [
+                get_pkg_name_parts(name)[0] for name in package_names
+            ]
```
mr-cal: I think this means the return of this function depends on whether we are installing packages or not.

What if we take the lead for the original function, where we always call `marked = _get_packages_marked_for_installation_apt_get(package_names)` and always use that in the return?

Also, can you keep the comment saying the return is best effort? I'd like to deprecate the return value in the future, since we aren't using it and it only makes this code more confusing.

mr-cal (commented):
Looking good, just some minor feedback.  Can you add another integration tests for me?  I'd like to see the `install_packages` get exercised.

I asked an LLM to write this test, so feel free to adjust it. It at least shows the intention of what I want to test and the hello package is a good choice:

```python
import subprocess
import pytest
from craft_parts.packages import deb


@pytest.fixture
def purge_hello():
   """Ensure hello is not installed before the test and clean up after."""
   subprocess.run(
       ["apt-get", "purge", "-y", "--autoremove", "hello"],
       check=False,
       capture_output=True,
   )
   yield
   subprocess.run(
       ["apt-get", "purge", "-y", "--autoremove", "hello"],
       check=False,
       capture_output=True,
   )


@pytest.mark.requires_root
def test_install_packages_installs_package_and_returns_version(purge_hello):
   """Verify install_packages installs a package and reports its version."""
   installed = deb.Ubuntu.install_packages(["hello"])

   assert any(pkg.startswith("hello=") for pkg in installed)
   assert Path("/usr/bin/hello").is_file()


@pytest.mark.requires_root
def test_install_packages_already_installed_returns_version(purge_hello):
   """Verify install_packages is idempotent and still returns a version."""
   deb.Ubuntu.install_packages(["hello"])
   deb.Ubuntu.refresh_packages_list.cache_clear()

   installed = deb.Ubuntu.install_packages(["hello"])

   assert any(pkg.startswith("hello=") for pkg in installed)
```

gcomneno:
Thanks, I addressed the feedback in `b264a11b`.

Changes:
- added root integration coverage for `Ubuntu.install_packages(["hello"])`, covering both fresh install and already-installed/idempotent cases;
- tightened the `get_installed_packages()` integration assertion to validate non-empty `package=version` entries;
- documented that the `dpkg-query` installed-package check does not resolve virtual packages;
- now always computes the `apt-get --simulate` marked set before returning from `install_packages()`;
- restored the best-effort return-value comment.

One note: `apt-get --simulate install hello` does not emit an `Inst hello` line when `hello` is already installed, so the return path also preserves the legacy concrete requested-package version reporting for the already-satisfied case. The new idempotency integration test covers that behavior.

Local validation:
- `uv run ruff format --check craft_parts/packages/deb.py tests/integration/packages/test_deb.py tests/unit/packages/test_deb.py`
- `uv run ruff check craft_parts/packages/deb.py tests/integration/packages/test_deb.py tests/unit/packages/test_deb.py`
- `uv run pytest -q tests/unit/packages/test_deb.py tests/integration/packages/test_deb.py -rs`
- `sudo env PATH="$PATH" VIRTUAL_ENV="$VIRTUAL_ENV" .venv/bin/python -m pytest -q tests/integration/packages/test_deb.py -m requires_root -rs`

CI is still finishing, but the current failing check is the existing `Security scan / OSV-scanner` policy job.

craft_parts/packages/deb.py
gcomneno: Addressed in `7eba9714`. `install_packages()` now consistently derives the returned manifest from the packages reported by the apt simulation, regardless of whether an actual installation is needed. I removed the branch that added already-satisfied requested packages back into the manifest and kept the existing best-effort comment. I also added a regression test for the already-satisfied / empty-marked-set case. `tests/unit/packages/test_deb.py` passes locally (74 passed).

craft_parts/packages/deb.py
gcomneno: Addressed in the current HEAD. `_check_if_all_packages_installed()` now documents that the dpkg-query check does not resolve virtual packages and explains that this can lead to a redundant `apt-get install` call that resolves to a no-op when the dependency is already satisfied. Thanks for calling this out.

craft_parts/packages/deb.py
mr-cal: FYI this was removed in #1671. Can you drop it here too?

craft_parts/packages/deb.py
```diff
@@ -524,65 +673,44 @@ def refresh_packages_list(  # pyright: ignore[reportIncompatibleMethodOverride]
...
     def _check_if_all_packages_installed(cls, package_names: list[str]) -> bool:
         """Check if all given packages are installed.
 
         Will check versions if using <pkg_name>=<pkg_version> syntax parsed by
         get_pkg_name_parts().  Used as an optimization to skip installation
         and cache refresh if dependencies are already satisfied.
 
+        This check uses dpkg-query and does not resolve virtual packages. If a
+        virtual package is requested, apt-get may still make a later install
+        call that resolves to a no-op when the dependency is already satisfied.
```
mr-cal, line 683: 👍

mr-cal (commented):
Hi @gcomneno, sorry for the delay in reviews.. This is looking great and very readable compared to the first draft. I have one last piece of feedback before I approve.

I'm also going to kick off another copilot review.

gcomneno:
Addressed your `list_only` request in the latest commit (`5fb6e7d1`).

I also fixed the installed-package status check to handle held packages (`hold ok installed`) consistently, with focused regression coverage.

Validation: 114 unit tests passed, and the current PR checks are green.

mr-cal:
@gcomneno, can you re-resolve the merge conflicts? Github is still complaining.

gcomneno:
The QA run is green except for `test-other-plugins / Slow tests (noble, s390x, edge, 3.12)`.

The failure occurs in the Bazel plugin test: Bazel exits with error code 14 (`Socket closed`) while starting the Bazel server on the s390x runner.

All other QA jobs passed, including the minimum-dependencies suite. This does not appear related to the Debian package-check changes in this PR.

mr-cal (approved):
Thanks for your all your hard work here! I can confirm this fixes #976 for me. This is also a good step towards eliminating craft-parts dependency on `python-apt`.

mr-cal (commented):
@gcomneno, I think your latest merge broke a few unit tests.

craft_parts/packages/deb.py
```diff
@@ -444,6 +444,165 @@ def get_packages_in_base(*, base: str) -> list[DebPackage]:
...
+
+
+def _get_apt_get_error_package(
+    package_names: list[str], err: subprocess.CalledProcessError
+) -> str:
+    """Return the package name mentioned by apt-get, if detectable."""
+    output = f"{err.stdout or ''}\n{err.stderr or ''}"
+    for package_name in package_names:
+        pkg_name, _ = get_pkg_name_parts(package_name)
+        if package_name in output or pkg_name in output:
```
cmatsuoka: This matching could fail if a package name is a substring of another package name (like `dbus` and `libdbus-1-3` or `python3-dbus`).

cmatsuoka (changes requested):

craft_parts/packages/deb.py
gcomneno: Fixed in the latest commit. The apt-get error matcher now requires Debian package-name boundaries instead of using plain substring matching, so names such as `dbus` no longer match `python3-dbus` or `libdbus-1-3`. I added regression coverage for both cases. Focused validation: 4 tests passed, Ruff passed, and `git diff --check` passed.

cmatsuoka (approved):
Thanks. Maybe there's a better way to do this type of matching but this one should work as expected!

mr-cal:
@gcomneno, can you run `make format`? I think ruff is failing: https://github.com/canonical/craft-parts/actions/runs/33416929495/job/99569815103?pr=1485#step:7:12

gcomneno:
Done, thanks for catching that. I ran `make format`; it only reformatted the new assertion in `tests/unit/packages/test_deb.py`. I then re-ran the focused tests, Ruff lint, Ruff format check, and `git diff --check`, all successfully.

mr-cal:
Thanks again, @gcomneno - this was a tough undertaking!

PR #1535, opened 2026-04-04
feat: organize files from build

cmatsuoka:
Add support for organizing files directly from a part’s build directory
via the `(build)` pseudo-partition. Files cannot be organized into the
build directory.

Also improve destination handling for trailing slashes so non-existing
`.../` destinations are treated as directories and created.

Fixes #1521 
CRAFT-5071

---


tests/unit/features/overlay_partitions/test_executor_organize.py
```diff
@@ -51,10 +51,12 @@ def test_organize(new_dir, data):
         setup_dirs=data.get("setup_dirs", []),
         setup_files=data.get("setup_files", []),
         setup_symlinks=data.get("setup_symlinks", []),
+        build_files=data.get("build_files", []),
```
lengau: Could you add a success test case here where we're organizing from the build dir to the overlay?

Likewise, can you check whether organizing *from* the overlay correctly errors?

lengau (changes requested):
Code LGTM, but I think we might need a few other test cases.

tests/unit/features/overlay_partitions/test_executor_organize.py
cmatsuoka: Test cases added.

lengau (approved):

mr-cal (commented):
I was just testing this out and realized - shouldn't we reserve `build` from being used as a partition name?

You can't organize into a `build` partition anymore, so craft-parts shouldn't allow a `build` partition. We already have a precedent and a place to put the validation, so it should be simple: https://github.com/canonical/craft-parts/blob/e4c55752d2f0e26a9c1c6097e47912d50fce23ad/craft_parts/utils/partition_utils.py#L79-L81


Here's an example with how confusing `build` is as a partition:
```python
import logging
import yaml

import craft_parts
from craft_parts import LifecycleManager, Step

logging.basicConfig(level=logging.DEBUG)

craft_parts.Features(enable_partitions=True)

parts_yaml = """
parts:
  hello:
    plugin: nil
    source: .
    override-build:
      touch $CRAFT_PART_BUILD/foo
      touch $CRAFT_PART_BUILD/bar
      craftctl default
    organize:
      # is the user trying to organize from the build directory or from the build partition?
      (build)/foo: (other)/foo
      # this line fails:
      (build)/bar: (build)/baz
 """

lcm = LifecycleManager(
    parts,
    application_name="example",
    cache_dir=".",
    partitions=["default", "build", "other"],
)

actions = lcm.plan(Step.PRIME)

with lcm.action_executor() as aex:
    aex.execute(actions)
```

mr-cal (approved):
LGTM, thanks!

lengau (changes requested):

lengau (approved):

PR #1541, opened 2026-04-09
fix(organize): make organizing to the overlay idempotent

### LINKED ISSUE #1488: Organizing to the overlay directory is not idempotent
### Bug Description

This may have the same root cause as #1487.

Parts that organize to the overlay cannot be re-built without erroring when trying to organize them.

### To Reproduce

1. Copy the part YAML below to `parts.yaml`
2. `mkdir /tmp/empty`
3. `sudo $(which python) -m craft_parts --verbose --overlay-base=/tmp/empty --partitions=default,other build`
4. Uncomment the line to create subdir-b
5. `sudo $(which python) -m craft_parts --verbose --overlay-base=/tmp/empty --partitions=default,other build`

### part yaml

```shell
parts:
  part-a:
    plugin: nil
    override-build: |
      mkdir -p $CRAFT_PART_INSTALL/my-dir/subdir-a
      # mkdir -p $CRAFT_PART_INSTALL/my-dir/subdir-b
    organize:
      '*': (overlay)/
```

### Relevant log output

```shell
Execute: Rebuild part-a ('override-build' property changed)
unknown argument ignored: lazytime
+ mkdir -p /home/lengau/Work/Code/craft-parts/parts/part-a/install/my-dir/subdir-a
Error: Failed to organize part 'part-a': trying to organize '*' to '(overlay)/', but '(overlay)/my-dir' already exists.
```

lengau:
This handles cases where organizing to the overlay could result in errors because you had previously organized to the overlay.

Fixes: #1488
CRAFT-5043

---


tests/unit/features/partitions/executor/test_organize.py
```diff
@@ -226,15 +325,25 @@
...
-        {
-            "build_files": ["my-dir/subdir/foo", "my-dir/bar"],
-            "organize_map": {"(build)/*": "(mypart)/"},
-            "expected": [
-                (["bar", "subdir"], "../partitions/mypart/parts/part-name/install"),
-                (["foo"], "../partitions/mypart/parts/part-name/install/subdir"),
-            ],
-            "check_copy": True,
-        },
+        pytest.param(
```
lengau: @cmatsuoka this is a change in the organize-from-build behaviour, but I believe it matches our intended behaviour.

tests/unit/features/partitions/executor/test_organize.py
```diff
@@ -35,6 +35,105 @@
             },
             "expected": [(["bar1", "baz1", "foo1", "qux1"], "")],
         },
+        # Actual behaviour in snapcraft 8.14. The next few tests ensure that our current
```
lengau, line 43: These added actual-behaviour-in-snapcraft-8.14 tests also succeed on the current main branch and on commit cb310175c34aa66413301a074b7e4ba560b865a5

craft_parts/utils/file_utils.py
```diff
@@ -344,3 +348,126 @@ def _file_reader_iter(
...
+    Checks that the two directories provided can be merged without overwriting files.
+
+    :param strict: if True, errors if overwriting a file, even if it's identical.
+    """
+    conflicts: dict[pathlib.Path, list[str]] = {}
+    for source_path in (*src_root.rglob("*"), src_root):
+        relative_path = source_path.relative_to(src_root)
+        dest_path = dst_root / relative_path
+        if not dest_path.exists() or dest_path.samefile(source_path):
+            continue
```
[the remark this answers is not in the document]
lengau: Yes the purpose here includes saying it's not a conflict if the destination is a broken link.

craft_parts/executor/organize.py
```diff
@@ -95,13 +99,63 @@ def organize_files(  # noqa: PLR0912
...
+            if dst_partition_pair.partition == OVERLAY_PARTITION:  # noqa: SIM102
+                # Organizing a symlink to its destination simply deletes the link.
+                if src_path.is_symlink():
+                    relative_src_target = src_path.readlink().relative_to(src_root)
+                    dst_target = (
+                        install_dir_map[OVERLAY_PARTITION] / relative_src_target
+                    )
+                    if dst_target.exists():
+                        src_path.unlink()
+                        continue
```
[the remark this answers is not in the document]
lengau: @copilot please write a test case demonstrating this.

craft_parts/executor/organize.py
```diff
@@ -95,13 +99,63 @@ def organize_files(  # noqa: PLR0912
...
+                    or dst_partition_pair.partition == OVERLAY_PARTITION
+                ):
+                    if dst_path in install_dir_map.values():
+                        real_dst_path = dst_path / src_path.name
+                    else:
+                        real_dst_path = dst_path
+                    if not overwrite:
+                        conflicts = file_utils.find_merge_conflicts(
+                            src_path,
+                            dst_path,
```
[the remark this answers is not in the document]
lengau: @copilot write a test case to demonstrate this.

craft_parts/executor/organize.py
```diff
@@ -120,6 +174,21 @@ def organize_files(  # noqa: PLR0912
                             "supposed to be a directory, end it with a slash."
                         ),
                     )
+                elif dst_partition_pair.partition == OVERLAY_PARTITION:
+                    equivalent, msg = file_utils.are_paths_equivalent(
+                        src_path, dst_path
+                    )
+                    if equivalent:
+                        src_path.unlink()
```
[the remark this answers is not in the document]
lengau: @copilot write a test case to demonstrate this.

craft_parts/executor/organize.py
```diff
@@ -144,6 +213,26 @@ def organize_files(  # noqa: PLR0912
...
                     rel_dst_string = os.path.join(dst_string, os.path.basename(src))  # noqa: PTH118, PTH119
+                    if dst_partition_pair.partition == OVERLAY_PARTITION:
+                        src_path = pathlib.Path(src)
+                        real_dst_path = pathlib.Path(real_dst)
+                        equivalent, msg = file_utils.are_paths_equivalent(
+                            src_path, real_dst_path
+                        )
+                        if equivalent:
+                            if src_path.is_dir():
+                                file_utils.move(src, real_dst)
```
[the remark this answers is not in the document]
lengau: @copilot write a test case to demonstrate this.

craft_parts/utils/file_utils.py
```diff
@@ -344,3 +348,126 @@ def _file_reader_iter(
...
+        file_type = "fifo"
+    elif stat.S_ISBLK(result.st_mode):
+        file_type = "blk"
+    elif stat.S_ISCHR(result.st_mode):
+        file_type = "chr"
+    elif stat.S_ISSOCK(result.st_mode):
+        file_type = "socket"
+    elif stat.S_ISREG(result.st_mode):
+        file_type = "file"
+    elif stat.S_ISWHT(result.st_mode):
```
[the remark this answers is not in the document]
lengau: @copilot write a test case to demonstrate this.

craft_parts/utils/file_utils.py
```diff
@@ -344,3 +348,126 @@ def _file_reader_iter(
...
+
+    if a_stat.st_gid != b_stat.st_gid:
+        differences.append(f"different gids ({a_stat.st_gid}, {b_stat.st_gid})")
+
+    if differences:
+        return False, differences
+
+    if a_mode != b_mode:
+        differences.append(f"different modes ({a_mode:o}, {b_mode:o})")
+
```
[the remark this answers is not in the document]
lengau: @copilot write a test case to demonstrate this.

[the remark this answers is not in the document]
lengau: @copilot write a test case that demonstrates this issue.

craft_parts/executor/organize.py
```diff
@@ -95,13 +107,85 @@ def organize_files(  # noqa: PLR0912
...
+                # Organizing a symlink to its destination simply deletes the link.
+                if src_path.is_symlink():
+                    relative_src_target = src_path.resolve().relative_to(src_root)
+                    dst_target = (
+                        install_dir_map[OVERLAY_PARTITION] / relative_src_target
+                    )
+                    if dst_path in install_dir_map.values():
+                        real_dst_path = dst_path / src_path.name
+                    else:
+                        real_dst_path = dst_path
```
bepri, line 163: This pattern could be made more readable if moved into another function

tests/unit/features/partitions/executor/test_organize.py
cmatsuoka: This is consistent with organize-to-partition behavior, but it seems to be different from organize within the partition. If you add the following test to `tests/unit/executor/test_organize.py` it should fail:
```
        {
            "setup_files": ["my-dir/foo.txt", "bar.txt"],
            "setup_dirs": ["my-dir", "your-dir"],
            "organize_map": {
                "my-dir": "your-dir/",
            },
            "expected": [
                ([], ""),
                (
                    ["bar.txt", "my-dir"],
                    "../partitions/mypart/parts/part-name/install/your-dir",
                ),
                (
                    ["foo.txt"],
                    "../partitions/mypart/parts/part-name/install/your-dir/my-dir",
                ),
            ],
        },
```
If organizing to partition and within partition are really inconsistent, we must decide on which side organizing from `(build)` should be -- organizing to partition seems to be the most intuitive, but the fact that the behaviors are different looks very wrong.

Also, what happens when organizing from/to `(default)`?

tests/unit/features/partitions/executor/test_organize.py
```diff
@@ -35,6 +40,105 @@
...
+                "setup_files": ["my-dir/foo.txt", "bar.txt"],
+                "organize_map": {
+                    "bar.txt": "other-dir",
+                    "my-dir": "other-dir",
+                },
+                "expected": [
+                    (["other-dir"], ""),
+                    (["bar.txt", "foo.txt"], "other-dir"),
+                ],
+            },
```
cmatsuoka, line 58: This is the case that seems to be inconsistent with `build-wildcard-to-partition`.

tests/unit/features/partitions/executor/test_organize.py
lengau: I feel like this is still important despite organize-from-build being reverted, but I can't place my finger on why.

I've put this edge case in the spec we're working on, but I'd like to cover it better here.

lengau:
@copilot resolve the merge conflicts in this pull request

craft_parts/executor/organize.py
```diff
@@ -67,9 +71,25 @@ def organize_files(  # noqa: PLR0912
     """
     for key in sorted(file_map, key=lambda x: ["*" in x, x]):
         src = get_src_path(key, part_name, install_dir_map, default_partition)
+        src_root = install_dir_map.get(None) or install_dir_map[default_partition]
+
+        # Remove the leading slash so the path actually joins
+        # Also trailing slash is significant, be careful if using pathlib!
```
cmatsuoka, line 117: We discussed leading slashes in the organize origin side and the conclusion was that we should error out when organizing an absolute path, should we do the same for the destination side? What was the previous behavior if we organize to absolute paths?

craft_parts/executor/organize.py
```diff
@@ -78,12 +98,83 @@ def organize_files(  # noqa: PLR0912
         src_count = 0
         for src in sources:
             src_count += 1
+            src_path = pathlib.Path(src)
+
+            # Special behaviour for the overlay partition to allow idempotence without
+            # needing overwrite to be True.
```
cmatsuoka: To be really idempotent I think directories must always organize to directories like in the non-partition case, otherwise  in the first run it will create the directory and in the second run it will place the origin inside the directory that was created.

craft_parts/executor/organize.py
```diff
@@ -126,6 +232,26 @@ def organize_files(  # noqa: PLR0912
...
                     rel_dst_string = os.path.join(dst_string, os.path.basename(src))  # noqa: PTH118, PTH119
+                    if dst_partition_pair.partition == OVERLAY_PARTITION:
+                        src_path = pathlib.Path(src)
+                        real_dst_path = pathlib.Path(real_dst)
+                        equivalent, msg = file_utils.are_paths_equivalent(
+                            src_path, real_dst_path
+                        )
+                        if equivalent:
+                            if src_path.is_dir():
+                                file_utils.move(src, real_dst)
```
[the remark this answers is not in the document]
lengau: @copilot write a test case that demonstrates this issue.

craft_parts/executor/organize.py
lengau: The leading slash gets removed and it becomes relative to the destination path there. See: https://github.com/canonical/craft-parts/blob/3495de6260763153625e923923ed7e88bc594e23/craft_parts/executor/organize.py#L182

What we don't have protection against is someone organizing to a relative path outside of the destination directory.

craft_parts/executor/organize.py
lengau: Isn't that what the test with id `merge-dirs-in-overlay-success` checks we aren't doing?

craft_parts/utils/file_utils.py
```diff
@@ -344,3 +349,133 @@ def _file_reader_iter(
...
+        differences.append(f"different types ({a_type}, {b_type})")
+
+    if a_stat.st_uid != b_stat.st_uid:
+        differences.append(f"different uids ({a_stat.st_uid}, {b_stat.st_uid})")
+
+    if a_stat.st_gid != b_stat.st_gid:
+        differences.append(f"different gids ({a_stat.st_gid}, {b_stat.st_gid})")
+
+    if differences:
+        return False, differences
```
bepri: Is there a good reason to do this early return of "simple" differences? I think as a user I'd like a comprehensive list of problems rather than solving issues just to be told there's even more it hadn't checked yet.

craft_parts/executor/organize.py
```diff
@@ -78,12 +98,91 @@ def organize_files(  # noqa: PLR0912
         src_count = 0
         for src in sources:
             src_count += 1
+            src_path = pathlib.Path(src)
+
+            # Special behaviour for the overlay partition to allow idempotence without
+            # needing overwrite to be True.
+            if dst_partition_pair.partition == OVERLAY_PARTITION:  # noqa: SIM102
+                # Organizing a symlink to its destination simply deletes the link.
+                if src_path.is_symlink():
```
bepri: Why did this noqa get added? The fix seems reasonable and would reduce rightward drift. If you want to keep the comment, do something like:

```py
if (
  x
  # comment
  and y
):
```

craft_parts/executor/organize.py
```diff
@@ -78,12 +98,91 @@ def organize_files(  # noqa: PLR0912
...
+                    try:
+                        # resolve() follows the symlink; relative_to() raises ValueError
+                        # if the target lies outside src_root (e.g. absolute symlinks
+                        # such as /usr/bin/true). Fall through to normal organise in
+                        # that case.
+                        relative_src_target = src_path.resolve().relative_to(src_root)
+                    except ValueError:
+                        pass
```
bepri:
```suggestion
                    except ValueError:
                        continue
```
This can allow removing the `else:` clause, reducing rightward drift again

craft_parts/executor/organize.py
```diff
@@ -78,12 +98,91 @@ def organize_files(  # noqa: PLR0912
...
+                    else:
+                        dst_target = (
+                            install_dir_map[OVERLAY_PARTITION] / relative_src_target
+                        )
+                        if dst_path in install_dir_map.values():
+                            real_dst_path = dst_path / src_path.name
+                        else:
+                            real_dst_path = dst_path
```
bepri: This section of code seems logically separate from what follows, consisting of variable declarations instead of calculations. Since so many `if` statements follow, it could be helpful to separate them out with some whitespace for readability.

craft_parts/executor/organize.py
```diff
@@ -78,12 +98,91 @@ def organize_files(  # noqa: PLR0912
...
+                            )
+                            raise errors.FileOrganizeError(
+                                part_name=part_name,
+                                message=(
+                                    f"trying to organize directory {key!r} to "
+                                    f"{file_map[key]!r} but conflicts exist while "
+                                    f"merging the directories:\n"
+                                    f"{conflicts_list}"
+                                ),
+                            )
```
bepri, line 215: nit: I think it would help to refactor this error-building code into its own error class, to help keep this method focused just on organize logic.

craft_parts/utils/file_utils.py
```diff
@@ -344,3 +349,143 @@ def _file_reader_iter(
...
+    elif stat.S_ISWHT(result.st_mode):
+        file_type = "whiteout"
+    else:
+        file_type = "unknown"
+    return file_type
+
+
+def are_paths_equivalent(  # noqa: PLR0912
+    a: Path, b: Path
+) -> tuple[Literal[True], None] | tuple[Literal[False], list[str]]:
```
bepri: `Literal[True], None` and `Literal[False], something` could just be `something` and then checks against the return can do `something is not None` instead of inspecting the first value of the tuple. This return should be simplified

bepri (commented):
I have some comments largely focusing on improving the readability of this code. I don't have any problems with the code's functionality.

This review focuses on the code changes for now, not the tests

craft_parts/executor/organize.py
lengau: Optimism (pessimism?) that I'd need more cases under the overlay partition later tbh

craft_parts/utils/file_utils.py
lengau: Yeah I agree — this was originally just returning a bool and then it kinda got away from me.

craft_parts/utils/file_utils.py
lengau: Yes — if any of these are hits, the further differences will be redundant. I especially don't want to do expensive checks like checksumming the files when these cheap checks pop up.

Would it be better better if I moved this to line 437 (so we might have redundant info but we won't have expensive hashes)?

craft_parts/executor/organize.py
lengau: No, that changes the logic :-/

craft_parts/executor/organize.py
cmatsuoka: I'm inclined to think that the mv-like behavior when using partitions is a bug, maybe introduced during a refactoring run -- maybe @upils knows more about this, but having different behaviors depending on cross-partition organization seems wrong.

craft_parts/utils/file_utils.py
bepri: I don't see where the redundant info would come from, unless you just mean they're already clearly different files so of course the other info would be different too?

I think your proposed move to just before the content checks makes sense.

craft_parts/executor/organize.py
bepri: I think it wouldn't be bad to add that indentation later, if needed. I'd prefer that over this potentially _not_ needing to change and just having an inexplicable noqa

craft_parts/executor/organize.py
```diff
@@ -26,20 +26,64 @@
...
+    key: str,
+    file_map: dict[str, str],
+    src_path: pathlib.Path,
+    dst_path: pathlib.Path,
+    dst_string: str,
+) -> bool:
+    """Check if source and destination are equivalent on the overlay partition.
+
+    If equivalent, removes the source (or merges if it's a directory) and returns True.
+    If not equivalent, raises FileOrganizeError.
```
bepri: This return value can be removed I believe, since it's always `True` or raises.

tests/unit/utils/test_file_utils.py
```diff
@@ -273,3 +277,433 @@ def test_create_similar_directory_permissions(tmp_path, mock_chown):
...
+    (tmp_path / "other_dir").mkdir(mode=0o700)
+    # This is the most permissive we can make it in GH runners
+    (tmp_path / "permissive_dir").mkdir(mode=0o750)
+    (tmp_path / "broken_link").symlink_to("this_does_not_exist")
+    (tmp_path / "another_broken_link").symlink_to("this_does_not_exist")
+    (tmp_path / "different_broken_link").symlink_to("that_does_not_exist")
+    (tmp_path / "symlink_to_file").symlink_to("empty_file")
+    (tmp_path / "different_symlink_to_file").symlink_to("another_empty_file")
+    (tmp_path / "alt_symlink_to_file").symlink_to("hardlink")
+    (tmp_path / "symlink_to_dir").symlink_to("empty_dir")
```
bepri, line 447: nit: I know only this test would use it, but these could be extracted into a fixture to avoid redoing this init for each parametrization

tests/unit/utils/test_file_utils.py
```diff
@@ -273,3 +277,433 @@ def test_create_similar_directory_permissions(tmp_path, mock_chown):
...
+        pytest.param(
+            {"my-blk": {"type": "blk", "mode": 0o600}},
+            {"my-blk": {"type": "chr", "mode": 0o600}},
+            {"my-blk": ["different types (blk, chr)"]},
+            id="block-device",
+        ),
+    ],
+)
+def test_find_merge_conflicts(
+    fs: FakeFilesystem,
```
bepri: Move to `pytest.mark.usefixtures()`?

bepri (approved):
I combed through the tests and I've only got nits left. Take 'em or leave 'em :)

For my remaining comments from the last review, I trust your judgment

cmatsuoka (approved):
I think we can land this but we still need to discuss the cross-partition organize inconsistency with @upils in Madrid. I suspect this is a bug introduced in a refactoring, and the correct behavior in both cases should be directory merging (so that `bin: (partition)/usr/bin` always works).

tests/unit/utils/test_file_utils.py
lengau: Actually each parametrized version would still get a fresh copy of the fixture (it relies on `tmp_path` which definitely needs to be fresh for each parametrization).

PR #1542, opened 2026-04-10
docs: add -i option example to both python plugin versions

asanvaq:
Recent UR feedback requested documentation updates  to provide a clear explanation of the -i option for the Python plugin, including its purpose, usage, and an example. 

This information has been added in the form of an example in both python plugin versions.

---


docs/common/craft-parts/reference/plugins/python_plugin.rst
```diff
@@ -101,3 +101,23 @@ During the build step, the plugin performs the following actions:
    ``python-packages`` keys.
 #. If the source contains a ``setup.py`` or ``pyproject.toml`` file, those
    files are used to install the dependencies specified by the package itself.
+
+Example
+-------
+
+The following snippet declares a part using the python plugin to define a custom
+mirror index in the ``python-requirements`` key:
```
medubelko:
```suggestion
The following snippet declares a part with the Python plugin, and selects the 
project's ``requirements.txt`` file with the ``python-requirements`` key:
```

docs/common/craft-parts/reference/plugins/python_plugin.rst
```diff
@@ -101,3 +101,23 @@ During the build step, the plugin performs the following actions:
...
+
+.. code-block:: yaml
+
+   parts:
+     my-part:
+       plugin: python
+       python-requirements:
+         - requirements.txt
+
+where ``requirements.txt`` contains:
```
medubelko:
```suggestion
The ``requirements.txt`` declares a custom mirror for its package index:
```

docs/common/craft-parts/reference/plugins/python_plugin.rst
```diff
@@ -101,3 +101,23 @@ During the build step, the plugin performs the following actions:
...
+The following snippet declares a part using the python plugin to define a custom
+mirror index in the ``python-requirements`` key:
+
+.. code-block:: yaml
+
+   parts:
+     my-part:
+       plugin: python
+       python-requirements:
+         - requirements.txt
```
medubelko, line 132: @tigarmo I thought it wasn't necessary to specify `python-requirements` unless there was no package manifest at the project root. Maybe we need to revise the _How it works_ section?

docs/common/craft-parts/reference/plugins/python_plugin.rst
medubelko: The point we're making here is orthogonal to standard use of the Python plugin. I think we should make it clear that this is a special situation. It might even work best as a second example.

asanvaq: Makes sense. There is already an example in the _Dependencies_ sections of both. Do you think it is still necessary to include another one?

medubelko: Yes. :)

docs/common/craft-parts/reference/plugins/python_plugin.rst
tigarmo: It depends on the project. Some Python projects (older ones) have a setup.py/pyproject.toml with pinned (or not pinned) dependencies, and some projects have just the dependency names on the "package manifest" (assuming that means setup.py/pyproject.toml) *plus* requirements files with pinned versions for direct and transitive dependencies.

In the later case one would indeed need to declare `python-requirements`. This is how our own apps did this before we moved to uv ([example](https://github.com/canonical/rockcraft/blob/1.2.0/snap/snapcraft.yaml#L60))

docs/common/craft-parts/reference/plugins/python_plugin.rst
asanvaq: Done!

docs/common/craft-parts/reference/plugins/python_plugin.rst
```diff
@@ -101,3 +101,40 @@ During the build step, the plugin performs the following actions:
...
+stage packages:
+
+.. code-block:: yaml
+
+  parts:
+    pyfiglet:
+      plugin: python
+      source: .
+      python-packages:
+        - pyfiglet==0.7.6 # Note: should match `version` above
```
bepri: It seems like this block came out of a real-world file where this comment made sense, but there's no longer a `version:` key for it to be referencing. Could we remove it in this example?

asanvaq: Yes, thanks for spotting that!

bepri (approved):

docs/common/craft-parts/reference/plugins/python_plugin.rst
medubelko: Are there any cases where I could have `requirements.txt` and I wouldn't need to map it with `python-requirements`?

asanvaq: Is this discussion blocking or can this be approved and merged as it is now? If there are any non blocking changes required, I can create an issue to address them later.

medubelko: I think this should be addressed later; it seems to me that the description of the `python-requirements` key isn't complete.

medubelko: Siphoned into a different issue: https://github.com/canonical/craft-parts/issues/1551

medubelko (approved):

docs/common/craft-parts/reference/plugins/python_plugin.rst
asanvaq: Thanks!

PR #1544, opened 2026-04-14
fix: prevent organizing a directory into itself

cmatsuoka:
Only move a directory after checking if it's not being organized into
itself (or a subpath of itself).

---

### DIFF
--- a/craft_parts/executor/organize.py
+++ b/craft_parts/executor/organize.py
@@ -28,6 +28,7 @@ import contextlib
 import os
 import shutil
 from glob import iglob
+from pathlib import Path
 from typing import TYPE_CHECKING
 
 from craft_parts import errors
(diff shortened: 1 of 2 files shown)


tests/unit/executor/test_organize.py
```diff
@@ -163,6 +163,12 @@
             "organize_map": {"**/bardir/*": "bardir/"},
             "expected": [(["bardir", "foodir"], ""), (["foo"], "bardir")],
         },
+        # Organize directory into itself
```
tigarmo, line 166: The change itself seems fine with me, but why is a directory being organized "into itself"? The install dir has only two files (no directories)

craft_parts/executor/organize.py
```diff
@@ -160,7 +160,9 @@ def organize_files(  # noqa: PLR0912
                 else:
                     file_utils.link_or_copy(src, dst)
             else:
-                file_utils.move(src, dst)
+                # Prevent moving a directory into itself or its own subtree
+                if not Path(src).is_dir() or not Path(dst).is_relative_to(src):
```
tigarmo, line 164: can you add a changelog entry to expedite the fix release?

tigarmo (approved):

cmatsuoka:
There are cases not covered by this solution, this needs to be redone.

tests/unit/executor/test_organize.py
lengau: The first time, yes. The second time you run `organize` it has the results of the first run in it as well.

cmatsuoka:
To be reworked with a new implementation of organize-from-build.

PR #1553, opened 2026-04-21
fix(deb): avoid re-downloading cached stage-packages

gcomneno:
## Summary

Mitigate repeated `.deb` downloads from a warm cache in the `stage-packages` / apt-cache path.

This wraps `package.candidate.fetch_binary(...)` with a narrow local patch for
`apt.package._file_is_same`, using:

- binary file mode
- order-independent hash comparison

so cached `.deb` files that already match the expected metadata are treated as cache hits.

## Context

Refs #1495.

This is intentionally scoped to the `stage-packages` / apt-cache download path.

It is related to, but separate from, #1485, which reduces some `python-apt`
usage in the installed-package check path.

## Testing

- `ruff check craft_parts/packages/apt_cache.py tests/unit/packages/test_apt_cache.py`
- `pytest -q tests/unit/packages/test_apt_cache.py -rs`
- `pytest -q tests/unit/packages/test_deb.py -rs`
- `pytest -q tests/unit/packages -rs`


gcomneno:
CI failed in an unrelated apt mirror/cache fetch while running the Java plugin tests on arm64: libgnutls30t64 returned 404 from the mirror. I don't have permissions to rerun failed jobs, but the branch is rebased and the relevant diff is limited to the apt cache change.

gcomneno:
Updated this branch and fixed the formatting issue reported by lint.

Local validation:
- `uv run ruff check craft_parts/packages/apt_cache.py tests/unit/packages/test_apt_cache.py`
- `uv run ruff format --check craft_parts/packages/apt_cache.py tests/unit/packages/test_apt_cache.py`
- `git diff --check`

I could not run `tests/unit/packages/test_apt_cache.py` locally in the uv environment because `python-apt` is not available there (`ModuleNotFoundError: No module named 'apt'`). The failure happens during test collection before reaching the PR-specific assertions.

gcomneno:
Following up after rechecking this PR against the current `main`.

The underlying `AptCache.fetch_archives()` path still calls `package.candidate.fetch_binary(...)` directly, so the cache-hit workaround remains applicable.

I also reapplied the PR diff to current `main` in a clean temporary checkout. The patch still applies cleanly, `git diff --check` and Ruff pass, and the two PR-specific regression tests pass with the system `python-apt` available:

`2 passed, 15 deselected`

I do not see any further code changes needed from my side, but I am happy to adjust the approach if maintainers prefer a different direction.

PR #1557, opened 2026-04-22
fix(deb): preserve architecture in primed stage packages

gcomneno:
## Summary
- preserve foreign-architecture package names when extracting deb package name/version metadata
- use `` instead of `` when reading deb metadata
- add a unit test covering architecture-qualified package names

## Related
- snapcraft#5998


gcomneno:
Closing this in favor of #1558, which supersedes this version of the same fix.

PR #1558, opened 2026-04-22
fix(packages/deb): preserve architecture in primed stage packages

gcomneno:
## Summary
- preserve foreign-architecture package names when extracting deb package name/version metadata
- use `` instead of `` when reading deb metadata
- add a unit test covering architecture-qualified package names

## Related
- snapcraft#5998


gcomneno:
I updated the PR title to `fix(deb): preserve architecture in primed stage packages`, but I cannot rerun the stale title check from my side because rerunning failed workflows requires higher repository permissions.

From the current failures:
- the OSV-scanner failure appears to come from a pre-existing `lxml` vulnerability in the generated requirements set, not from this deb metadata change
- the `lint / files` and `test-other-plugins / Slow tests (noble, s390x, edge, 3.12)` failures look unrelated to this patch and seem to come from workflow/matrix issues rather than from the code changes here

Happy to adjust anything in the patch itself if needed.

lengau (approved):
Thanks! This looks good to me, but I'm requesting @cmatsuoka who has a bit more knowledge of what this could potentially break.

cmatsuoka (approved):
Looks nice, thanks!

PR #1562, opened 2026-04-25
fix(organize): reject sources outside install dir

gcomneno:
## Summary

Reject `organize` source entries that resolve outside the part install directory.

This fixes cases where `organize` could read from absolute paths or use directory traversal to escape the install directory.

## Testing

- `ruff check craft_parts/executor/organize.py tests/unit/executor/test_organize.py tests/unit/features/partitions/executor/test_organize.py`
- `pytest -q tests/unit/executor/test_organize.py -rs`
- `pytest -q tests/unit/features/partitions/executor/test_organize.py -rs`

## Notes

This keeps normalized in-tree paths working, such as `dir/../foo`, while rejecting absolute and escaping source paths.


gcomneno:
The failing OSV-scanner check appears unrelated to this change.

It is reporting GHSA-vfmq-68hx-4jfw for `lxml 6.0.2` from `requirements/uv-requirements.--all-extras.txt`. The PR itself only updates `craft_parts/executor/organize.py` and its unit coverage.

craft_parts/executor/organize.py
```diff
@@ -166,7 +166,37 @@ def get_src_path(
...
+                    message=(
+                        f"trying to organize from {key!r}, but source paths must stay within "
+                        f"the install directory"
+                    ),
+                )
+            normalized_parts.pop()
+            continue
+        normalized_parts.append(part)
+
+    src_path = base_dir.joinpath(*normalized_parts)
```
steinbro: This feels more granular than necessary. `Path.resolve()` should remove any `.` and `..` references, so why not just concatenate `base_dir` with `src_inner_path`, `resolve()` the result, then check that it still begins with `base_dir`?

gcomneno:
Ops.. my own fault! I simplified the source path validation to use resolve() for the containment check, while keeping the original non-resolved source path for the actual organize operation so existing symlink behavior is preserved.

gcomneno:
The remaining QA failure appears unrelated to this change. The failing job is test-java-plugins / Slow tests (noble, s390x, edge, 3.12), specifically test_jlink_plugin_with_jar[partitions], where OpenJDK 21 jlink aborts with SIGSEGV on linux-s390x. The same job also failed in the previous QA run before the latest update.

steinbro (approved):
That looks better, thanks!

cmatsuoka (approved):
Looks good!

steinbro:
@lengau to review and check what's going on with  the test failures.

craft_parts/executor/organize.py
```diff
@@ -320,7 +320,19 @@ def get_src_path(
     if src_partition == DEFAULT_PARTITION:
         src_partition = default_partition
 
-    return os.path.join(install_dir_map[src_partition], src_inner_path)  # noqa: PTH118
+    base_dir = Path(install_dir_map[src_partition]).resolve()
+    src_path = base_dir / src_inner_path
+
+    if not src_path.resolve().is_relative_to(base_dir):
```
lengau: Thanks for the PR! It mostly looks good, but right here we need to not resolve symlinks. If the file being organized is a symlink to `/usr/bin/true` (as in one of the failing test examples), the destination file should be a symlink to `/usr/bin/true`.

cmatsuoka:
It seems that tests are failing because `resolve()` is following the symlink so the link gets resolved to a file outside the install dir. We should prevent files from outside to be organized into the package, but symlinks pointing to files outside should be allowed.

cmatsuoka (changes requested):
Symlinks to a file outside the install dir should be allowed.

cmatsuoka (approved):
Thanks. This is a simple and nice fix that solves the symlink issue.

PR #1567, opened 2026-05-06
feat(overlay): plugins can define overlay commands

lengau:
Allow plugins to define overlay commands that run inside of the overlay
chroot. If the user adds an `override-overlay`, they can replace those
commands, or they can wrap them and let `craftctl default` signal where
the plugin's original commands would run.

Unlike other steps, only `craftctl default` is accepted in `override-overlay`.
This is not a regression because previously the `craftctl` command couldn't
be used in an override-overlay scriptlet at all.

In addition to `overlay-script` being mutually exclusive with `override-overlay`,
`overlay-script` is now also mutually exclusive with the plugin offering overlay
commands.

---
CRAFT-5027

### DIFF
--- a/craft_parts/executor/step_handler.py
+++ b/craft_parts/executor/step_handler.py
@@ -451,10 +451,10 @@ class StepHandler:
             "export -f __craftctl_default\n"
             "\n"
             "craftctl() {\n"
-            '    if [ "${1-}" = "default" ]; then\n'
+            '    if [ "$#" -eq 1 ] && [ "$1" = "default" ]; then\n'
             "        __craftctl_default\n"
             "    else\n"
-            '        echo "Error: craftctl ${1-} cannot be used in override-overlay" >&2\n'
+            f'        echo "Error: craftctl ${{1-}} cannot be used in {scriptlet_name}" >&2\n'
             "        return 1\n"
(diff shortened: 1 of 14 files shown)


smethnani (approved):
Thanks!

craft_parts/executor/step_handler.py
```diff
@@ -411,6 +422,64 @@ def run_scriptlet(
...
+        scriptlet: str,
+        *,
+        work_dir: Path,
+        scriptlet_name: str = "override-overlay",
+    ) -> None:
+        """Execute an override-overlay scriptlet inside a chroot.
+
+        Since the chroot doesn't have access to the craftctl binary, this
+        injects a bash function that implements ``craftctl default`` by running
+        the plugin's chroot commands inline.
```
cmatsuoka, line 436: This is an interesting trick.

cmatsuoka (approved):
Looks good. I have mixed feelings about the craftctl default trick but I don't see an obvious alternative to that.

craft_parts/executor/step_handler.py
lengau: Yeah I'm not sure it's ideal but I don't think we have any other way to do it.

PR #1569, opened 2026-05-06
feat: deprecate OsRelease in favor of distro

lengau:
We've been using `distro` in several places, we might as well use it in craft-parts too.

This doesn't remove it yet though since that would require a major release.

### DIFF
--- a/craft_parts/overlays/errors.py
+++ b/craft_parts/overlays/errors.py
@@ -66,13 +66,13 @@ class OverlayChrootExecutionError(OverlayError):
 class IncompatibleChrootError(OverlayError):
     """Failed to use host package sources because chroot is incompatible.
 
-    :param key: os-release key which was tested.
-    :param host: value from host os-release which was tested.
-    :param chroot: value from chroot os-release which was tested.
+    :param key: Distribution attribute which was tested.
+    :param host: Value from the host distribution which was tested.
+    :param chroot: Value from the chroot distribution which was tested.
     """
 
     def __init__(self, key: str, host: str, chroot: str) -> None:
-        self.message = f"key {key} in os-release expected to be {host} found {chroot}"
+        self.message = f"attribute {key} expected to be {host} found {chroot}"
         brief = f"Unable to use host sources: {self.message}"
 
         super().__init__(brief=brief)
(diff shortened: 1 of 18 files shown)


craft_parts/overlays/errors.py
```diff
@@ -66,13 +66,13 @@ def __init__(self, message: str) -> None:
...
-    :param host: value from host os-release which was tested.
-    :param chroot: value from chroot os-release which was tested.
+    :param key: Distribution attribute which was tested.
+    :param host: Value from the host distribution which was tested.
+    :param chroot: Value from the chroot distribution which was tested.
     """
 
     def __init__(self, key: str, host: str, chroot: str) -> None:
-        self.message = f"key {key} in os-release expected to be {host} found {chroot}"
+        self.message = f"attribute {key} expected to be {host} found {chroot}"
```
bepri, line 75:
```suggestion
        self.message = f"OS attribute {key!r} expected to be {host!r}, found {chroot!r}"
```
This lost some information, I think this is enough to fix it back up though (plus some nits because I'm a complainer)

tests/unit/utils/test_os_utils.py
```diff
@@ -245,21 +245,25 @@ def _write_os_release(self, contents) -> str:
             f.write(contents)
         return path
 
+    def _new_release(self, contents: str) -> os_utils.OsRelease:
+        with pytest.deprecated_call(
+            match=r"'OsRelease' is deprecated\. Use the 'distro' module instead\."
+        ):
+            return os_utils.OsRelease(os_release_file=self._write_os_release(contents))
+
     def test_blank_lines(self):
```
bepri, line 254: Several of these tests strike me as something that can be deleted now. We don't need to test that `distro` works as expected, right? Just that our thin wrapper around it returns what we'd expect?

bepri (approved):

tests/unit/utils/test_os_utils.py
lengau: OSRelease isn't a thin wrapper around distro. It has separate logic.

bepri: Ah, I lost track of where I was in the file and thought this was a test for the general utils. Thanks

mr-cal (approved):

lengau:
superseded by https://github.com/canonical/craft-parts/pull/1581

PR #1583, opened 2026-05-12
ci: separate flaky tests into their own job

lengau:
Move tests marked with `@pytest.mark.flaky` into a dedicated `test-flaky` CI job so they can be rerun independently and quickly when they fail due to transient issues (e.g. Ubuntu archive syncs).

This avoids needing to re-run the full suite of slow tests if flaky tests fail.

Artifact upload failure is fixed in https://github.com/canonical/starflow/pull/139


mr-cal (approved):
LGTM, although there are flaky tests not marked as flaky, like the one currently failing in tests/integration/packages/test_chisel.py.

bepri (approved):

PR #1586, opened 2026-05-18
fix(organize): allow merging directories in overlay with metadata differences

### LINKED ISSUE #1487: Multiple parts cannot organize to the same directory in the overlay
### Bug Description

When pre-filling the overlay directory, the full directory trees from multiple parts that add to the overlay directory must be disjoint. If you try to put files into the same overlay directory from separate parts, craft-parts complains that the parent directory already exists.

### To Reproduce

1. Copy the part YAML below into `parts.yaml`
2. `mkdir /tmp/empty`
3. `sudo $(which python) -m craft_parts --verbose --overlay-base=/tmp/empty --partitions=default,other overlay`

### part yaml

```shell
parts:
  part-a:
    plugin: nil
    override-build: |
      mkdir $CRAFT_PART_INSTALL/my-dir
    organize:
      '*': (overlay)/

  part-b:
    plugin: nil
    override-build: |
      # This works
      mkdir -p $CRAFT_PART_INSTALL/my-other-dir/subdir

      # This fails.
      mkdir -p $CRAFT_PART_INSTALL/my-dir/subdir
    organize:
      '*': (overlay)/
```

### Relevant log output

```shell
Execute: Pull part-a
Execute: Pull part-b
Execute: Overlay part-a
Execute: Build part-a (organize contents to overlay)
unknown argument ignored: lazytime
+ mkdir -p /home/lengau/Work/Code/craft-parts/parts/part-a/install/my-dir/subdir-a
Execute: Overlay part-b
Execute: Build part-b (organize contents to overlay)
unknown argument ignored: lazytime
+ mkdir -p /home/lengau/Work/Code/craft-parts/parts/part-b/install/my-other-dir/subdir-b
+ mkdir -p /home/lengau/Work/Code/craft-parts/parts/part-b/install/my-dir/subdir-b
Error: Failed to organize part 'part-b': trying to organize '*' to '(overlay)/', but '(overlay)/my-dir' already exists.
```

lengau:
Multiple parts should be able to contribute to the same directory in the overlay partition, even if they have different permissions or ownership. Previously, such differences would trigger a FileOrganizeError during the organization phase.

Fixes: #1487

### DIFF
--- a/craft_parts/executor/organize.py
+++ b/craft_parts/executor/organize.py
@@ -232,8 +232,9 @@ def organize_files(  # noqa: PLR0912, PLR0915
                     file_utils.link_or_copy_tree(
                         src_path,
                         real_dst_path,
-                        overwrite_metadata=dst_partition_pair.partition
-                        != OVERLAY_PARTITION,
+                        overwrite_metadata=(
+                            dst_partition_pair.partition != OVERLAY_PARTITION
+                        ),
                     )
                     shutil.rmtree(src)
                     continue
--- a/craft_parts/executor/organize.py
+++ b/craft_parts/executor/organize.py
@@ -66,9 +66,11 @@ def _check_overlay_equivalence(
     :raises FileOrganizeError: If the paths are not equivalent.
     """
     msg = file_utils.get_path_differences(src_path, dst_path)
-    if not msg:
+    if not msg or (
+        not src_path.is_symlink() and src_path.is_dir() and dst_path.is_dir()
+    ):
         if not src_path.is_symlink() and src_path.is_dir():
-            file_utils.link_or_copy_tree(src_path, dst_path)
+            file_utils.link_or_copy_tree(src_path, dst_path, overwrite_metadata=False)
             shutil.rmtree(src_path)
         else:
             src_path.unlink()
@@ -207,15 +209,32 @@ def organize_files(  # noqa: PLR0912, PLR0915
                             strict=dst_partition_pair.partition != OVERLAY_PARTITION,
                         )
                         if conflicts:
-                            raise errors.FileOrganizeError.from_merge_conflicts(
-                                part_name=part_name,
-                                key=key,
-                                destination=file_map[key],
-                                conflicts=conflicts,
-                            )
+                            if dst_partition_pair.partition == OVERLAY_PARTITION:
+                                conflicts = {
+                                    p: msg
+                                    for p, msg in conflicts.items()
+                                    if not (
+                                        (src_path / p).is_dir()
+                                        and not (src_path / p).is_symlink()
+                                        and (real_dst_path / p).is_dir()
+                                        and not (real_dst_path / p).is_symlink()
+                                    )
+                                }
+                            if conflicts:
+                                raise errors.FileOrganizeError.from_merge_conflicts(
+                                    part_name=part_name,
+                                    key=key,
+                                    destination=file_map[key],
+                                    conflicts=conflicts,
+                                )
                     # Where the key is a glob, we get the contents of the dir to
                     # organize, so we need to add the source's name back in.
-                    file_utils.link_or_copy_tree(src_path, real_dst_path)
+                    file_utils.link_or_copy_tree(
+                        src_path,
+                        real_dst_path,
+                        overwrite_metadata=dst_partition_pair.partition
+                        != OVERLAY_PARTITION,
+                    )
                     shutil.rmtree(src)
                     continue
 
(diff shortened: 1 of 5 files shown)


craft_parts/executor/organize.py
```diff
@@ -207,12 +209,24 @@ def organize_files(  # noqa: PLR0912, PLR0915
...
+                                conflicts = {
+                                    p: msg
+                                    for p, msg in conflicts.items()
+                                    if not (
+                                        (src_path / p).is_dir()
+                                        and not (src_path / p).is_symlink()
+                                        and (real_dst_path / p).is_dir()
+                                        and not (real_dst_path / p).is_symlink()
+                                    )
+                                }
```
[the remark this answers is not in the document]
lengau: This is true, we should still error out if metadata is different.

lengau:
@copilot apply changes based on the comments in [this thread](https://github.com/canonical/craft-parts/pull/1586#pullrequestreview-4307320414)

tests/unit/features/overlay_partitions/test_executor_organize.py
```diff
@@ -313,3 +314,97 @@ def test_organize_no_overwrite_idempotent(tmp_path, data):
...
+    overlay_dir = new_path / "overlay_dir"
+
+    install_dir_a.mkdir()
+    install_dir_b.mkdir()
+    overlay_dir.mkdir()
+
+    # Part A contributes a directory
+    (install_dir_a / "my-dir").mkdir(mode=0o755)
+    (install_dir_a / "my-dir" / "file-a").touch()
+
```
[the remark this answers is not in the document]
lengau: @copilot apply changes based on [this feedback](https://github.com/canonical/craft-parts/pull/1586#discussion_r3256940963)

lengau:
@copilot apply changes based on the comments in [this thread](https://github.com/canonical/craft-parts/pull/1586#pullrequestreview-4307320414)

craft_parts/executor/organize.py
```diff
@@ -207,15 +209,32 @@ def organize_files(  # noqa: PLR0912, PLR0915
...
+                                    conflicts=conflicts,
+                                )
                     # Where the key is a glob, we get the contents of the dir to
                     # organize, so we need to add the source's name back in.
-                    file_utils.link_or_copy_tree(src_path, real_dst_path)
+                    file_utils.link_or_copy_tree(
+                        src_path,
+                        real_dst_path,
+                        overwrite_metadata=dst_partition_pair.partition
+                        != OVERLAY_PARTITION,
```
tigarmo: this formatting is a bit unfortunate

craft_parts/utils/file_utils.py
```diff
@@ -193,6 +193,8 @@ def link_or_copy_tree(
     destination_tree: str | Path,
     ignore: Callable[[str, list[str]], list[str]] | None = None,
     copy_function: Callable[..., None] = link_or_copy,
+    *,
+    overwrite_metadata: bool = True,
```
tigarmo, line 197: the bit about overwriting the metadata feels off, can you give some more context?
The dirs on the parts.yaml reproducer on the original issue have the same metadata right? So which case is this meant to address?

And how do we know that *not* overwriting the metadata is the right call here?

lengau: We want to error out if the two directories have different metadata because it likely means the two separate parts created them with different metadata. In that case we don't want to select the metadata from a specific part, but rather want the user to ensure both parts have the same metadata. Take for example:

```yaml
parts:
  a:
    plugin: nil
    override-build: |
      mkdir $CRAFT_PART_INSTALL/my-dir
      chown 0 $CRAFT_PART_INSTALL/my-dir
    organize:
      '*': (overlay)/
  b:
    plugin: nil
    override-build: |
      mkdir $CRAFT_PART_INSTALL/my-dir
      chown 1000 $CRAFT_PART_INSTALL/my-dir
    organize:
      '*': (overlay)/
```

Should `my-dir` be owned by root or uid 1000?

tigarmo: ah I see now that the intent is to have it fail if the metadata is different, rather than leaving the metadata of the existing items. Maybe the docstring could be updated to reflect this?

---

Also, this rabbit hole makes me think that maybe organizing to the overlay should, itself, use overlays 🤯

tigarmo (approved):

craft_parts/utils/file_utils.py
lengau: I'm not too sure how that would work, since this is for creating the base layer.

PR #1588, opened 2026-05-20
Add Patches Directive

### LINKED ISSUE #1587: Add a Patches Directive to Allow Custom Patching on top of the Source Tree
### What needs to get done

It would be beneficial if there were a patching mechanism native to the craft-parts library such that when building an artifact, for example a rock, you could define custom patches that you would like to apply atop the source from which you are pulling. This can be useful for local development, as well as performing and testing backports.

### Why it needs to get done

In particular, sometimes customers hit issues in supported and released artifacts and we need to provide them with a patched version of the product. It can be messy to have loose patches floating around and then manually overriding the build to apply the patches. For example: see [this commit](https://github.com/canonical/metallb-rocks/commit/efed043c6655e05d7817db0638c61e071ec714a1#diff-61a282cc92dd2d8e7179253df11385f89e43dbe6dfef59cdebbe206dcac22c9a) to patch a rock, where the patch had to be added and specified in override-build. Note how the patch had to be manually applied with
```
      git am /root/parts/patches/src/0001-L2Status-controller-refactor-to-leader-node.patch
```

After also being manually dumped into the build container with a separate part defined by

```
  patches:
    plugin: dump
    source: .
```

Fortunately in this case, there was already an override-build section. However, if there wasn't already such a section, it could be difficult to figure out the correct expected build steps that the plugin applies by default and ensure the patch is applied at the right time.

The proposal of this idea is to add a patches directive that takes a list of patches that are present locally and automatically apply them in the defined order during the pull stage. This is more coherent with the Debian packaging process where distro patches are added to `debian/patches/`, and the order of patch application adheres to the list of patches defined in `debian/patches/series`

Maybe this functionality could also be extended to take in patches from remote sources or the local git tree, and the identifier could be a url, commit hash, or branch name (if all the patches were to be put in one branch). However, such extensibility would be quite complex and may require work to determine the details. For now, local patches would be a great start

bryanfraschetti:
Add a directive called patches which takes a list of filenames that should be applied on top of the source tree during the pull stage, similar to how Debian packaging enables patching through local files in debian/patches/ and a series file that defines the patch order.

Closes: #1587 

---

### DIFF
--- a/craft_parts/executor/source_patches.py
+++ b/craft_parts/executor/source_patches.py
@@ -1,0 +1,94 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+#
+# Copyright 2026 Canonical Ltd.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License version 3 as published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Helpers to apply source patches declared in part definitions."""
+
+import logging
+from pathlib import Path
+from typing import TextIO
+
+from craft_parts import errors
+from craft_parts.parts import Part
+from craft_parts.utils import process
+
+logger = logging.getLogger(__name__)
+
+Stream = TextIO | int | None
+
+
+def apply_source_patches(part: Part, *, stdout: Stream, stderr: Stream) -> None:
+    """Apply all patches configured for a part to its source tree."""
+    if not part.spec.patches:
+        return
+
+    for patch in part.spec.patches:
+        patch_path = Path(patch).expanduser()
+        if not patch_path.is_absolute():
+            patch_path = part.dirs.project_dir / patch_path
+
+        if not patch_path.is_file():
+            raise errors.PatchApplyError(
+                part_name=part.name,
+                patch=patch,
+                message=f"patch file not found: {patch_path}",
+            )
+
+        command = [
+            "patch",
+            "--strip",
+            "1",
+            "--forward",
+            "--batch",
+            "--input",
+            str(patch_path),
+        ]
+
+        logger.debug("Applying patch %r for part %r", patch_path, part.name)
+        try:
+            process.run(
+                command,
+                cwd=part.part_src_dir,
+                stdout=stdout,
+                stderr=stderr,
+            )
+        except process.ProcessError as process_error:
+            output = process_error.result.combined.decode("utf-8", errors="replace")
+            output_lower = output.lower()
+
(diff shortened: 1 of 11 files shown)


bryanfraschetti:
Some tests failed when running `make test` locally, particularly several TestSequencer suite tests failed with the error:
```
E         At index 0 diff: Action(part_name='bar', step=Step.PULL, action_type=ActionType.RERUN, reason="'patches' property changed", project_vars=ProjectVarInfo(root={}), properties=ActionProperties(changed_files=None, changed_dirs=None)) != Action(part_name='bar', step=Step.PULL, action_type=ActionType.SKIP, reason='
```
I figured I would solicit opinions on the proposed changes before fixing the tests, since the implementation may influence the tests

craft_parts/parts.py
```diff
@@ -167,6 +167,16 @@ class PartSpec(BaseModel):
...
     """
 
+    patches: list[str] = Field(
+        default=[],
+        description="A list of patch files to apply to the part's source.",
+        examples=[["patches/fix-build.patch", "patches/adjust-paths.patch"]],
+    )
+    """A list of patch files to apply to the part's source.
+
+    During the pull step, entries are applied to the pulled source tree.
```
medubelko, line 177:
```suggestion
    During the pull step, after the source is obtained, each patch file is applied 
    to the source tree.
```

medubelko (approved):
Thanks, @bryanfraschetti! I suggested some clarifying text in the description, but it's non-blocking.

craft_parts/parts.py
```diff
@@ -167,6 +167,16 @@ class PartSpec(BaseModel):
     This key does not affect commands specified with ``override-build``.
     """
 
+    patches: list[str] = Field(
```
cmatsuoka, line 170: I would call this `source-patches` and have them applied right after the source handler pulls the source. @lengau, @tigarmo, wdyt?

craft_parts/executor/source_patches.py
```diff
@@ -0,0 +1,94 @@
...
+            raise errors.PatchApplyError(
+                part_name=part.name,
+                patch=patch,
+                message=f"patch file not found: {patch_path}",
+            )
+
+        command = [
+            "patch",
+            "--strip",
+            "1",
```
cmatsuoka, line 52: `-p1` is fairly standard, but should this be the only possibility?

craft_parts/executor/source_patches.py
```diff
@@ -0,0 +1,94 @@
...
+                cwd=part.part_src_dir,
+                stdout=stdout,
+                stderr=stderr,
+            )
+        except process.ProcessError as process_error:
+            output = process_error.result.combined.decode("utf-8", errors="replace")
+            output_lower = output.lower()
+
+            # `patch --forward` reports already-applied patches as non-zero.
+            if "previously applied" in output_lower:
```
cmatsuoka, line 72: Is there a better way to check if previously applied? Relying on output messages from a specific implementation tends to cause unexpected problems.

cmatsuoka:
My overall feeling is that although this is good for patch application convenience, the design must be refined a bit to resolve the problem of `-p1` being hardcoded. I would suggest overriding pull for now and do something along the lines of https://github.com/cmatsuoka/mpg123-snap/blob/master/snap/snapcraft.yaml#L30 to get more insight about real use cases until we decide what to do.

craft_parts/parts.py
lengau: Agreed.

PR #1594, opened 2026-05-23
fix(partitions): allow plus signs in partition names

### LINKED ISSUE #1593: Allow the + character in partition names
### What needs to get done

Partition name validation should allow using the `+` character.

### Why it needs to get done

This is a valid package name in Debian packages. Partitions are used by Debcraft to create deb packages.

gcomneno:
## Summary

Allow `+` in partition names so Debian package names can be used as Debcraft partitions.

This keeps partition segment validation strict: `+` is allowed inside a segment, but not at the beginning or end of a segment. It also treats `+` like `-` and `/` when detecting partition environment variable name conflicts.

Fixes #1593.

## Tests

- `python -m pytest -q tests/unit/utils/test_partition_utils.py tests/unit/utils/test_path_utils.py tests/unit/features/partitions/test_parts.py -rs`


craft_parts/utils/partition_utils.py
```diff
@@ -22,16 +22,20 @@
 
 from craft_parts import errors, features
 
-# Allow alphanumeric characters, hyphens and slashes, not starting or ending
-# with a hyphen or a slash
-VALID_PARTITION_REGEX = re.compile(r"(?!-|/)[a-z0-9-/]+(?<!-|/)", re.ASCII)
+# Allow alphanumeric characters, hyphens, plus signs and slashes, not starting
+# or ending with a hyphen, a plus sign or a slash
```
bepri: This explanation for what's allowed (and all of its related user-facing messages) seems a bit unwieldy now. How about something like:

> Alphanumeric characters, hyphens, plus signs, and slashes are allowed. Names must begin and end with an alphanumeric character.

bepri (approved):
The code seems reasonable to me. Requesting @cmatsuoka's review for a second opinion.

gcomneno:
Thanks, I've simplified the partition name wording and kept the validation behavior unchanged.

gcomneno:
Good catch on the partition environment names. I've updated the environment variable translation so `+` is handled consistently with partition conflict detection, and added coverage for the generated `CRAFT_FOO_BAR_*` variables.

bepri (commented):
There's a number of failing tests relating to partitions. Could you take a look?

gcomneno:
Thanks for the heads-up. The failures were from a stale test expectation for the invalid partition-name message after the wording was clarified.

I've updated the lifecycle manager test expectation; validation remains unchanged.

gcomneno:
I checked the current failures.

The previous partition-related failures should be fixed by the latest branch update. The current non-OSV failures now appear unrelated to this PR: the failing jobs are hitting Ubuntu archive/chisel fetch errors with archive digest mismatches, and the remaining red jobs are cancellations from the same CI run.

The PR is currently mergeable and approved, so I don’t think there is any further code change needed from my side unless a rerun still shows partition-related failures.

cmatsuoka:
Consolidated in #1612

PR #1595, opened 2026-05-23
docs(python): clarify python-requirements usage

### LINKED ISSUE #1551: `python-requirements` key is missing important details
## Situation

The [`python-requirements` key](https://canonical-craft-parts.readthedocs-hosted.com/latest/common/craft-parts/reference/plugins/python_v2_plugin/#python-requirements) in the Python plugin references doesn't make it clear when it comes into play with the various dependency mechanisms: `requirements.txt`, `setup.cfg`, and `pyproject.toml`. Notably, it's not completely clear when a part _doesn't_ need to use the key.

## Request

Update all Python plugin refs to describe the conditions of the key in more detail.

### Why it needs to get done

More accuracy and completeness for the Python plugin reference.

gcomneno:
## Summary

Clarify when the `python-requirements` key is needed in the Python plugin references.

The updated docs explain that requirements files must be listed explicitly, and that the key is not needed when dependencies are already declared in project metadata used during package installation, such as projects installed through `setup.py` or `pyproject.toml`.

Fixes #1551.

## Tests

- `git diff --check`
- `python -m pytest -q tests/unit/plugins/test_python_plugin.py tests/unit/plugins/python_v2/test_python_plugin.py -rs`


gcomneno:
Heads up: the OSV Scanner failure does not seem related to this PR’s changes.

I checked the failed job logs and the source scan reports:

```text
osv-scanner source
No issues found
```

The failure appears to come from the generated exported requirements file:

requirements/uv-requirements.--all-extras.txt
idna==3.11
urllib3==2.6.3

with OSV reporting current advisories for those generated dependency versions.

I’m seeing the same OSV failure pattern on other unrelated open PRs as well, so this looks like a shared dependency-scan issue rather than a regression introduced here.

Possible fix direction: update or constrain the exported dependency set used by the OSV scan so it no longer resolves to the affected idna / urllib3 versions, or adjust the scan policy if these are transient/unfixed advisories from generated CI artifacts rather than source dependencies.

gcomneno:
Friendly ping on this PR. It has been idle for a while, and it should still be ready from my side.

Please let me know if you would like me to rebase, refresh the checks, or adjust anything else.

Thanks!

medubelko:
Sorry @gcomneno, this one didn't auto-ping the engineers.

@mr-cal Could you give this a quick technical review?

lengau (approved):
LGTM, thanks!

medubelko (approved):

PR #1596, opened 2026-05-23
fix(organize): nest overlay directory globs

### LINKED ISSUE #1590: inconsistency in organize to overlay
### Bug Description

When organizing a directory to a directory, craft-parts exhibits two different behaviors: MERGE and NEST, depending on the presence of glob keys. If the source is a glob key and destination is the `(overlay)` pseudo-partition the behavior is different from all other glob key cases, as shown in the table below:

| Key type         | Destination type  | Behavior |
|------------------|-------------------|----------|
| Explicit (no *)  | Any               | MERGE    |
| Glob (*)         | Partition root    | NEST     |
| Glob (*)         | Overlay partition | MERGE    |
| Glob (*)         | Regular path      | NEST     |

The inconsistency is at lines 199-202 of organize.py:

    if dst_path in install_dir_map.values():
        real_dst_path = dst_path / src_path.name
    else:
        real_dst_path = dst_path

The "else" branch handles the overlay partition case and sets
real_dst_path to dst_path directly, causing a MERGE instead of a NEST.

To make the overlay partition consistent with other destinations, the
fix would be to also nest the source directory name for the overlay
partition:

    if dst_path in install_dir_map.values():
        real_dst_path = dst_path / src_path.name
    elif dst_partition_pair.partition == OVERLAY_PARTITION:
        real_dst_path = dst_path / src_path.name
    else:
        real_dst_path = dst_path

This ensures that a glob key always nests, regardless of whether the
destination is a partition root, overlay partition, or regular path.

The only case that should MERGE is when the key is explicit (no *),
since the user is explicitly naming the source directory and
specifying where its contents should go.

[organize_behavior.txt](https://github.com/user-attachments/files/28116512/organize_behavior.txt)

### To Reproduce

Create a part that organizes from a glob key to a directory under `(overlay)/`.

### part yaml

```shell

```

### Relevant log output

```shell
N/A
```

gcomneno:
## Summary

Make directory glob organize behavior consistent when targeting the overlay pseudo-partition.

Previously, globbed directories organized to `(overlay)/...` were merged directly into the destination, unlike globbed directories organized to partition roots or regular paths. This now nests the source directory name for overlay destinations too.

Fixes #1590.

## Tests

- `python -m pytest -q tests/unit/features/overlay_partitions/test_executor_organize.py -rs`
- `git diff --check`

### DIFF
--- a/craft_parts/executor/organize.py
+++ b/craft_parts/executor/organize.py
@@ -198,10 +198,7 @@ def organize_files(  # noqa: PLR0912, PLR0915
                     dst_path in install_dir_map.values()
                     or dst_partition_pair.partition == OVERLAY_PARTITION
                 ):
-                    if dst_path in install_dir_map.values():
-                        real_dst_path = dst_path / src_path.name
-                    else:
-                        real_dst_path = dst_path
+                    real_dst_path = dst_path / src_path.name
                     if not overwrite:
                         conflicts = file_utils.find_merge_conflicts(
                             src_path,
--- a/tests/unit/features/overlay_partitions/test_executor_organize.py
+++ b/tests/unit/features/overlay_partitions/test_executor_organize.py
@@ -250,10 +250,12 @@ def test_organize(new_dir, data):
                 },
                 "expected": [
                     (["dir"], "../overlay_dir"),
-                    (["child"], "../overlay_dir/dir"),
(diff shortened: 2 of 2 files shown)


cmatsuoka:
Superseded by #1602

PR #1597, opened 2026-05-23
docs(dump): add usage examples

gcomneno:
## Summary

Add practical Dump plugin examples showing how to copy a local directory as-is, copy files to specific locations, organize directory contents, and copy files into a charm payload.

Fixes #1531.

## Tests

- `git diff --check`
- `make docs`


docs/common/craft-parts/reference/plugins/dump_plugin.rst
```diff
@@ -40,21 +40,74 @@ During the build step, the plugin performs the following actions:
 #. Copy all contents and preserve the directory structure to the part's install
    directory.
 
-Example
--------
+Examples
+--------
```
jahn-junior, line 44: Several of the added examples are imperative, making this read as if its a dump plugin how-to. Any examples added here should be declarative.

docs/common/craft-parts/reference/plugins/dump_plugin.rst
```diff
@@ -40,21 +40,74 @@ During the build step, the plugin performs the following actions:
...
 
 .. code-block:: yaml
 
   parts:
     config:
       plugin: dump
-      after:
-        - nginx
-      source: .
+      source: ./files
```
jahn-junior, line 74: Was there a reason behind the change in source? I'd like if one of the examples pulled files from the directory containing the project file.

docs/common/craft-parts/reference/plugins/dump_plugin.rst
```diff
@@ -40,21 +40,74 @@ During the build step, the plugin performs the following actions:
...
+      organize:
+        my-app.service: usr/lib/systemd/system/my-app.service
+        my-app.conf: etc/my-app/my-app.conf
+
+Copy directories with ``organize``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
jahn-junior, line 80: I don't think the distinction between organizing files and directories is meaningful enough to warrant another example.

docs/common/craft-parts/reference/plugins/dump_plugin.rst
```diff
@@ -40,21 +40,74 @@ During the build step, the plugin performs the following actions:
...
+  parts:
+    resources:
+      plugin: dump
+      source: ./resources
+      organize:
+        scripts/*: usr/bin/
+        templates/*: usr/share/my-app/templates/
+
+Copy files into a charm payload
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
jahn-junior, line 97: This page is consumed in its entirety by several downstream applications, so we shouldn't make references to specific artifacts. The references to charms were carried over when the issue was transferred from Charmcraft.

The previous examples also already demonstrate the same keys used here.

docs/common/craft-parts/reference/plugins/dump_plugin.rst
jahn-junior, line 1: @cmatsuoka I'm curious to hear what use cases you think would be best to cover here. You'd know better than I.

jahn-junior (changes requested):
Thanks for the contribution! I pointed out some changes I'd like to see regarding the tone of the prose, as well as some redundancy between the added examples.

I'll hold off on a more detailed pass until @cmatsuoka has a chance to share his thoughts.

gcomneno:
Addressed the review feedback in b6f3d726: the examples are now declarative, one keeps `source: .`, the redundant directory-specific example was removed, and the charm-specific example was dropped.

Local `make docs` and `git diff --check` pass. The required CI is green except for `test-common / Minimum dependencies (all tests)`, which fails on `os.mknod(..., S_IFCHR, ...)` with `Operation not permitted`; I do not have permission to rerun the upstream workflow.

Happy to adjust further once @cmatsuoka has had a chance to weigh in.

docs/common/craft-parts/reference/plugins/dump_plugin.rst
cmatsuoka: The supplied example is a typical real life use case and looks appropriate for the documentation.

cmatsuoka (approved):

jahn-junior (approved):
LGTM. Thanks for incorporating my suggestions!

jahn-junior:
Lint failure is unrelated to these changes. It's because we haven't incorporated https://github.com/canonical/starbase/pull/551 into these docs.

PR #1598, opened 2026-05-23
feat(organize): support build pseudo-partition source

gcomneno:
## Summary

Add support for using `(build)` as an `organize` source pseudo-partition.

This allows files generated in the part build directory to be copied into install partitions during build organization, without moving/removing the original build artifacts.

The change keeps `(build)` source-only:
- `(build)` is accepted as an `organize` source
- `(build)` is rejected as an `organize` destination
- `build` is reserved and cannot be used as a real partition name
- `part_install_dirs` is not globally polluted with `build`; the build directory is only injected into the organize call path

## Tests

- `git diff --check`
- `uv run --group lint --group types ruff check ...`
- `uv run --group lint --group types ruff format --check ...`
- `uv run --group dev python -m pytest -q tests/integration/features/partitions/lifecycle/test_lifecycle.py::test_organize_from_build_with_partitions -rs`
- `uv run --group dev python -m pytest -q tests/unit/executor/test_organize.py tests/unit/features/partitions/executor/test_organize.py tests/unit/features/overlay_partitions/test_executor_organize.py tests/unit/executor/test_part_handler.py tests/unit/features/partitions/executor/test_part_handler.py tests/unit/features/partitions/test_parts.py tests/unit/features/overlay_partitions/test_parts.py tests/unit/test_infos.py tests/unit/utils/test_partition_utils.py -rs`

Final targeted result: `1107 passed, 1 xfailed`.

---
CRAFT-5181


craft_parts/executor/organize.py
```diff
@@ -318,23 +328,30 @@ def organize_files(  # noqa: PLR0912, PLR0915
...
     """Return the full path for a relative source."""
-    src_partition, src_inner_path = path_utils.get_partition_and_path(
-        key, default_partition
-    )
+    src_partition, src_inner_path = src_partition_path
 
     if src_partition and src_partition not in [
         default_partition,
         DEFAULT_PARTITION,
+        BUILD_PARTITION,
```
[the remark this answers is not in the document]
lengau: ^ This is correct. We need to avoid modifying the build directory at all.

gcomneno:
Thanks, this was a real gap. I pushed a fix so `(build)` sources are preserved in the overlay equivalence and symlink paths as well, and added targeted overlay organize tests covering equivalent overlay files and build symlinks.

Validation:
- `uv run ruff format craft_parts/executor/organize.py tests/unit/features/overlay_partitions/test_executor_organize.py`
- `uv run ruff check craft_parts/executor/organize.py tests/unit/features/overlay_partitions/test_executor_organize.py`
- `uv run pytest -q tests/unit/features/overlay_partitions/test_executor_organize.py -rs` → 120 passed

craft_parts/executor/organize.py
```diff
@@ -112,7 +116,14 @@ def organize_files(  # noqa: PLR0912, PLR0915
         the default partition.
     """
     for key in sorted(file_map, key=lambda x: ["*" in x, x]):
-        src = get_src_path(key, part_name, install_dir_map, default_partition)
+        src_partition_pair = path_utils.get_partition_and_path(key, default_partition)
+        src = get_src_path(
+            src_partition_pair,
+            key,
```
cmatsuoka: Since you're passing the source partition pair, it shouldn't be necessary to pass the key (the error message in ` get_src_path` can get the partition from the pair).

craft_parts/executor/organize.py
```diff
@@ -121,7 +132,7 @@ def organize_files(  # noqa: PLR0912, PLR0915
             file_map[key].lstrip("/"), default_partition
         )
         dst, dst_string = get_dst_path(
-            key, file_map, install_dir_map, default_partition
+            key, file_map, part_name, install_dir_map, default_partition
```
cmatsuoka: For symmetry, we can also get `dst_partition_pair` and pass it to `get_dst_path` replacing `key` and `file_map`.

tests/integration/features/partitions/lifecycle/test_lifecycle.py
```diff
@@ -377,3 +377,56 @@ def test_partition_symlinks_default_partition(new_dir):
         ctx.execute(actions)
 
     assert sorted(next(os.walk(Path("partitions")))[1]) == ["binaries", "docs"]
+
+
+def test_organize_from_build_with_partitions(new_dir):
```
cmatsuoka, line 382: Please also add some test cases to ‎`tests/unit/executor/test_organize.py` and `‎tests/unit/features/overlay_partitions/test_executor_organize.py` organizing from `(build)/`, including cases where the destination directory doesn't exist (it should be created). This will require a change to `organize_and_assert` to allow setup of `build_files` as well.

cmatsuoka (changes requested):

tests/integration/features/partitions/lifecycle/test_lifecycle.py
cmatsuoka: The test cases can cover both explicit directories from `(build)/` and also globs when the target directory already exists, since the behavior for these two cases is different (merge vs nest).

gcomneno:
Thanks, I pushed updates for the requested changes.

Summary:
- refactored `get_src_path()` so it no longer needs the original organize key when the source partition pair is already available;
- refactored `get_dst_path()` to take the destination partition pair directly, matching the source-side flow;
- added `(build)` parsing coverage in `test_path_utils.py`;
- extended `organize_and_assert()` so unit tests can set up files and directories in the build pseudo-partition;
- added unit coverage in `tests/unit/executor/test_organize.py` for `(build)/` sources, including destination directory creation, explicit directory merge, and glob directory nesting;
- added overlay partition coverage in `tests/unit/features/overlay_partitions/test_executor_organize.py`.

Validation:
- `ruff check` on touched files: passed
- `pyright` on touched files: 0 errors
- `pytest -q tests/unit/utils/test_path_utils.py tests/unit/executor/test_organize.py tests/unit/features/overlay_partitions/test_executor_organize.py -rs`: 403 passed

cmatsuoka (approved):
Thank you, this addresses the shortcomings of #1535 and will allow debcraft to use mainline craft-parts.

lengau (approved):
Sorry about the delay here — I wanted to check this in imagecraft. LGTM!

PR #1599, opened 2026-05-24
fix(sources): detect local file source changes

### LINKED ISSUE #1364: The dump plugin ignores updates to "source-type: file"
### Bug Description

The dump plugin ignores updates to the file.

### To Reproduce

```
parts:
  test-service:
    plugin: dump
    source: test-service.service
    source-type: file
```

1. Prime this part.
2. Change test-service.service and try to prime (pack) it again. The resulting content will not be updated.

### part yaml

```shell

```

### Relevant log output

```shell
N/A
```

gcomneno:
## Summary

Fixes #1364.

This adds change detection support for local file sources handled by `FileSourceHandler`.

When a local file source is newer than the pull state file, the source handler now reports it as outdated, exposes the changed file through `get_outdated_files()`, and updates the pulled source by re-running the file source pull path.

Remote URL-based file sources are left unchanged and are not treated as locally outdated. `file://` sources are treated as local files for change detection.

## Validation

- `uv run pytest -q tests/unit/sources/test_base.py -rs`
- `uv run pytest -q tests/unit/sources/test_base.py tests/unit/sources/test_file_source.py tests/unit/sources/test_tar_source.py tests/unit/sources/test_zip_source.py tests/unit/sources/test_deb_source.py tests/unit/sources/test_rpm_source.py tests/unit/sources/test_snap_source.py tests/unit/sources/test_sevenzip_source.py -rs`
- `uv run ruff check craft_parts/sources/base.py tests/unit/sources/test_base.py`
- `uv run ruff format --check craft_parts/sources/base.py tests/unit/sources/test_base.py`
- `git diff --check`


craft_parts/sources/base.py
```diff
@@ -238,9 +238,49 @@ def __init__(
...
 
     # pylint: enable=too-many-arguments
 
+    @override
+    def check_if_outdated(
+        self, target: str, *, ignore_files: list[str] | None = None
+    ) -> bool:
+        """Check if the source file changed since target was created."""
+        if url_utils.is_url(self.source):
+            return False
```
lengau: Won't this result in the same issue if the source is a `file:///` url?

craft_parts/sources/base.py
```diff
@@ -238,9 +238,49 @@ def __init__(
...
+        if source_mtime >= target_mtime:
+            self._updated_files.add(source_name)
+            return True
+
+        return False
+
+    @override
+    def get_outdated_files(self) -> tuple[list[str], list[str]]:
+        """Obtain lists of outdated files and directories."""
+        return (sorted(self._updated_files), sorted(self._updated_directories))
```
lengau, line 283: This depends on `check_if_outdated` having been run first. Is that guaranteed?

gcomneno:
Addressed the review feedback in the latest commit.

Changes:
- `file://` sources are now treated as local files when checking whether a file source is outdated.
- Remote URLs still return `False` for local change detection.
- The outdated file/directory tracking sets are reset at the start of `check_if_outdated()` to avoid stale state.
- `get_outdated_files()` now documents that `check_if_outdated()` should be called first, matching the existing source update flow.

Validation:
- `uv run pytest -q tests/unit/sources/test_base.py -rs` → 18 passed
- `uv run ruff check craft_parts/sources/base.py tests/unit/sources/test_base.py`
- `uv run ruff format craft_parts/sources/base.py tests/unit/sources/test_base.py`
- `git diff --check`

craft_parts/sources/base.py
gcomneno: Yes. I rechecked the current lifecycle to make sure this is guaranteed rather than just relying on the docstring.

`StateManager.check_if_outdated()` calls `source_handler.check_if_outdated()` first and only calls `get_outdated_files()` inside the true branch:

```python
if source_handler.check_if_outdated(str(state_file)):
    files, dirs = source_handler.get_outdated_files()
```

The update paths follow the same ordering: `PartHandler` calls `check_if_outdated()` before `update()` (and the local-source path explicitly notes that this is required by `source.update()`).

So `get_outdated_files()` is consumed after the handler has populated its update state in the lifecycle paths that use it.

lengau (approved):
Thanks!

PR #1602, opened 2026-05-27
fix(organize): make organize to overlay behavior consistent

cmatsuoka:
An inconsistency was introduced in commit a1178308 (PR #1541) that caused
glob-key organizing to the overlay merge directory contents instead of
nesting. The expected behavior is to merge when explicit keys are used
and nest when globs are used. Fixed the organize logic to make glob-key
organizing to overlay nest instead of merge to remain consistent with
the results of organizing to other targets.

---

### DIFF
--- a/craft_parts/executor/organize.py
+++ b/craft_parts/executor/organize.py
@@ -188,7 +188,7 @@ def organize_files(  # noqa: PLR0912, PLR0915
             # Organize a dir to a dir
             if not src_path.is_symlink() and src_path.is_dir():
                 # When an explicit key is used the source directory is moved
-                # directly to the destination path, replacing it. Multiple
+                # directly to the destination path, replace it. Multiple
                 # explicit keys mapping to the same destination merge their contents.
                 # If a glob key is used, each matching directory is nested inside
                 # the destination preserving its name.
--- a/craft_parts/executor/organize.py
+++ b/craft_parts/executor/organize.py
@@ -187,6 +187,11 @@ def organize_files(  # noqa: PLR0912, PLR0915
 
             # Organize a dir to a dir
             if not src_path.is_symlink() and src_path.is_dir():
+                # When an explicit key is used the source directory is moved
+                # directly to the destination path, replacing it. Multiple
+                # explicit keys mapping to the same destination merge their contents.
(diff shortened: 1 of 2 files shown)


cmatsuoka:
The Chisel test errors are unrelated.

bepri (approved):

craft_parts/executor/organize.py
```diff
@@ -187,6 +187,11 @@ def organize_files(  # noqa: PLR0912, PLR0915
 
             # Organize a dir to a dir
             if not src_path.is_symlink() and src_path.is_dir():
+                # When an explicit key is used the source directory is moved
+                # directly to the destination path, replacing it. Multiple
```
mr-cal:
```suggestion
                # directly to the destination path, replace it. Multiple
```

mr-cal (approved):

tests/unit/features/overlay_partitions/test_executor_organize.py
```diff
@@ -250,10 +250,12 @@ def test_organize(new_dir, data):
...
                 "expected": [
                     (["dir"], "../overlay_dir"),
-                    (["child"], "../overlay_dir/dir"),
+                    (["dir1", "dir2"], "../overlay_dir/dir"),
+                    (["child"], "../overlay_dir/dir/dir1"),
+                    (["child"], "../overlay_dir/dir/dir2"),
                 ],
             },
-            id="merge-dir-globs-in-overlay-success",
+            id="nest-dir-globs-in-overlay-success",
```
lengau, line 258: How does this behave with two parts, each of which organize as:

```yaml
organize:
  '*': (overlay)/
```

cmatsuoka: This should be the same behavior as organizing to a partition or a different location in the default partition.

cmatsuoka: Specifically the `test_organize_glob_to_overlay_with_shared_directory_across_parts` was not changed and it didn't fail.

lengau: Thanks, I didn't remember whether that test had made its way to main.

lengau (approved):

PR #1617, opened 2026-06-11
fix: migrate overlay files before running stage or prime

tigarmo:
This commit switches the order of calling _run_step() and _migrate_overlay* so that the overlay migration always happens *before* the default step behavior. This is the desired behavior, but the previous code had a corner-case when a part has content coming from both 'build' and 'overlay', like this:

```yaml
parts:
  p1:
    override-build: |
      mkdir ${CRAFT_PART_INSTALL}/my-dir
    override-overlay: |
      mkdir my-dir
      touch my-dir/my-file
```

... in this situation, the previous code would *first* stage the `my-dir`directory coming from `override-build` and *then* migrate `my-dir` and `my-dir/my-file` from the overlay to stage. Because of the way the migration happens, the `my-dir` dir coming from build would "shadow" the `my-dir` dir coming from stage, which meant the migrate state would *not* record `my-dir` as coming from the overlay. This is a bug in and of itself but it was usually harmless.

A failure happens if another, separate part gets primed *before* `p1`. In that case, we would try to do the migration of overlay contents from stage to prime, which in this case meant migrating the `my-dir/my-file` file but **not** the `my-dir` dir because it did not get recorded. Then, the migration of `my-dir/my-file` would *maybe* work:

- if `my-file` is a regular file, the code would follow a path that recreates the directory structure to hardlink the file (to save storage);
- if `my-file` is a special file like a symlink the code would try to copy it to `${prime}/my-dir/`, which doesn't exist. Then a failure happens with a cryptic message that suggests that `${stage}/my-dir/my-file` doesn't exist, but this is just bad formatting of the error message.

In my view the main cause of this bug is the fact that overlay files are not correctly recorded in the state file - I added a test to check this scenario and I think it's sufficient to validate this particular bugfix.
---

### DIFF
--- a/craft_parts/executor/part_handler.py
+++ b/craft_parts/executor/part_handler.py
@@ -546,6 +546,8 @@ class PartHandler:
         """
         self._make_dirs()
 
+        self._migrate_overlay_files_to_stage()
+
         contents = self._run_step(
             step_info=step_info,
             scriptlet_name="override-stage",
@@ -553,8 +555,6 @@ class PartHandler:
             stdout=stdout,
             stderr=stderr,
         )
-
-        self._migrate_overlay_files_to_stage()
 
         # Overlay integrity is checked based by the hash of its last (topmost) layer,
         # so we compute it for all parts. The overlay hash is added to the stage state
@@ -603,6 +603,8 @@ class PartHandler:
         """
         self._make_dirs()
 
+        self._migrate_overlay_files_to_prime()
+
         contents = self._run_step(
             step_info=step_info,
             scriptlet_name="override-prime",
@@ -611,7 +613,6 @@ class PartHandler:
             stderr=stderr,
         )
 
-        self._migrate_overlay_files_to_prime()
         default_partition = self._part_info.default_partition or DEFAULT_PARTITION
 
         primed_stage_packages: set[str]
--- a/docs/reference/changelog.rst
+++ b/docs/reference/changelog.rst
@@ -19,6 +19,24 @@ Changelog
 
   For a complete list of commits, check out the `X.Y.Z`_ release on GitHub.
 
+
+.. _release-2.35.0:
+
+2.35.0 (unreleased)
+-------------------
+
+New features:
+
+- The plus sign is allowed in partition names. More restrictive rules must be
+  enforced at application level.
+
+Bug fixes:
+
+- Always migrate overlay files to Stage and Prime first, even if the part being staged
+  or primed also has contents coming from the install directory.
+
+For a complete list of commits, check out the `2.35.0`_ release on GitHub.
+
 .. _release-2.34.0:
 
 2.34.0 (2026-06-10)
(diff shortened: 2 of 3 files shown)


docs/reference/changelog.rst
```diff
@@ -19,6 +19,19 @@ Changelog
 
   For a complete list of commits, check out the `X.Y.Z`_ release on GitHub.
 
+
+.. _release-2.34.1:
+
+2.34.1 (2024-06-15)
```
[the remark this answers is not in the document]
tigarmo: true! In fixing this I noticed that an entryh for 2.34 actually hasn't been released yet, so I moved it over to the new 2.35.0

tests/unit/features/overlay/test_executor_part_handler.py
```diff
@@ -741,6 +741,57 @@ def test_migrate_overlay_filter_whiteout(self, mocker, new_dir, partitions):
         assert Path("prime/.wh.file1").exists() is False
         assert Path("prime/.wh.file2").exists()
 
+    def test_migrate_overlay_with_install(self, mocker, new_dir, partitions):
+        """Test that overlay contents are correctly migrated when the part has
+        relevant contents coming from the install dir."""
```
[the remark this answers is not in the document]
tigarmo: It's true that I didn't add a test for the specific circumstance described in the PR, but in my opinion that scenario is just one possible consequence of the actual bug - the incorrect recording of overlay contents on state files. I added a test that validates the "core" of the bugfix and I think that's enough; I'm going to add a regression test in Rockcraft when the bugfix is released there.

craft_parts/executor/part_handler.py
```diff
@@ -611,7 +613,6 @@ def _run_prime(
             stderr=stderr,
         )
 
-        self._migrate_overlay_files_to_prime()
```
lengau, line 614: How does this affect overlay files with `prime: [-my/file]`

cmatsuoka: Is this scenario covered by a test? It seems that only a couple of (flaky) arm tests failed.

cmatsuoka (approved):
The overlays spec doesn't mandate a staging order, but the ordering proposed is in this PR makes the content migrated from overlay visible to override-stage (which is the correct thing to happen since we also see the content staged from build). The fix looks good, just please add a test covering the case @lengau mentioned to make sure we didn't change this behavior.

craft_parts/executor/part_handler.py
tigarmo: I'll add a test but I'll preface this by saying that I don't think either the `stage` or `prime` keys affect content coming from the overlay. I built this with stable rockcraft (so without these changes):
```yaml
parts:
  my-part:
    plugin: nil
    override-overlay: |
      echo "from-overlay" > from-overlay.txt
    override-build: |
      echo "from-build" > from-build.txt
    prime:
      - "-from-overlay.txt"
      - "-from-build.txt"
```

... and after priming `from-build.txt` is gone but `from-overlay.txt` isn't:

```
# ls -l /root/prime
total 20
drwxr-xr-x 2 root root 4096 Jun 16 09:10 etc
-rw-r--r-- 3 root root   13 Jun 16 09:10 from-overlay.txt
drwxrwxrwt 2 root root 4096 Jun 16 09:10 tmp
drwxr-xr-x 3 root root 4096 Jun 16 09:10 usr
drwxr-xr-x 3 root root 4096 Jun 16 09:10 var
```

tigarmo: added the test

lengau (approved):
Haha! Yes, thanks for confirming the weirdness 😊

PR #1630, opened 2026-07-07
feat(PROTOTYPE): root_dir for handling monorepos

lengau:

---

These are the changes that Copilot thought necessary in order to add separate root directory handling for monorepos.

### DIFF
--- a/craft_parts/dirs.py
+++ b/craft_parts/dirs.py
@@ -28,9 +28,13 @@ class ProjectDirs:
 
     :param work_dir: The parent directory containing the parts, prime and stage
         subdirectories. If not specified, the current directory will be used.
+    :param root_dir: The root directory of the project (e.g. the git repository
+        root). When set, local sources are staged from this directory so sibling
+        packages outside ``work_dir`` are accessible during the build.
     :param partitions: If partitions are enabled, the list of partitions.
 
     :ivar work_dir: The root of the work directories used for project processing.
+    :ivar root_dir: The project root directory, if set.
     :ivar parts_dir: The directory containing work subdirectories for each part.
     :ivar overlay_dir: The directory containing work subdirectories for overlays.
     :ivar overlay_mount_dir: The mountpoint for the overlay filesystem.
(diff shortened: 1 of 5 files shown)


tigarmo (commented):
Thanks for the prototype. I do think this merits a spec, with examples and possibly some diagrams (I'm a visual learner ;) )

craft_parts/lifecycle_manager.py
```diff
@@ -400,3 +408,27 @@ def _validate_part_dependencies(part: Part, parts_data: dict[str, Any]) -> None:
     for name in part.dependencies:
         if name not in parts_data:
             raise errors.InvalidPartName(name)
+
+
+def _rewrite_local_sources(parts_data: dict[str, Any], root_dir: Path) -> None:
+    """Rewrite local sources so they're staged from root_dir with source-subdir set.
+
+    This preserves sibling directories (e.g. ../common) in the staging area,
+    which would otherwise be inaccessible when only the charm subdir is staged.
```
cmatsuoka, line 417: This should be "project subdir" since we're not charm-specific.

craft_parts/dirs.py
```diff
@@ -45,10 +49,12 @@ def __init__(
         *,
         partitions: Sequence[str] | None = None,
         work_dir: Path | str = ".",
+        root_dir: Path | None = None,
     ) -> None:
         partition_utils.validate_partition_names(partitions)
         self.project_dir = Path().expanduser().resolve()
         self.work_dir = Path(work_dir).expanduser().resolve()
+        self.root_dir = root_dir
```
cmatsuoka, line 57: Should `root_dir` be set to the regular project dir if we're not in a monorepo, and have a single codepath from here? Then we can have monorepo as a general supported feature and non-monorepo as a special case where repo and project dirs happen to be the same.

craft_parts/lifecycle_manager.py
```diff
@@ -400,3 +408,27 @@ def _validate_part_dependencies(part: Part, parts_data: dict[str, Any]) -> None:
...
+    """
+    for spec in parts_data.values():
+        if not isinstance(spec, dict):
```
cmatsuoka, line 420: Instead of rewriting parts data, maybe we could make  root_dir really first class and pass it along so operations that need to be carried on the repo root will do it without "rebasing" the project in pre-processing? This is just an idea, unsure how complex would be the refactoring to propagate root awareness end-to-end.

lengau: For the real implementation I would prefer to do that if we can. I'll have the same LLM that implemented this give it a go on a separate branch to see about the complexity.

craft_parts/dirs.py
lengau: This was my design decision, as I thought having the root directory be separate might be valuable. In retrospect looking at the implementation I'm not sure it is, but I'm not confident either way.

What if we made a property on this that selects the root dir if it's a path or the project dir otherwise and make that the way to get it?

craft_parts/lifecycle_manager.py
lengau: Trying that on https://github.com/canonical/craft-parts/pull/1635

lengau:
Closing this since we've decided to use relative parent paths for this instead.

PR #1631, opened 2026-07-07
docs: document autotools disable-parallel key

### LINKED ISSUE #1603: docs: for the autotools plugin relevant flags are not described, e.g. disable_parallel
*Please describe the question or issue you're facing with "Autotools plugin - Snapcraft 9.0 documentation".*

I tried to understand if the autotools plugin uses parallel builds. But this information is lacking.

There is even an option [disable_parallel](https://github.com/canonical/craft-parts/blob/bbc6f9f163be0f95e9c3fd99dba170d725e783ce/craft_parts/plugins/autotools_plugin.py#L113) which is not described.

Please, describe the option.

---
*Reported from: https://documentation.ubuntu.com/snapcraft/stable/common/craft-parts/reference/plugins/autotools_plugin/

GreyViperTooth:
The [Autotools plugin reference](https://documentation.ubuntu.com/craft-parts/en/latest/common/craft-parts/reference/plugins/autotools_plugin/)
didn't mention that the plugin builds in parallel, nor did it document the
`disable-parallel` key, even though the option is supported by the plugin.

This PR documents both:

- Adds a `disable-parallel` key entry under **Keys** (type boolean, default
  `false`), explaining that the plugin runs `make` with `-j` set to
  `CRAFT_PARALLEL_BUILD_COUNT` by default, and that setting the key to `true`
  forces a serial build for projects whose build system isn't parallel-safe.
- Updates the **How it works** section so the `make` step notes the parallel
  behavior and the opt-out.
- Adds a changelog entry under the Unreleased section.

Docs only — no functional changes.

Fixes #1603


docs/common/craft-parts/reference/plugins/autotools_plugin.rst
```diff
@@ -73,7 +88,8 @@ During the build step the plugin performs the following actions:
      ``bootstrap`` with any set ``autotools-bootstrap-parameters`` without *configuring*
      the project.
 #. Call ``configure`` with any set ``autotools-configure-parameters``.
-#. Call ``make`` to build.
+#. Call ``make`` to build. The build runs in parallel with ``-j`` set to
```
bepri, line 91: I'm on the fence about this particular change and I'm wondering what others think. It could be useful information for one who's just skimming the page, but it feels like it describes the plugin's inner workings in more depth than necessary here.

I could go either way (hence the approval) but I was uncertain enough that I thought I'd leave a note for others.

bepri (approved):
Thanks for the contribution, @GreyViperTooth!

Approving with minor question for other reviewers.

docs/common/craft-parts/reference/plugins/autotools_plugin.rst
```diff
@@ -40,6 +40,21 @@ autotools-configure-parameters
...
+
+**Default:** false
+
+Whether to disable parallel execution of ``make`` during the build step.
+
+By default, the plugin builds in parallel by running ``make`` with ``-j`` set to
+the number of concurrent jobs given by
+:ref:`CRAFT_PARALLEL_BUILD_COUNT <craft_parts_step_execution_environment>`. Set
+this key to ``true`` to run ``make`` without the ``-j`` flag, forcing a serial
+build. This is useful for projects whose build system is not parallel-safe.
```
jahn-junior, line 56: This key wasn't previously documented here because it isn't specific to the plugin. This behavior is important to capture, but if it's relevant to all parts, it should be documented in the [Part properties](https://documentation.ubuntu.com/craft-parts/latest/common/craft-parts/reference/part_properties/#disable-parallel) reference.

@bepri @cmatsuoka Does this key behave consistently across all plugin build steps? I'm wondering if this could be tweaked and incorporated into the [key's docstring](https://github.com/canonical/craft-parts/blob/main/craft_parts/parts.py#L251-L254), which is rendered in the craft apps' project file references.

docs/common/craft-parts/reference/plugins/autotools_plugin.rst
jahn-junior: I'm not necessarily a proponent of it, but these steps are already very close to the machinery for most plugins. However, I do think that calling out the specific option is probably more detail than users will need.

Would you have the same concern if the text read:

> Jobs are run in parallel unless the ``disable-parallel`` key is set to ``true``.

jahn-junior (commented):
Thanks for the contribution!

I think you've highlighted some important details, but I want to make sure they land where they can have the most impact.

docs/common/craft-parts/reference/plugins/autotools_plugin.rst
lengau: It does not act consistently unfortunately.

docs/common/craft-parts/reference/plugins/autotools_plugin.rst
lengau: Yeah I'd rather just leave it as it was.

docs/common/craft-parts/reference/plugins/autotools_plugin.rst
cmatsuoka: It should (ideally) be used by build systems that control parallelism, but not all of them support that. Also setting `...PARALLEL_BUILD_COUNT` to `true` to run without parallelism doesn't sound right to me. Setting to `1` sounds much more natural.

lengau:
@GreyViperTooth just checking up on this PR.

GreyViperTooth:
Thanks for the nudge @lengau

On @bepri's thread: agreed, I'll revert the "How it works" line back to
`#. Call ``make`` to build.`

On the placement question (@jahn-junior): how would you like me to proceed?

1. Keep it in the autotools reference (optionally with a one-line "autotools-only"
   note), or document it more broadly?
2. Want the codebase docstring updated in this PR too

docs/common/craft-parts/reference/plugins/autotools_plugin.rst
medubelko: We have sufficient information about parallel builds sprinkled in the project file, part env var, and library references of the apps. If we want to explain this as global behaviour, I recommend https://documentation.ubuntu.com/craft-parts/latest/common/craft-parts/explanation/parts/#build-process.

PR #1650, opened 2026-07-17
docs: group plugins by language

### LINKED ISSUE #1571: Group related plugins
### What needs to get done

This should at least be done for documentation, but could also be reflected in the organization of the source code.

Proposed organization:
1. Built-in
    1. dump
    1. nil
1. Multi-language
    1. ant
    1. autotools
    1. bazel
    1. cmake
    1. colcon
    1. jlink
    1. make
    1. meson
    1. qmake
    1. scons
1. .NET
    1. dotnet
    1. dotnet_v2
1. Go
    1. go
    1. go_use
1. Java
    1. gradle
    1. java
    1. maven
    1. maven_use
1. JavaScript
    1. npm
    1. npm_use
1. Python
    1. poetry
    1. python
    1. python_v2
    1. uv
1. Ruby
    1. ruby
1. Rust
    1. cargo_use
    1. rust
```

### Why it needs to get done

The number of plugins exceeds what's reasonable for a flat list. As they are sorted alphabetically, users can miss relevant plugins because they stop scrolling, e.g. choosing to use the Python plugin without realizing a uv plugin exists.

GreyViperTooth:
Group plugins by the language or technology they support, in both the documentation and the source tree.

**Documentation** (first commit): the plugins reference index is organized into nine groups — Built-in, Multi-language, .NET, Go, Java, JavaScript, Python, Ruby, and Rust — each with a short description, instead of one flat alphabetical list. Page URLs are unchanged.

**Source** (second commit): plugin modules move into per-language subpackages that mirror the documentation groups (for example `craft_parts/plugins/python/python_plugin.py`). Every previous module path remains importable through a compatibility alias module with explicit re-exports, so applications that do `from craft_parts.plugins import python_plugin` (as Snapcraft and Rockcraft do) are unaffected. A new unit test (`tests/unit/plugins/test_module_aliases.py`) checks for that guarantee. The `python_v2` package is already grouped and stays where it is.

This change was developed with the assistance of an AI tool (Claude Fable 5). I have reviewed and tested all of it before making the commits.

Fixes #1571

### DIFF
--- a/docs/how-to/document-plugin.rst
+++ b/docs/how-to/document-plugin.rst
@@ -12,6 +12,9 @@ headings and main text.
 
 Replace the instructions with suitable descriptions.
 
+After writing the page, add it to the appropriate language or technology
+group in the :ref:`plugins` reference index.
+
 
 <Name> plugin
 -------------
--- a/docs/reference/changelog.rst
+++ b/docs/reference/changelog.rst
@@ -40,6 +40,11 @@ Bug fixes:
   they were absent when packages are installed with ``--no-install-recommends``,
   causing builds to fail.
 
+Documentation:
+
+- Group the plugins in the reference by the language or technology they
+  support.
+
 .. _release-2.34.1:
 
 2.34.1 (2026-07-09)
--- a/docs/reference/plugins.rst
+++ b/docs/reference/plugins.rst
@@ -4,40 +4,130 @@ Plugins
 =======
 
 Plugins are used to provide support for specific technologies during the parts
-lifecycle.
+lifecycle. They are grouped here by the language or technology they support.
+
+
+Built-in
+--------
+
+These plugins cover the basic cases -- copying files verbatim, or bypassing
+the build altogether. They are available in every application built on Craft
+Parts.
 
 .. toctree::
    :maxdepth: 1
 
-   /common/craft-parts/reference/plugins/dotnet_v2_plugin.rst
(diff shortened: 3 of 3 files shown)


steinbro:
Hi @GreyViperTooth ! Thanks for the contribution.

My proposal in #1571 was primarily aimed at a more navigable documentation sidebar; I contemplated reorganizing the code to match, but I'm not convinced that's the right approach for the reasons visible in this PR: we'd meed to maintain stubs at the existing paths to keep all the imports working. The result is a noisier source tree, when the only motivation for the change was to make things cleaner.

Consequently, the only change to satisfy that issue would be to docs/reference/plugins.rst, which would make for a much simpler PR. Sorry for all the unnecessary work.

GreyViperTooth:
Hi @steinbro 
I understand your concern. The compatibility stubs were indeed noisy. I've dropped the source reorganization and rebased the branch down to a single docs commit: the grouped `docs/reference/plugins.rst` index and a matching note in the plugin-documentation how-to. 

Let me know how it looks now.

GreyViperTooth:
Hi @steinbro, I've finished making the changes. When you get a chance, please let me know how it looks now. Happy to make further adjustments.

steinbro:
Thanks. I've looped in @medubelko to assess how this might fit with other ongoing changes to the docs structure for apps like Snapcraft where the plugin references get included.

PR #1652, opened 2026-07-17
fix(parts): run override build on working subdir

soumyaDghosh:

---


lengau:
Gah this is going to be a problem. You are absolutely correct that this is a bug, but it's now load-bearing so we probably need to make it configurable so we can make this breaking change for future bases.

I'd like @tigarmo and @cmatsuoka to weigh in on this though.

From @tigarmo — this should be something in the LifecycleManager, similar to the enable usrmerge configuration

PR #1656, opened 2026-07-24
feat: add support for ignoring phasing

nicolasbock-canonical:
Debian packages use phasing to gradually push package updates to client
machines [1]. The eligibilaty of a client to receive an udpate during
phasing is deterministic and based on the machine ID, among other
things.  In a CI or build infrastructures using ephemeral builders, the
machine ID is not predictable and phasing becomes random to the extent
the builder selection is random.

In most cases, the user will want to most recent version of the packages
installed in the craft artifact, and phasing is not useful.

This change configures apt in the build container to always install
phased packages.

[1] https://ubuntu.com/project/docs/how-ubuntu-is-made/concepts/phased-updates/


craft_parts/packages/apt_cache.py
```diff
@@ -95,6 +95,10 @@ def __init__(
 
     # pylint: disable=attribute-defined-outside-init
     def __enter__(self) -> Self:
+        # Always ignore Ubuntu's phased updates so the latest available
+        # package version is installed.
+        apt_pkg.config.set("APT::Get::Always-Include-Phased-Updates", "true")
```
lengau, line 100: I guess my question is whether we should also be ignoring phasing on build snaps

nicolasbock-canonical: I agree, that would make sense. I can add that to this PR.

cmatsuoka (approved):
Thanks!

mr-cal (approved):
Thanks!

craft_parts/packages/apt_cache.py
lengau: Let's do it separately since this is approved already.

lengau (approved):

PR #1669, opened 2026-08-13
feat: add stage-slices key

mr-cal:
- [x] Have you successfully run `make lint && make test`?https://docs.google.com/document/d/1tQUv6iMq06uq5-dMPAX0qnOr0Qx0NASA7aceuoZUc-I/edit?tab=t.0

---

Adds a `stage-slices` key per [ST179](https://docs.google.com/document/d/1tQUv6iMq06uq5-dMPAX0qnOr0Qx0NASA7aceuoZUc-I/edit?tab=t.0).

Split into 3 commits for your reviewing pleasure.

This is how I understood the division of responsibility between craft-parts and craft-application/downstream apps:

#### craft-parts

- All validation and errors related to defining slices in `stage-packages` and `stage-slices`.

#### craft-application

- Deprecation warnings about using slices in `stage-packages`.
- Deciding when to allow slices in `stage-packages`.

(CRAFT-5253)


craft_parts/parts.py
```diff
@@ -603,37 +621,57 @@ def validate_overlay_mutually_exclusive(self) -> Self:
...
+        if has_slices(stage_packages):
+            if has_packages(stage_packages):
+                raise ValueError("cannot mix packages and slices in 'stage-packages'")
+
+            context: dict[str, Any] = info.context or {}
+            if not context.get("stage_packages_slice_support", True):
+                raise ValueError(
+                    "Chisel slices cannot be declared in 'stage-packages'. "
+                    "Use the 'stage-slices' key instead."
+                )
```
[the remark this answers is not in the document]
mr-cal: I'm not going to make `stage-packages` more strict, that would risk breaking existing project files.

craft_parts/utils/deb_utils.py
```diff
@@ -36,3 +36,29 @@ def extract_deb(
...
+
+
+def is_chisel_slice(name: str) -> bool:
+    """Return whether a name is a Deb package or a Chisel slice.
+
+    A Chisel slice uses the `<package-name>_<slice-name>` syntax.
+
+    :param name: A package or slice name.
+    """
+    return "_" in name
```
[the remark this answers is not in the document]
mr-cal: I made it clearer that this functions assumes it's either a valid package or a valid slice.

craft_parts/utils/deb_utils.py
mr-cal, line 1: This refactoring wasn't strictly necessary, but the duplication was bothering me and it made the validator more readable.

Plus craft-application will be able to leverage these for deprecation warnings.

craft_parts/constraints.py
```diff
@@ -25,6 +26,35 @@
 Tv = TypeVar("Tv")
 
 
+def get_validator_by_regex(
```
mr-cal, line 29: This is a verbatim copy from [craft-application](https://github.com/canonical/craft-application/blob/50d8cc69a88df107818a7b11e7f7241278ddb4d1/craft_application/models/constraints.py#L39).

craft_parts/parts.py
```diff
@@ -325,10 +327,26 @@ class PartSpec(BaseModel):
     """During the stage step, these packages are included in the stage environment
     alongside the build artifacts.
 
-    Chisel slices can be listed using the ``<package-name>_<slice-name>`` syntax.
+    Chisel slices should be defined with the ``stage-slices`` key.
```
mr-cal: @jahn-junior - can you review the stage-packages and stage-slices sections?

https://canonical-ubuntu-documentation-library--1669.com.readthedocs.build/craft-parts/1669/common/craft-parts/reference/part_properties/#stage-packages

craft_parts/executor/part_handler.py
```diff
@@ -1281,7 +1281,9 @@ def _fetch_stage_packages(self, *, step_info: StepInfo) -> list[str] | None:
 
         :raises StagePackageNotFound: If a package is not available for download.
         """
-        stage_packages = self._part.spec.stage_packages
+        # 'stage-packages' and 'stage-slices' are mutually exclusive, so at most one
+        # of these is populated.
+        stage_packages = self._part.spec.stage_packages or self._part.spec.stage_slices
```
cmatsuoka, line 1286: We originally had debs or slices in stage packages, and the installation logic invoked `_unpack_stage_packages`/`_unpack_stage_slices` for each of the cases based on the contents of `stage_packages`. Now that we have specialized stage slices (which can be deb or other types of slices), it looks like we could move this decision one level up and invoke `unpack_stage_packages`/`unpack_stage_slices` from there instead of conflating everything into stage packages, and have two clean, separate paths for each format, wdyt?

mr-cal: I'm not sure if that would gain us much. Since we still support slices in `stage-packages`, we still need to keep the logic that de-conflate slices from a `stage-packages` entry.

If we had entirely removed that, then yes I think your refactoring idea would be a great idea.

I'm not opinionated on whether that logic lives in `craft_parts.packages.Repository` or `craft_parts.executor.PartHandler`, but I don't see a strong reason to move it in this PR.

craft_parts/parts.py
jahn-junior:
```suggestion
    Chisel slices should be declared with the ``stage-slices`` key. Support for
    chisel slices in the ``stage-packages`` key is deprecated and will not be
    supported in a future release.
```
I'd add this notice here and drop the next paragraph.

craft_parts/parts.py
```diff
@@ -325,10 +327,26 @@ class PartSpec(BaseModel):
...
 
-    Listing both packages and slices in the same ``stage-packages`` key is not
-    currently supported.
+    (deprecated) Chisel slices can be listed using the ``<package-name>_<slice-name>``
+    syntax. Listing both packages and slices in the same ``stage-packages`` key is not
+    supported.
```
jahn-junior:
```suggestion
```
Going along with https://github.com/canonical/craft-parts/pull/1669/changes#r3785533425, documenting the syntax for a deprecated feature feels like giving someone a footgun. If switching to the new key required a nontrivial effort from users, I might be more inclined to leave this in.

craft_parts/parts.py
```diff
@@ -325,10 +327,26 @@ class PartSpec(BaseModel):
...
+
+    This key is mutually incompatible with the ``stage-slices`` key.
+    """
+
+    stage_slices: list[ChiselSliceStr] = Field(
+        default=[],
+        description="The Chisel slices to include in the stage environment.",
+        examples=["[ca-certificates_data, bash_bins]"],
+    )
+    """During the stage step, these Chisel slices are cut into the stage environment
```
jahn-junior, line 344:
```suggestion
    """During the stage step, these Chisel slices are included in the stage environment
```
Are the slices actually created during the stage step? If so, the original wording is fine and you can ignore this.

craft_parts/parts.py
```diff
@@ -325,10 +327,26 @@ class PartSpec(BaseModel):
...
+    alongside the build artifacts.
+
+    Each entry must use the ``<package-name>_<slice-name>`` syntax.
```
jahn-junior:
```suggestion
    Each entry must be listed as ``<package-name>_<slice-name>``.
```
This one's not a big deal. What you have here is consistent with the `stage-packages` docstring. I just never liked the wording much.

jahn-junior (changes requested):

craft_parts/parts.py
mr-cal: Thanks for the perspective, I'll drop it.

I don't like advertising deprecated functionality, but on the other hand, if someone saw slices in an existing `stage-packages` entry, they would be confused to why the documentation doesn't mention it.

craft_parts/parts.py
mr-cal: I think we'll support this forever for existing bases. 

The plan is that it won't be supported in 26.10 and higher for all apps, except Snapcraft, where it won't be supported in core28 and higher. 

But I don't think there's a good way for us to document this generically in craft-parts.  Should I just exclude the "_and will not be supported in a future release_" part?

craft_parts/parts.py
mr-cal: We do call `chisel cut` during the stage step, so yes I think the original wording is accurate.

craft_parts/parts.py
mr-cal: https://github.com/canonical/craft-parts/pull/1669/commits/6e50eca35ab8a678641233113e87c5c94696d64a

craft_parts/parts.py
mr-cal: https://github.com/canonical/craft-parts/pull/1669/commits/6e50eca35ab8a678641233113e87c5c94696d64a

craft_parts/parts.py
mr-cal: Preemptively dropping the future release section: https://github.com/canonical/craft-parts/pull/1669/commits/6e50eca35ab8a678641233113e87c5c94696d64a

jahn-junior (approved):
The updated key descriptions LGTM. Thanks!

craft_parts/parts.py
jahn-junior: That's what I would've suggested. This is another one that makes me wish we had base-specific docs. :crying_cat_face:

craft_parts/utils/deb_utils.py
```diff
@@ -36,3 +36,32 @@ def extract_deb(
...
+
+def has_slices(names: Iterable[str]) -> bool:
+    """Return whether a list contains any Chisel slices.
+
+    :param names: An iterable of packages and slices.
+    """
+    return any(_is_chisel_slice(name) for name in names)
+
+
+def has_packages(names: Iterable[str]) -> bool:
```
tigarmo: maybe `has_debs` instead? I think the name 'package' is a bit overloaded; it means exclusively 'debs' in some places and a generic thing in others

craft_parts/parts.py
```diff
@@ -603,37 +618,57 @@ def validate_overlay_mutually_exclusive(self) -> Self:
...
-        has_slices = any(name for name in stage_packages if is_slice(name))
-        has_packages = any(name for name in stage_packages if not is_slice(name))
 
-        if has_slices and has_packages:
-            raise ValueError("Cannot mix packages and slices in stage-packages")
+        if has_slices(stage_packages):
+            if has_packages(stage_packages):
+                raise ValueError("cannot mix packages and slices in 'stage-packages'")
+
+            context: dict[str, Any] = info.context or {}
```
tigarmo, line 652: Interesting; this `context` is only populated during the LifecycleManager constructor, right? Which means it's quite late in the application lifecycle (we would have already created the build instance, for example). I'm OK with this extra check but I think we need a way to 'fail earlier'. One suggestion that comes to mind is a free function like `check_slices_on_packages` that applications can call during their model validation?

tigarmo (commented):
Thanks. The only thing I think we need to agree on is the exact mechanism for how applications will block slices-in-packages depending on project base. Specifically, craft-parts' role in this.

craft_parts/parts.py
mr-cal: Good point. I think we should update [`validate_part()`](https://github.com/canonical/craft-parts/blob/6e50eca35ab8a678641233113e87c5c94696d64a/craft_parts/parts.py#L1405) to take an optional context. That way, craft-application doesn't need to call anything besides [this](https://github.com/canonical/craft-application/blob/043bc646c1e75d37d84fa3f865c75cf1ba9cae9d/craft_application/models/project.py#L107-L110).

craft_parts/parts.py
tigarmo: what? where?

tigarmo: If memory serves, chisel slices are cut at the same time deb packages would've been extracted - just before running the `build` step

craft_parts/parts.py
tigarmo: I'm fine with updating `validate_part()`, but I'll note that the way it's used currently in craft-app is on [Part's type annotation](https://github.com/canonical/craft-application/blob/043bc646c1e75d37d84fa3f865c75cf1ba9cae9d/craft_application/models/project.py#L114), which means we can't easily make it 'dynamic' based on the project's base.

In practice I think it means that apps will need a whole new validator, probably a root one, to validate against all parts based on the `base` and `build-base`.

mr-cal: Hm, ok. We could do your suggestion of downstream apps being able to call a `check_slices_on_packages()` from craft-parts. We could also give downstream apps a function they can override (see code below). It's a bit more hairy due to pydantic, but the upside is that downstream models wouldn't need to be aware of craft-parts or import anything from it.

Anyway, I'm not going to prepare craft-parts for either option and be stuck with an unused API. Instead, I'll leave this decision to whomever first needs to ban slices in `stage-packages` for 26.10 in a downstream app.


#### craft-app Project model:
```python
    @classmethod
    def _stage_packages_slice_support(cls, project_data: dict[str, Any]) -> bool:
        """To be overridden by downstream apps."""
        return True

    # runs before validation from the Parts' annotation 
    @pydantic.field_validator("parts", mode="before")
    @classmethod
    def _stash_stage_packages_slice_context(cls, value: Any, info: pydantic.ValidationInfo) -> Any:
        if info.context is not None:
            info.context["stage_packages_slice_support"] = cls._stage_packages_slice_support(info.data)
        return value

    def _validate_part(part: dict[str, Any], info: pydantic.ValidationInfo) -> dict[str, Any]:
        craft_parts.validate_part(part, context=info.context)
        return part
```

#### downstream app Project model:
```python
    @override
    @classmethod
    def _stage_packages_slice_support(cls, project_data: dict[str, Any]) -> bool:
        base = project_data.get("build_base") or project_data.get("base")
        return base in ["20.04", ..., "26.04"]
```

tigarmo: > Anyway, I'm not going to prepare craft-parts for either option and be stuck with an unused API.

👍🏾

tigarmo (approved):
thanks!

craft_parts/executor/part_handler.py
cmatsuoka: My main point of concern about the slices support is that it won't be .deb slice specific anymore and currently our chisel support in in packages/deb. But we can refactor this later when new slices are actually implemented.

cmatsuoka (approved):
This will work for the current state of affairs but we'll need refactoring as the slice definition changes in the near future.

craft_parts/executor/part_handler.py
mr-cal: Ah, I understand now. I agree that we can wait until slices are no longer specific to debs.

PR #1690, opened 2026-08-23
feat: plugins use native tls certificates

### LINKED ISSUE #1257: Plugins should configure build tools to use native TLS certificates
### What needs to get done

We should use the native TLS certificate store rather than a plugin's built-in cert store. For example, the `uv` plugin should [set `UV_NATIVE_TLS=true`](https://docs.astral.sh/uv/concepts/authentication/certificates/#system-certificates)

### Why it needs to get done

For building behind proxies

cmatsuoka:
Update uv, poetry and npm plugins to use the default system-wide TLS
certificates.

Fixes #1257

---


tests/unit/plugins/test_uv_plugin.py
```diff
@@ -116,3 +116,32 @@ def test_get_out_of_source_build(plugin):
...
+)
+def test_get_build_environment_respects_user_tls_settings(
+    new_dir, plugin, env_var, monkeypatch
+):
+    """When the user already set the variable, the plugin must not override it.
+
+    The actual user-provided value will come from the build environment, so it
+    should not appear in the plugin's returned dictionary.
+    """
+    monkeypatch.setenv(env_var, "false")
```
lengau: Instead of testing like this, could we test it using the `build-environment` key?

bepri (approved):

tests/unit/plugins/test_uv_plugin.py
cmatsuoka: Now testing value overrides using `build-environment` (this is set after the plugin defaults so no need to test for the presence of previously set values there).

cmatsuoka: The same fix was applied on the npm and poetry plugins.

lengau (approved):
LGTM, thanks!

I don't understand why the minimum version tests have failures on the conflicts test, especially since those tests shouldn't be affected by these changes.

cmatsuoka:
There are some test failures related to mknod permission errors, but these don't seem to be caused by this change. Maybe something changed in the runner?

lengau:
@cmatsuoka agreed. I'm re-running the job from main [here](https://github.com/canonical/craft-parts/actions/runs/33785619738/job/101105843473) just to validate. If main fails I'm happy for us to override and merge.

cmatsuoka:
The same issue is happening there.

cmatsuoka:
This branch:
`= 4 failed, 5511 passed, 19 skipped, 212 deselected, 1 xfailed, 15 warnings, 6 errors in 310.92s (0:05:10) =`

Main branch:
`= 4 failed, 5485 passed, 19 skipped, 212 deselected, 1 xfailed, 15 warnings, 6 errors in 245.97s (0:04:05) =`

PR #1691, opened 2026-08-24
fix(validator): use stdout for command output

### LINKED ISSUE #1447: snap build fails if go must download the version
### Bug Description

I see `snapcraft pack` fail with:

```
Environment validation failed for part 'semver': invalid go compiler version 'go: downloading go1.25.6 (linux/amd64)'.
Failed to run snapcraft in instance
```

It works when I execute `snapcraft pack` again. I assume that's because the required Go version was then already downloaded on the instance.

I took a quick look at the code where this error originates from. [This](https://github.com/canonical/craft-parts//blob/71291d50ec95aed813215e89463866dd09aaaa5e/craft_parts/plugins/go_plugin.py#L66-L71) is the call to `go version` which calls `subprocess.run` and [returns the stderr](https://github.com/canonical/craft-parts//blob/71291d50ec95aed813215e89463866dd09aaaa5e/craft_parts/plugins/validator.py#L145-L152) it it's non-empty, which it then assumes is the go version. It should use the output on stdout instead.

### To Reproduce

I encounter this when trying to build the authd snaps:

```bash
git clone https://github.com/canonical/authd
cd authd
# Prepare snapcraft.yaml
./scripts/prepare-broker-variant-files
# Build the snap
snapcraft pack
```

### part yaml

```yaml
  semver:
    source: authd-oidc-brokers/tools
    source-type: local
    plugin: go
    override-build: |
      go mod download all
      go mod vendor
      go build -o ${GOBIN}/semver semver/semver.go
```

### Relevant log output

I assume the above is enough to understand the issue. Let me know if you do need logs.

gcomneno:
## Summary

Fix plugin environment validation when a dependency writes informational messages to stderr while its semantic version is written to stdout.

This affects the Go plugin when Go automatically downloads a toolchain during `go version`. In that case Go can emit:

`go: downloading go1.25.6 (linux/amd64)`

to stderr, and the previous validator implementation incorrectly treated stderr as the command output.

This PR makes `_execute()` return stdout as the command output.

Closes #1447.

## Validation

- TDD regression test initially failed with the previous implementation:
  `stdout-value` expected, `stderr-value` returned.
- Regression test passes after the fix.
- `tests/unit/plugins/test_validator.py`: 6 passed
- `tests/unit/plugins/test_go_plugin.py`: 19 passed
- Ruff check: passed
- Ruff format check: passed
- `git diff --check`: passed

## AI Disclosure

AI assistance was used during analysis and implementation of this contribution.

- Tool: ChatGPT
- Scope: codebase analysis, debugging/reasoning, regression-test draft, implementation draft, and commit/PR wording
- Human review: completed; the resulting code and tests were reviewed and validated locally by the contributor


gcomneno:
The failing QA job is:

`test-java-plugins / Slow tests (noble, s390x, edge, 3.12)`

The failure appears unrelated to this PR.

I compared it with the same QA job on `main` before this PR was created:

- `main` run `32527752312` (2026-08-21)
- PR #1691 run `32693609794` (2026-08-24)
- both use Ubuntu Noble, `s390x`, edge, Python 3.12
- both use OpenJDK `21.0.11+10`
- both fail inside `/usr/lib/jvm/java-21-openjdk-s390x/bin/jlink`
- both terminate with exit code `134` / `Aborted (core dumped)`
- both report a native JVM `SIGSEGV` inside `libjvm.so` / `PhaseOutput::*`

The exact failing test differs between the runs:

- on `main`: `test_jlink_plugin_set_modules[none]`
- on #1691: `test_jlink_plugin_embedded_jar[none]`

The crash location also differs within `PhaseOutput`, so I am not claiming identical crash signatures. However, the prior failure on `main` demonstrates that this `s390x` / OpenJDK 21 `jlink` failure mode predates this PR.

No code changes were made in response to this CI failure.

gcomneno:
Addressed the Copilot review feedback in commit `d874e282`.

The `_execute()` docstring now accurately documents that the method returns command stdout, matching its current behavior.

Validation after the follow-up:

- `tests/unit/plugins/test_validator.py`: 6 passed
- `tests/unit/plugins/test_go_plugin.py`: 19 passed
- Ruff check: passed
- Ruff format check: passed
- `git diff --check`: passed

No functional behavior was changed by this follow-up.

gcomneno:
The follow-up requested by Copilot was a one-line docstring fix, and the local validation passes.

The new QA run then failed in `docs-lint` because the external link checker timed out while accessing:

https://packages.ubuntu.com/noble/dotnet-sdk-8.0

So after fixing one line of documentation, CI replied with a network timeout from `packages.ubuntu.com`. 🙂

All code linters and the documentation build itself passed; the failure is limited to the external link check.

No further code changes appear necessary from this failure.

PR #1692, opened 2026-08-25
fix: refresh build snaps if already installed

### LINKED ISSUE #496: part `build-snaps` channel change doesn't work without a `clean`
### Bug Description

In my `charmcraft.yaml` file I updated the charm version via the `build-snaps` directive, and then a `charmcraft pack` but the changes were not reflected, ie no new charm version was installed.

Only when I ran `charmcraft clean` and another `charmcraft pack` it worked as intended and used the new charm version from the `charmcraft.yaml` configuration.

### To Reproduce

- create a simple `charmcraft.yaml` with e.g. `build-snaps: [charm/2.x/stable]`
- run `charmcraft pack` (which creates a container)
- update `charmcraft.yaml` to `build-snaps: [charm]`
- run `charmcraft pack`
- the new charm version is not used

(I noticed that as I added an option which is only available with the new charm version).

### Environment

- Ubuntu 20.04
- charmcraft -> version 2.3.0
- using lxd containers

### charmcraft.yaml

```shell
see above
```

### Relevant log output

```shell
none
```

cmatsuoka:
If a build snap is already installed, refresh it to the current channel
to reflect changes in the project. Also add it to the build state so
the part will be rebuilt in case of channel changes.

Fixes #496

---


tests/unit/packages/test_snaps.py
```diff
@@ -22,8 +22,25 @@
...
+        ("snap", "expected_channel"),
+        [
+            ("fake-snap", "latest/stable"),
+            ("fake-snap@stable", "latest/stable"),
+            ("fake-snap@beta", "latest/beta"),
+            ("fake-snap@candidate", "latest/candidate"),
+            ("fake-snap@edge", "latest/edge"),
+            ("fake-snap@beta/branch", "latest/beta/branch"),
+            ("fake-snap@edge2/stable", "edge2/stable"),
+            ("fake-snap@stable-track/candidate", "stable-track/candidate"),
```
lengau, line 37: Can you add some cases for `<risk>/<branch>` and `<track>/<risk>/<branch>` too?

lengau (approved):
LGTM with one request

tests/unit/packages/test_snaps.py
cmatsuoka: Added cases `edge/branch` and `track/candidaate/branch`, and also one track called `betamax` just to cover the case of something starting with `beta` not being considered a risk.

mr-cal (approved):
LGTM

craft_parts/packages/errors.py
```diff
@@ -223,12 +223,14 @@ class SnapInstallError(PackagesError):
...
-    def __init__(self, *, snap_name: str, snap_channel: str) -> None:
+    def __init__(
+        self, *, snap_name: str, snap_channel: str, detail: str | None = None
+    ) -> None:
         self.snap_name = snap_name
         self.snap_channel = snap_channel
         brief = f"Error installing snap {snap_name!r} from channel {snap_channel!r}."
 
-        super().__init__(brief=brief)
+        super().__init__(brief=brief, details=detail)
```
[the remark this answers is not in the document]
cmatsuoka: Fixed.

Makefile
```diff
@@ -256,7 +256,7 @@ ifneq ($(shell which chisel),)
 else ifeq ($(shell which snap),)
 	$(warning Cannot install chisel without snap. Please install it yourself.)
 else
-	sudo snap install chisel --channel latest/candidate
+	sudo snap install chisel --channel latest/stable
```
mr-cal: Why? Isn't this so we catch issues with upcoming chisel releases before they make it to stable?

cmatsuoka: Oh, I thought this was added back at the time when there was no stable chisel available. Reverting this change.,

cmatsuoka: Should we also move the internal chisel tests to use latest/candidate? Many of them install from stable.

mr-cal: I think that's a good idea, I'm fine if that's done here or separately.

mr-cal (approved):

craft_parts/packages/snaps.py
```diff
@@ -45,6 +45,29 @@
...
 
 
+def _normalize_channel(channel: str) -> str:
+    if not channel:
+        return ""
+
+    if any(
+        channel == risk or channel.startswith(f"{risk}/") for risk in _CHANNEL_RISKS
+    ):
+        return f"latest/{channel}"
```
lengau, line 55: I don't think this is correct. If you install `lxd` for instance and you set `--channel=edge`, it installs from the `5.21/edge` channel since `5.21` is the default track.

PR #1712, opened 2026-09-07
feat: fpc plugin

lczyk:

---

This PR adds a Free Pascal (`fpc`) plugin. Each entry in `fpc-programs` becomes one executable in `$CRAFT_PART_INSTALL/bin`. Units and include files are resolved from `fpc-unit-paths` and `fpc-include-paths`. `fpc-parameters` passes additional compiler options and is placed after the generated flags so it can override them.

The plugin does not provision the compiler, to leave the choice of version open. It is expected from `build-packages`, or from a part named `fpc-deps`, following the convention of the other compiler plugins.

Search paths go through `shlex.quote`, so a trailing `/*` reaches the compiler unexpanded and fpc resolves it as "every subdirectory of this one".

Part of a set:

- https://github.com/canonical/craft-parts/pull/1712 **(this PR)**
- ~~https://github.com/canonical/craft-parts/pull/1713~~


lczyk:
CI failures look unrelated and happen on other PRs too

craft_parts/plugins/fpc_plugin.py
```diff
@@ -0,0 +1,152 @@
...
+        :raises PluginEnvironmentValidationError: If fpc is invalid
+          and there are no parts named fpc-deps.
+        """
+        version = self.validate_dependency(
+            dependency="fpc",
+            plugin_name="fpc",
+            part_dependencies=part_dependencies,
+            argument="-iV",
+        )
+        if not _VERSION_PATTERN.match(version) and (
```
lengau, line 74: Since it appears the full output is just a `major.minor.patch`, should we give this a `fullmatch` and update the pattern to include the patch?

lengau (commented):
This is pretty cool, but are there any real packages that use this and don't have some sort of Makefile or equivalent that would wrap it?

PR #1717, opened 2026-09-09
feat(part): add build-slices model

tigarmo:
'build-slices' is a Part property that lists Chisel slices that must be cut to form the part's 'build root'. This commit just adds the model gated by a new Feature, plus validation.

---

### DIFF
--- a/craft_parts/parts.py
+++ b/craft_parts/parts.py
@@ -380,6 +380,18 @@ class PartSpec(BaseModel):
     Build packages must be listed by their name on the host system.
     """
 
+    build_slices: list[ChiselSliceStr] = Field(
+        default=[],
+        description="The Chisel slices to make available during the build.",
+        examples=["[python3.12_standard, gcc-12_bins]"],
+    )
+    """The Chisel slices to make available during the build step.
+    The slices from all parts are collected and cut into a single directory that
+    becomes the system root of the build step. Since slices and Debian packages
+    cannot be safely mixed in a single filesystem, ``build-slices`` is mutually
+    exclusive with ``build-packages`` and ``build-snaps`` in the same part.
+    """
+
     build_environment: list[dict[str, str]] = Field(
         default=[],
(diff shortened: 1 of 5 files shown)


craft_parts/parts.py
```diff
@@ -380,6 +380,18 @@ class PartSpec(BaseModel):
     Build packages must be listed by their name on the host system.
     """
 
+    build_slices: list[ChiselSliceStr] = Field(
```
lengau, line 383: This matches with just a slice name, but doesn't handle the channel stuff being discussed elsewhere. Is that intentional?

craft_parts/parts.py
```diff
@@ -624,6 +636,23 @@ def validate_overlay_mutually_exclusive(self) -> Self:
             )
         return self
 
+    @field_validator("build_slices")
+    @classmethod
+    def validate_build_slices_feature(cls, item: _T_validate) -> _T_validate:
+        """Check if build-slices is specified when the feature is disabled."""
+        if item and not Features().enable_build_slices:
+            raise ValueError("'build-slices' are not supported")
```
cmatsuoka: Should this be *is* not supported because `'build-slices'` refers to the key and not to build slices themselves, like in "The 'build-slices' key is not supported"?

craft_parts/parts.py
tigarmo: yes. Once we have some slices using channels I think we should do separate PR updating `ChiselSliceStr`

craft_parts/parts.py
tigarmo: I like your suggestion, thanks!

craft_parts/parts.py
lengau: Cool, sounds good to me.

Honestly I wish we had a place to put our generic models, as I'd love to have a model that describes just a track, so we could build compound models out of stuff like that.

lengau (approved):
LGTM, thanks!

mr-cal (approved):

PR #1724, opened 2026-09-17
feat: run 'build' in build-slices root

tigarmo:

---


tigarmo:
@cmatsuoka @mr-cal ready for a first review of the overall approach and the integration test.

craft_parts/overlays/chroot.py
```diff
@@ -231,10 +264,14 @@ def mount_to(self, chroot: Path, *args: str) -> None:
             return
         self.create_dst(chroot)
         self._mount(self.src, chroot, *self.args, *args)
+        self._mounted = True
 
     def unmount_from(self, chroot: Path, *args: str) -> None:
         """Unmount `self.dst` within chroot."""
         logger.debug(f"Mounting {self.dst}")
```
mr-cal, line 271: unrelated: should this be "unmounting"?

tests/integration/lifecycle/data/build-slices/parts.yaml
```diff
@@ -0,0 +1,42 @@
...
+    override-build: |
+      printf '%s\n' dependency > "$CRAFT_PART_INSTALL/dependency"
+
+  sliced:
+    after: [dependency-regular-build]
+    plugin: nil
+    build-slices:
+      - bash_bins
+      - base-files_base
+    override-build: |
```
mr-cal, line 13: does `craftctl` work in here? Should it?

tests/integration/lifecycle/test_build_slices.py
```diff
@@ -0,0 +1,60 @@
...
+    completed_process = subprocess.run(
+        [tmp_homedir_path / "prime/bin/hello"],
+        check=True,
+        capture_output=True,
+        text=True,
+    )
+    assert completed_process.stdout == "hello from build slices\n"
+    assert not (tmp_homedir_path / "stage/forbidden").exists()
+    assert host_only_file.exists()
+    assert os.path.ismount(tmp_homedir_path / "stage") is False
```
mr-cal, line 60: What are you checking here - that the unmount occurred? If so, shouldn't it be build-slices/stage?

craft_parts/overlays/chroot.py
mr-cal, line 1: At some point we should move this out of overlays, since it isn't specific to overlays anymore.

mr-cal (approved):
First draft looks good to me.

craft_parts/overlays/chroot.py
cmatsuoka: Yes, I think this should be Unmounting.

tests/integration/lifecycle/data/build-slices/parts.yaml
cmatsuoka: From the user's point of view, I would expect that `craftctl default` works (if using a non-nil plugin).

craft_parts/executor/part_handler.py
```diff
@@ -484,13 +485,25 @@ def _run_build(
...
             self._overlay_manager, top_part=self._part, condition=needs_overlay
         ):
-            self._run_step(
-                step_info=step_info,
-                scriptlet_name="override-build",
-                work_dir=self._part.part_build_dir,
-                stdout=stdout,
-                stderr=stderr,
-            )
+            if self._part.spec.build_slices:
```
cmatsuoka, line 488: Maybe we could have a `has_slices` part property to hide this and make the testing more uniform? (We were previously testing partition presence using the partitions definition directly, and the test strategy ended up not being consistent)

cmatsuoka (approved):
This implements what was discussed and I don't see conflicts with ST185, but please consider addressing some of the comments.

tests/integration/lifecycle/data/build-slices/parts.yaml
tigarmo: `craftctl` at the moment doesn't work, because the executable is not available inside the chroot:

```
E           :: + craftctl default
E           :: /tmp/tmp5ym7mvxa/scriptlet.sh: line 5: craftctl: command not found
E           Review the scriptlet and make sure it's correct.
```

It's not as simple as just copying it over because it's a Python script that needs the libraries from the app snap. We definitely need to address this but it's a limitation at the moment.

craft_parts/executor/part_handler.py
tigarmo: There's already a `has_slices` property that checks in `stage-packages`; the idea is good so I added a `has_build_slices` instead

tests/integration/lifecycle/test_build_slices.py
tigarmo: thanks yeah this was wrong. Updated.

mr-cal (approved):
LGTM

